PR #434, opened 2013-09-27
Self Signed SSL - Third Times the Charm

jesseproudman:
Trying for the last time to get Travis to build.

See #432, #430 and #422.


mephmanx:
For anyone on this, would like to ask a question..  Does having this set to true or false mean anything when a non-https url is used?  I am trying to use an all-http (non https) openstack cloud to set up BOSH and am getting errors creating the director disk using bbl.  I am trying to work out if it is a bbl thing, my configuration, or if glance images require ssl....  I know this is a very old change...it came up kind of related through google....

PR #445, opened 2013-10-25
Adding boot from volume support for OpenStack.

amhuber:
This change adds support for boot from volume to the OpenStack CPI.  To use, add the following property to the Micro BOSH or BOSH deployment manifest:

```
cloud:
  plugin: openstack
  properties:
    openstack:
      boot_from_volume: true
```

Changes:
- Add a reference to the fog Volume API which supports created volumes with an imageRef
- Create a boot volume before creating the VM with the disk size specified by the flavor
- Add a block device mapping to the VM on creation to mount the new volume at /dev/vda
- Skip setting the "personality" property on VM creation since file injection is not supported with BFV

### DIFF
--- a/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
+++ b/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
@@ -37,6 +37,7 @@ module Bosh::OpenStackCloud
       @default_security_groups = @openstack_properties["default_security_groups"]
       @state_timeout = @openstack_properties["state_timeout"]
       @stemcell_public_visibility = @openstack_properties["stemcell_public_visibility"]
+      @boot_from_volume = @openstack_properties["boot_from_volume"]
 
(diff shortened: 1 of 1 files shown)


shashankmjain:
Does the boot from volume survive restart, If some data exists on boot volume(say some packages), will Bosh boot from the same volume in case of say VM restart or recreation?

cppforlife:
vm restart is typically not bosh involved so thats up to openstack. vm recreation  gurantees that root and ephemeral disks are new. packages are reinstalled everytime vm is created.

 if you need to store some data i would recommend placing it on /var/vcap/store which is for persistent data.

shashankmjain:
Thanks.
Yes for persistent data we use store.
Question is about packages and then also the advantage of booting from a
volume. Openstack provides a reserve feature in case we want to get the
same volume for the rootdisk.
Thanks
Shashank

PR #521, opened 2014-02-14
Rebased Version of Director blobstores for swift

reneedv:
I've squashed the history from Dr. Nic's changes from PR #450 and mine as requested in PR #520


cppforlife:
We have decided for now to only support swift through its S3 compatible api. Closing.

Issue #573, opened 2014-05-08
AWS VPC 2.0 AssociatePublicIpAddress support


akranga:
I am trying to create bosh at AWS VPC subnetwork. The deployment manifest looks like the following.

```
...
network:
  type: manual
  ip: 10.0.0.6
  dns:
  - 10.0.0.2
  cloud_properties:
    subnet: CHANGE_ME_PLEASE
...
```

Manual network, I specify private IP, but don't want to specify Elastic IP (this resource is limited in AWS).

When I run "bosh micro deploy ..." command I am getting error:

```
.../bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager/aws.rb:124:in `discover_client_services_ip': Failed to discover public ip address (RuntimeError)
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager/aws.rb:67:in `client_services_ip'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager/aws.rb:29:in `remote_tunnel'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:453:in `wait_until_agent_ready'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:130:in `block in create'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:85:in `step'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:128:in `create'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:201:in `update'
        from /home/ubuntu/.gem/ruby/1.9.3/gems/bosh_cli_plugin_micro-1.2524.0/lib/bosh/deployer/instance_manager.rb:102:in `block in update_deployment'
[...]

PR #575, opened 2014-05-09
Issue #573   associate public ip address support 

akranga:
Issue #573 resolution

Added support for AssociatePublicIpAddress network parameter for manual networks.

AssociatePublicIpAddress can be explicitly described in deployment manifest (aws or resource pool props)
or assigned by convention (manual network without vip)


cppforlife:
Implemented in AWS CPI release: https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/releases/tag/v55

PR #579, opened 2014-05-14
Allow resource limits to be specified for Warden in the director config

peterellisjones:
When using bosh-lite to run processes that need many open file descriptors (eg Cassandra, Jenkins etc.), the default 1024 limit is insufficient. This allows us to change these limits within the warden containers by adding them to the director.yml:

``` yml
---
name: Bosh Lite Director
# ...
cloud:
  plugin: warden
  properties:
    warden:
      unix_domain_socket: "/tmp/warden.sock"
      rlimits:
        nofiles: 10240
# ...
```

All options that can be specified within the rlimits property are defined in [Warden::Protocol::ResourceLimits](https://github.com/cloudfoundry/warden/blob/master/warden-protocol/lib/warden/protocol/pb/resource_limits.proto)

### DIFF
--- a/bosh_warden_cpi/lib/cloud/warden/helpers.rb
+++ b/bosh_warden_cpi/lib/cloud/warden/helpers.rb
@@ -99,9 +99,23 @@ module Bosh::WardenCloud
         request.handle = handle
         request.privileged = true
         request.script = '/usr/sbin/runsvdir-start'
+        request.rlimits = resource_limits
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
Closing - if anyone is still interested, please feel free to send a PR to the warden CPI's repository now at https://github.com/cppforlife/bosh-warden-cpi-release.

PR #671, opened 2014-10-09
Enable the bosh_micro_enabled parameter

liuhewei:
Currently, 'bosh_micro_enabled' is hardcoded to 'yes' in `builder_options.rb`, so the setting in `settings.bash` is useless. This parameter is used to control if the stemcell includes the micro-bosh package or not when building stemcell. It will reduce ~100 mb size if excluding the micro-bosh package from stemcell.

In this modification, we removed the hardcode in `builder_options.rb`, and re-organized the code of `apply.sh`, `config.sh` according to the 'bosh_micro_enabled' setting value. Then, the default value is kept 'yes' and user can control the stemcell to include micro-bosh package or not through the 'bosh_micro_enabled' setting in `settings.bash`.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/builder_options.rb
@@ -69,7 +69,6 @@ module Bosh::Stemcell
 
     def bosh_micro_options
       {
-        'bosh_micro_enabled' => 'yes',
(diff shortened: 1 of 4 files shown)


cppforlife:
Introduced a `BOSH_MICRO_ENABLED` env variable instead.

PR #678, opened 2014-10-15
Only delete buckets from current region

danhigham:
Despite S3 buckets not being associated to a region, they totally are! If you try and delete a bucket associated to a region whilst your S3 creds are pointing to another, errors occur!


cppforlife:
bosh_cli_plugin_aws is deprecated. We will be removing source for it soon.

drnic:
BTW, the aws plugin was the only place I found documentation for setting up
ELBs (the TCP:80 health check, 80, 443, 4443 listeners etc) for CF. Perhaps
ask the CF/ops docs ppl to write up a little something about this.

PR #683, opened 2014-10-17
Allow BOSH to load external CPIs

nterry:
Fixed commit history, bumped rugged gem to support development on OSX hosts, and fixed loading of external CPIs


cppforlife:
Closing this PR. micro CLI plugin is on its way out since we are replacing it with bosh-init which would not require modifications to support external CPIs.

PR #725, opened 2015-01-06
Allow openstack configurations that have a single storage AZ but multiple compute AZ's

johnmcteague:
We are utilizing Ceph for our openstack storage and this is being presented as a single AZ from cinder, but we will have multiple AZ's for compute.

As of v1.2776 a bug was fixed that was actually permitting this behaviour, the bug did not pass the servers AZ to the compute api when creating disks. I agree that this should have been fixed, this PR is to explicitly allow storage and compute AZ's not to match by ignoring the AZ of the server and allowing openstack to fall back to the default zone for cinder.

In the public cloud world having this mismatch makes no sense (AZ == datacenter usually and we do not want vm's using storage in a different DC). With openstack in a private datacenter, the types of configuration will vary more, and systems like Ceph manage the availability of volumes on our behalf rather than relying on openstack az's being presented.


cppforlife:
btw story scheduled: https://www.pivotaltracker.com/story/show/85645682

cppforlife:
@johnmcteague can you confirm this feature works as you expected in the latest build?

johnmcteague:
Deployed BOSH on top of Microbosh using v2827. All works as expected, servers and storage are in different zones when I set ignore_server_availability_zone: true

Running with ignore_server_availability_zone: false gave me the desired outcome of an Openstack error.

PR #726, opened 2015-01-14
More explanation about OS image files

jfuerth:
Some extra text explaining stuff we had to dig to discover the first time we built an OS image. Probably worth proofreading to make sure we haven't added misinformation. :)


mariash:
Hi @jfuerth, thanks for the pull request! This is very useful. We merged it into develop branch https://github.com/cloudfoundry/bosh/commit/2806339142f47d1b767e9a32efba9249b10cddcb

Thanks!

PR #734, opened 2015-01-20
Update default mysql engine to 5.5.40a for AWS RDS

epoelke:
The old MySQL engine version (5.5.31) seems to no longer be available as an
option for RDS.  Updated default version to 5.5.40a.


epoelke:
@cppforlife I sent in the CLA yesterday and got a response back that it was good to go.  Don't know what else I need to do..

cppforlife:
@epoelke thanks. we'll pull this in soon: https://www.pivotaltracker.com/story/show/86960650

PR #735, opened 2015-01-20
Copy out the stemcell packages list and add it to stemcell tgz

mmc2004jp:
This PR corresponds to following ticket and add corresponding tests:
- BOSH should copy out the stemcell packages list and add it to the repo
  (https://www.pivotaltracker.com/n/projects/956238/stories/83337650)

Changes:
- Moved 'bosh_dpkg_list' stage (which creates packages list file from 
  Ubuntu env) to later stage, so that the file (dpkg_list.txt) can be copied into stemcell tgz file.
- Added unit tests which check created stemcell's tgz files.
- Modified unit tests for modified classes (stage_collection.rb, build_environment.rb)
	
Note:
- Initially I thought the rspec execution (called in stemcell.rake) should be like below:
  `STEMCELL_TGZ=<tgz-file> bundle exec rspec -fd <specfile>_spec.rb`
and temporary directory should be created and deleted in the global hook (config.before(:all) and config.after(:all) in 'bosh-stemcell/spec/support/stemcell_tgz.rb').
However, I didn't find the best way to pass the variable (the path of the temporarydirectory created in global hook) from global hook to each spec file.
Therefore in this implementation, the path is created in stemcell.rake file and pass the path to global hook/spec files by using environment variable (STEMCELL_TGZ_WORKDIR).
So the rspec execution (called in stemcell.rake) is like below:
  `STEMCELL_TGZ=<tgz-file> STEMCELL_TGZ_WORKDIR=<tmpdir>  bundle exec rspec -fd <specfile>_spec.rb`
Any comment to improve this is appreciated.
- The 'bosh_dpkg_list' stage works for Ubuntu only. Corresponding stage for CentOS has not been added.

### DIFF
--- a/bosh-dev/lib/bosh/dev/tasks/stemcell.rake
+++ b/bosh-dev/lib/bosh/dev/tasks/stemcell.rake
@@ -124,6 +124,9 @@ namespace :stemcell do
 
     sh(environment.stemcell_rspec_command)
 
+    stemcell_tgz_workdir = Dir.mktmpdir('stemcell_tgz')
+    sh(environment.stemcell_tgz_rspec_command(stemcell_tgz_workdir))
+
     mkdir_p('tmp')
     cp(environment.stemcell_file, 'tmp')
   end
--- a/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/build_environment.rb
@@ -51,6 +51,16 @@ module Bosh::Stemcell
       ].join(' ')
     end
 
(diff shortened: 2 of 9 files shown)


jbayer:
@cppforlife the CLA is on-file and this PR can proceed from that perspective. thanks!

mariash:
Hi @mmc2004jp,

Thank you for submitting this pull request. This is very useful. We used some of it, but not everything, since it was changing the stage collection that was significantly updated. We made commit as following: https://github.com/cloudfoundry/bosh/commit/50b8b01fdcd79beb40d0a84f1f251d98546aaa52

Thanks,
Kam & Maria

PR #736, opened 2015-01-22
Bump kernel version in Ubuntu image specs

uzzz:
Hi,

Ubuntu kernel updated again and it led to failed specs when I tried to create a stemcell.


cppforlife:
already done in https://github.com/cloudfoundry/bosh/commit/f07c98c. it's going through our CI.

PR #743, opened 2015-01-29
Update monit to 5.2.5

uzzz:
Hi

I saw a couple pull request been made with monit update and that they were rejected as far as you don't want to force users to use latest features. But here I offer to update monit to the latest 5.2.x minor release as far as there are couple bugs fixed, one of them I experienced personally: I was trying to monitor process, that does not leave PID file. "OK, monit 5.2.4 supports process match by pattern", but it didn't work. After I debugged for a while I opened changelog of monit and found that there was a bug about process match that was fixed in 5.2.5:

`* Fix process match check - when the monitored process failed and was restarted by Monit, Monit didn't recognized  it is  running after the restart and reported  start failure (similar on stop). Thanks to  Kenichi Futatsumori for report and  helping to root cause the problem.`


mkocher:
Is this version of monit agpl?

uzzz:
@mkocher do you mean AGPL license? If so then no, it's GPLv3, as well as 5.2.4 was.

cppforlife:
@uzzz Sounds reasonable. We'll review if in the minor release there were any significant changes. Thanks.

dpb587-pivotal:
Merged - thanks! Should show up in the next or following stemcell.

Related tracker story was [#88572424](https://www.pivotaltracker.com/story/show/88572424).

dpb587-pivotal:
Stemcell v3169 has monit 5.2.5.

PR #744, opened 2015-01-29
Retry AWS API throttling

adamstegman:
When AWS throttles your requests, retry them until they work. We addressed VM creation and ELB deregistration since those are the cases we've had issues with.


adamstegman:
This is easily reproducible if you create or delete 15-20 instances at once.

PR #745, opened 2015-02-02
Add VPC support

mmc2004jp:
Current Vagrantfile for stemcell building VM is written for EC2-Classic environment.
However, those who created AWS account after March 2013 can't use EC2-Classic environment (they can use EC2-VPC environment only).
This PR is to add EC2-VPC support to Vagrantfile for stemcell building VM.


cppforlife:
Merged.

PR #746, opened 2015-02-03
Add note about AWS region.

mmc2004jp:
This is to add a note that US East (Northern Virginia) needs to be used when preparing stemcell building VM.
AMI, keypair and security group are region-specific. So if someone creates keypair and security group on other region, 'vagrant up' will fail.


jbayer:
@cppforlife the CLA is on-file and this PR can proceed from that perspective. thanks!

PR #747, opened 2015-02-04
aws and openstack agent config

mariash:


mariash:
this was merged to develop

PR #749, opened 2015-02-05
Check for user and password also in DAV blobstores

voelzmo:
According to the list in https://github.com/cloudfoundry/bosh/blob/f91fce17854855fb242e8cc24f1d777703f24260/blobstore_client/lib/blobstore_client/client.rb#L4 DAV is a valid option for a blobstore client


voelzmo:
merged with commit https://github.com/cloudfoundry/bosh/commit/5a1a8dfc33543f95c59583b4226e454163575b1c

PR #757, opened 2015-02-20
No need to manually install pg gem

voelzmo:
bundler does a pretty good job doing that once postgresql is installed

### DIFF
--- a/docs/workstation_setup.md
+++ b/docs/workstation_setup.md
@@ -10,9 +10,6 @@ Get mysql libraries (needed by the mysql2 gem):
 
 Get postgresql libraries (needed by the pg gem):
 * `brew install postgresql`
-
-Install pg gem manually by specifying your architecture:
-* `(sudo) env ARCHFLAGS="-arch x86_64" gem install pg -v '0.15.1'`
  
 Get redis:
 * `brew install redis`


karlkfi:
I was having trouble getting bosh to bundle correctly on ruby 2.1.2 today until after manually installing pg. But I didn't seem to need it after upgrading to ruby 2.1.5. 

Which version did you try on?

voelzmo:
I was also confused that I didn't have to install the pg gem and specify the architecture manually anymore - when I wrote the initial version of this setup guide I needed to do this. My current ruby version is 2.1.5 as well, so there this seems not to be an issue.

So shall we keep this in with the hint for people on ruby <2.1.5?

bosh-ci-push-pull:
It'd be nice to know which version of ruby fixes the issue so we can be
more explicit.

IIRC our tests in docker are running on 1.9.3 and 2.1.3. We'll probably
want to upgrade our CI before we say we officially support 2.1.5.

-Karl

karlkfi:
It looks like 2.1.4 is included in the bosh release as of November: https://github.com/cloudfoundry/bosh/commit/2b24dc104db1b297c278bad6e7e0b3f26784add5#diff-dac5e3e96b99077502f016c9dbb90af0

But the CLI is being tested with 1.9.3 & 2.1.2: https://github.com/cloudfoundry/bosh/blob/master/gocd/docker/install-ruby.sh

We've tried bundling without manually installing pg on fresh installs of ruby 1.9.3, 2.1.2, 2.1.3, & 2.1.5. None of them seemed to fail.

We updated the docs to be a bit more detailed and made that step optional: https://github.com/cloudfoundry/bosh/blob/develop/docs/workstation_setup.md

PR #760, opened 2015-02-20
Added resume-from-stage feature for stemcell building

jfuerth:
[#88530374]

This commit also includes a fix for the add_on_exit function,
which was giving undefined variable error because the scripts
are now running with `set -u`.


jfuerth:
We've pushed this to develop: ee9361cd981f9ca6b422bd51e5ea6d420919c62f

PR #761, opened 2015-02-24
Fix log filtering in bosh-director

Amit-PivotalLabs:


jfuerth:
Thanks! This is on develop now as 2440b7689abe48715572d3ea8a267d5edb877249

-Nader and Jonathan

PR #764, opened 2015-03-03
Add a cloud_property 'boot_volume_size' ...

mrwangxc:
which allows user to customize the volume size in MB when 'boot_from_volume' is enabled.
Besides, some Openstack deployments set the flavor.disk of flavors to zero, which cause the following 'create_boot_disk' call fail. This patch can also handle this.

To use it, add this option in the 'cloud_properties' section of a 'resource_pools' instance or the 'compilation' configuration. Use case:

compilation:
  cloud_properties:
    boot_volume_size: 8192
    ......
resource_pools:
  - name: pool1
    cloud_properties:
      boot_volume_size: 32768
      ......
  - name: pool2
    cloud_properties:
      boot_volume_size: 4096
      ......

### DIFF
--- a/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
+++ b/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
@@ -276,7 +276,7 @@ module Bosh::OpenStackCloud
         server_params[:availability_zone] = availability_zone if availability_zone
 
         if @boot_from_volume
-          boot_vol_size = flavor.disk * 1024
+          boot_vol_size = resource_pool['boot_volume_size'] || (flavor.disk * 1024)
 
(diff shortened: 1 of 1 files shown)


mrwangxc:
Hi cfdreddbot,
Thanks for inform me of this.
I have submit a CLA but haven't got any response yet for one month, so I just try to make a commit to see whether I was approved ...
Anyway, I have just submit the CLA again according to your suggestion.

chipchilders:
This contributor has filed an ICLA, so the pull request is OK to be merged. We will get the CLA whitelist updated for him soon.

dpb587-pivotal:
Thanks for the PR! The various CPIs have been extracted into their own repos now (see https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release), so we're closing this particular PR on the core bosh repo. If you're still interested in this boot volume sizing feature for openstack, perhaps bring it up on that repo.

PR #766, opened 2015-03-03
[stemcell builder]: don't fetch release candidate if it is already downloaded

allomov:
Hi, everyone.

Currently I need to rebuild stemcell several times a day. I've noticed that bosh release it fetched each time, so I decided to make this update to enhance speed of stemcell rebuilding.

Best wishes,
Alex L.


dpb587-pivotal:
Closing - we're no longer including a release in the stemcell so this shouldn't be necessary any more. Nice idea though (and sorry the PR sat for so long!)

allomov:
Thank you anyway for going through PRs. It is hard work to keep everything uptodate.

PR #771, opened 2015-03-09
Fix "Too many levels of symbolic links" on vagrant provision remote

jfuerth:
https://www.pivotaltracker.com/story/show/86943394


jfuerth:
We've pushed this to develop: ffe72bb4a269629a45519da7e7b43fab6c71c326

PR #772, opened 2015-03-11
Add support for building non-bootable CentOS 7 stemcell

jfuerth:
Tracker story:
https://bosh-jenkins-artifacts.s3.amazonaws.com/bosh-vagrant-boxes/

We factored out the common bits between the centos 6.5 and 7 specs. This is important enough now, but will become critical when the RHEL changes appear a few commits from now. Let us know if you'd have preferred that we did it differently.


jfuerth:
We've pushed this to develop: 1c0852b05f85d60f3902e5fab0cb84ea3c0ee67a

PR #774, opened 2015-03-16
lets you require director client only without cli

bonzofenix:
Hi Bosh team,
I was trying to use `cli/client/director` without requiring all dependencies in the [cli](https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli.rb).
This speeds up loading when the goal is only to use the director client.

### DIFF
--- a/bosh_cli/lib/cli/errors.rb
+++ b/bosh_cli/lib/cli/errors.rb
@@ -1,53 +1,55 @@
-module Bosh::Cli
-  class CliError < StandardError
-    attr_reader :exit_code
+module Bosh
+  module Cli
+    class CliError < StandardError
+      attr_reader :exit_code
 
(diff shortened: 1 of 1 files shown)


cppforlife:
I would recommend not using it directly, but if you must to make this more maintainable https://github.com/bonzofenix/bosh/blob/master/bosh_cli/lib/cli/errors.rb should be changed to define 

```
module Bosh
  module Cli
```

instead of adding `module Bosh; end` to this file.

bonzofenix:
@cppforlife I changed the commit. 
Why you would not recommend using only the director class only?
Thanks

cppforlife:
Because it's a private implementation detail and can change at any time - not an official API.

cppforlife:
Merged.

PR #778, opened 2015-03-21
Add 'available' option for the cloud_properties of network definition

mrwangxc:
e.g.

networks:
- name: test
  type: manual
  subnets:
  - cloud_properties:
    available:
    - 192.168.100.1 - 192.168.100.20
      ...

When 'available' is given, the available_dynamic_ips is derived from it instead of 'range'.
The reason to introduce this option is as following:
When deploying multiple bosh releases, we often get conflicting vm ip allocations related
to dynamic ips if releases share the same network. It is unnecessary and complex to use
different network for each release.
To make things easy and clear, we use 'available' to isolate multiple releases' address ranges
in one network. And it's very convenient to arrange vm networks of multiple releases when we
put them in one file and use the 'available' option.

### DIFF
--- a/bosh-director/lib/bosh/director/deployment_plan/network_subnet.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/network_subnet.rb
@@ -72,8 +72,21 @@ module Bosh::Director
         first_ip = @range.first(:Objectify => true)
         last_ip = @range.last(:Objectify => true)
 
-        (first_ip.to_i .. last_ip.to_i).each do |ip|
-          @available_dynamic_ips << ip
+        available_ips = safe_property(subnet_spec, "available", :optional => true)
+
+        if !available_ips.nil?
+          each_ip(available_ips) do |ip|
+            if ip < first_ip.to_i || ip > last_ip.to_i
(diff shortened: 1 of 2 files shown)


drnic:
Awesome.

Will/could "available" support a list of ranges (like reserved/static do?)

mrwangxc:
Sure. This option supports a list of ranges, thus gives us full control of multiple releases' network deployment.

drnic:
Lovely

cppforlife:
We are the in the middle of implemented https://github.com/cloudfoundry/bosh-notes/blob/master/global-networking.md on global-net branch. To simplify future migration for users, we will try to pull this PR in once complete global networking work.

jaresty:
merge story [#152973548](https://www.pivotaltracker.com/story/show/152973548)

dpb587-pivotal:
Since we've mostly migrated to using cloud-configs at this point for shared network definitions, and we haven't heard this particular feature requested lately, I'm going to go ahead and close this. Feel free to reopen or add further comments if this is still of interest to you and you'd like to rebase the pull request with the latest code. Thanks!

PR #784, opened 2015-03-25
Regenerate server_objects.rb with latest from pyvmomi.

mmb:
This file was originally generated from the pyvmomi Python SDK by a script
that is no longer in the bosh repo (see commit 4a23c5).

This version was generated using https://github.com/mmb/pyvmomi_to_ruby.

core_types.rb was also regenerated but had no changes.

Passes vSphere lifecycle with one cluster.

Not an official toolsmiths PR (sent as an individual).


cppforlife:
Awesome thanks!

cppforlife:
Merged into develop.

cppforlife:
Updated. Thanks!

PR #794, opened 2015-04-02
permit creation of dev release with version

vito:
this is useful for e.g. creating release candidate tarballs in CI


vito:
woops, just noticed I submitted this against master.

PR #797, opened 2015-04-03
Allow passing --all flag to `bosh cleanup`

monkeyherder:
* Including `--all` deletes all unused releases and stemcells from BOSH, rather than leaving two most recent
 * Helpful for fully cleaning up soon to be killed BOSH


drnic:
Good idea

PR #805, opened 2015-04-17
UAA asymmetric key encryption support for oAuth tokens

rkoster:
This PR adds support for [asymmetric token signing](https://github.com/cloudfoundry/uaa/blob/master/docs/Sysadmin-Guide.rst#token-signing). Most of the implementation has been borrowed from the cloud_controller_ng: [UaaTokenDecoder](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/vcap/uaa_token_decoder.rb) and [UaaVerificationKey](https://github.com/cloudfoundry/cloud_controller_ng/blob/master/lib/vcap/uaa_verification_key.rb).


cppforlife:
Thanks for the contribution. Latest release allows configuration of the symmetric and asymmetric keys.

rkoster:
Nice will have a look

Issue #802, opened 2015-04-13
Allow director behind shared reverse proxy


rkoster:
Currently when referencing a director by hostname, the host portion of the uri is replaced by the [resolved ipaddress](https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/client/director.rb#L31) before [making the request](https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/client/director.rb#L634).

This behaviour prevents the reverse proxy usecase because the Host Header is lost.
I would love to create a PR to fix this behaviour but I wanted to first understand the usecase for the current behaviour.

cppforlife:
It seems that following [commit](https://github.com/cloudfoundry/bosh/commit/55699d7b6f61c51ad0f541882111b92b82c7c0d0) introduced it to reduce number of DNS requests made. I'll ask the guys who wrote it...

rkoster:
@cppforlife figured out the simplest solution which would not affect the current behaviour is adding a Host header. I have created a PR which implements that solution #806.

cppforlife:
Fair enough, I'll close this issue and will update the PR when we pull it in.

PR #806, opened 2015-04-21
Added Host header to http requests

rkoster:
Figured the easiest way to solve the issue raised in https://github.com/cloudfoundry/bosh/issues/802 would be by adding a Host header.


rkoster:
thanks @cppforlife

PR #812, opened 2015-04-29
[stemcell-builder] Fix calling `apt-get distr-upgrade` command in base_apt stage

allomov:
Hey, all.

I tried to build base os image using following command `bundle exec rake stemcell:build_os_image[ubuntu,trusty,base_os_image_path]` and got this error stack trace: https://gist.github.com/allomov/7f51891b0e502638f7cf. I used base os image from this URL: http://s3.amazonaws.com/bosh-os-images/bosh-ubuntu-trusty-os-image.tgz.

Using advices from [this ticket](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1377624) I was able to fix this problem and run `stemcell:build_os_image` without errors.

This PR shows what I needed to do for the fix.

### DIFF
--- a/stemcell_builder/stages/base_apt/apply.sh
+++ b/stemcell_builder/stages/base_apt/apply.sh
@@ -16,4 +16,8 @@ EOS
 
 # Upgrade upstart first, to prevent it from messing up our stubs and starting daemons anyway
 pkg_mgr install upstart
+run_in_chroot $chroot "apt-get download libudev1"
+run_in_chroot $chroot "apt-get download udev"
+run_in_chroot $chroot "dpkg -i --force-confmiss libudev1*.deb"
+run_in_chroot $chroot "dpkg -i --force-confmiss udev_*.deb"
 pkg_mgr dist-upgrade


cppforlife:
@allomov have you tried running os image build on master? we have not seen this failure in our ci.

allomov:
A good question. I used bosh 2858 (this commit b72546515fe71a0fe123a0ef688d93088d19aec8) to run `stemcell:build_os_image` task.

Currently I can't see what could fix this issue in the master for changelog and [diff](https://github.com/cloudfoundry/bosh/compare/b72546515fe71a0fe123a0ef688d93088d19aec8...ac625bbb70d9e58946e0b369cad22e9c64554a24). Do you use the same base os image for tests? Is it possible that this test image has correct version of udev?

cppforlife:
What do you mean by: "Do you use the same base os image for tests?"

OS image is locked down via the json file during the stemcell building.

allomov:
I thought that if you use updated base OS image or smth like this. Could you tell what CI scripts do you use?

allomov:
That's ok if you don't see errors, I'll close the issue. I just needed to make this updates to make it work on my side.

PR #820, opened 2015-05-05
Initial working plugin and tests for a consul event and ttl check forwarder for bosh monitor. 

byllc:
This is a basic plugin for bosh-monitor that forwards NATS messages as events and registers heartbeats as TTL checks in Consul.  It currently works best by using the address of an agent. I've also included a README file in the plugins directory with details about the plugin and stubs for some of the other plugins. 

This submission has been supported by Swisscom and is needed for one of their current efforts. 

CLA Note: My contribution agreements were signed when I started at Stark & Wayne and should have been associated with billchapman@starkandwayne.com or wwchapman@gmail.com .


cppforlife:
@byllc that's an interesting addition. i'm curious how are you guys using it propagated consul events?

byllc:
@cppforlife the idea is that they can be used to help us decide how to act and if a response is needed when something is not in the right state, in general with bosh resurrection on we shouldn't need to intervene  but  the heartbeat events especially have a great deal of information about job nodes (mem,cpu, disk) which is useful upstream of the bosh because it provides a lot of color to our sense and response framework.

byllc:
Should I have synched with the develop branch and PR'd this there?

byllc:
Superceded by : https://github.com/cloudfoundry/bosh/pull/821

PR #821, opened 2015-05-06
Intial working consul event plugin and passing tests

byllc:
This is a basic plugin for bosh-monitor that forwards NATS messages as events and registers heartbeats as TTL checks in Consul. It currently works best by using the address of an agent. I've also included a README file in the plugins directory with details about the plugin and stubs for some of the other plugins.

This submission has been supported by Swisscom and is needed for one of their current efforts.

I originally PR'd this against master and it was suggested that I move it here, against develop.   I've now tested against both branches, Sorry for the duplication. The original PR is here https://github.com/cloudfoundry/bosh/pull/820


cppforlife:
Do you mind pulling out defaults from the code into the spec file (e.g. port, ttl note, etc.)?

byllc:
@cppforlife I'm finding that when I move those defaults out,  my tests now fail because the spec file isn't really in play when I run tests but I  want an assurance that defaults are working. Do you have any suggestions for this situation?  The other plugins don't have very many non trivial defaults so I don't see a good example of how to ensure my defaults from spec.yml are present when I run my tests.

cppforlife:
There is no way to propagate job defaults to your tests. Ideally happy path would be covered by an integration test; however, in this case it might not be worth it given the setup is fairly expensive. Btw as a side effect of moving defaults into the job spec you can remove conditionals from the erb templates:
- if_p("hm.consul_event_forwarder.port")
- if_p("hm.consul_event_forwarder.protocol")
- if_p("hm.consul_event_forwarder.events")
- if_p("hm.consul_event_forwarder.heartbeats_as_alerts")

PR #835, opened 2015-06-03
Add environment variables for proxies

raespinosa:
So that it can connect to vcenters over proxies


raespinosa:
Sorry seems like it is already in cloudfoundry-incubator/bosh-vsphere-cpi-release@3d3932020395a71199bbb506696ce8c06561a9ab

Issue #836, opened 2015-06-03
PowerDNS recursor not working when MicroBosh has multiple IPs


vlerenc:
Hello,

we are having an issue with PowerDNS when using a recursor. It doesn't work if the MicroBosh has multiple IPs, e.g. an additional dynamic/elastic IP. PowerDNS may run the query to the recursor with the wrong IP and/or return the response with the wrong IP, which makes the client reject the response with "reply from unexpected source".

The solution is actually already in PowerDNS: local-address and query-local-address. When defined in the PowerDNS configuration, the issue is gone. Unfortunately, it is currently not possible to set it/configure it properly through the MicroBosh PowerDNS job properties.

Cheers, Vedran

vlerenc:
Please see proposal https://github.com/cloudfoundry/bosh/pull/837.

cppforlife:
Fix was merged. Thanks!

PR #837, opened 2015-06-03
Make PowerDNS local-address and query-local-address configurable via …

vlerenc:
#### Purpose
Make PowerDNS settings ```local-address``` and ```query-local-address``` configurable.

#### Motivation
When the MicroBosh is having multiple IPs (e.g. an additional dynamic/elastic IP) and a DNS recursor is configured, PowerDNS may run the query to the recursor with the wrong IP and/or return the response with the wrong IP, which makes the client reject the response with "reply from unexpected source".

#### Proposal
PowerDNS has settings to control this behavior, but they are not exposed as job properties, so the proposal is to expose them as optional job properties.

#### Changes
* Add two new properties to spec (```local-address``` and ```query-local-address```)
* Apply them in configuration template if set (optionally)

#### Steps to Test or Reproduce
Define PowerDNS recursor, let the MicroBosh have multiple IPs, make a DNS query from a different machine (e.g. with dig @<microbosh> <something>) using not the primary IP of the MicroBosh (e.g. if multiple network adapters are defined not the primary network adapter IP or if a dynamic/elastic IP is defined that instead of the primary network adapter IP).

#### Issues
https://github.com/cloudfoundry/bosh/issues/836

#### CC
@holgerkoser @voelzmo

### DIFF
--- a/release/jobs/powerdns/spec
+++ b/release/jobs/powerdns/spec
@@ -35,6 +35,10 @@ properties:
     default: 2
   dns.recursor:
     description: If recursion is desired, IP address of a recursing nameserver (optional)
+  dns.local_address:
+    description: IP address to which to bind to (optional; useful with multiple IP addresses)
+  dns.query_local_address:
+    description: IP address to use as a source address for sending queries (optional; useful with multiple IP addresses)
   dns.address:
     description: Address of the primary PowerDNS instance
 
--- a/release/jobs/powerdns/templates/pdns.conf.erb
+++ b/release/jobs/powerdns/templates/pdns.conf.erb
@@ -25,6 +25,12 @@ module-dir=/usr/local/lib
 <% if_p('dns.recursor') do |recursor| %>
 recursor=<%= recursor %>
 <% end %>
+<% if_p('dns.local_address') do |local_address| %>
+local-address=<%= local_address %>
+<% end %>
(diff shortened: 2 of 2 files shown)


cppforlife:
lgtm. i wonder if it's better to ask for the network name and template can determine the ip addresses.

vlerenc:
Hello @cppforlife,

I believe you could come up with some automatism to handle multiple network adapters, but this will probably not cover the case with dynamic/elastic IPs of which only the client, but not MicroBosh knows of. For the client it is important that the response IP matches the DNS lookup request IP (via UDP).

Therefore and because it is a less complex and more general solution (in line with PowerDNS), I would rather stick to the PowerDNS settings and just expose them.

Cheers, Vedran

vlerenc:
Hi @cppforlife,

@voelzmo helped me understand your proposal and sure, you are the expert. Such a solution would definitely also solve the issue.

On the other hand, personally I would rather tend to stick to a direct mapping. Less headache later, if assumptions break down.

So what do you propose? Is this PR OK (pro: most direct way/mapping and least complex) or should we prepare a more elaborated proposal where we extract the IP automatically based on the given named network (pro: no explicit IP, but network name only)?

Best regards, Vedran

voelzmo:
Merged into develop

PR #838, opened 2015-06-05
openstack cpi should wait for the server to be deleted

xingzhou:
In the patch, we added the following things:
- wait_resource call when destroying the vm instance
- log error messages during deleting the vm instance
- one new unit test for the code changes

[#95584442]

### DIFF
--- a/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
+++ b/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
@@ -305,6 +305,12 @@ module Bosh::OpenStackCloud
 
           with_openstack { server.destroy }
 
+          begin
+            wait_resource(server, [:terminated, :deleted], :state, true)
+          rescue Bosh::Clouds::CloudError => delete_server_error
+            @logger.warn("Failed to destroy server: #{delete_server_error.inspect}\n#{delete_server_error.backtrace.join('\n')}")
+          end
+
           raise Bosh::Clouds::VMCreationFailed.new(true), e.message
         end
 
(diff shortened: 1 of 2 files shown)


bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
```diff
@@ -305,6 +305,12 @@ def create_vm(agent_id, stemcell_id, resource_pool,
 
           with_openstack { server.destroy }
 
+          begin
+            wait_resource(server, [:terminated, :deleted], :state, true)
+          rescue Bosh::Clouds::CloudError => deleteServerError
```
cppforlife: convention is to use snake case

bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
```diff
@@ -305,6 +305,12 @@ def create_vm(agent_id, stemcell_id, resource_pool,
...
+            @logger.warn("Failed to destroy server: #{deleteServerError.message}")
```
cppforlife: use `#{delete_server_error.inspect}\n#{delete_server_error.backtrace.join("\n")}` to get the class name and the backtrace of the error

bosh_openstack_cpi/spec/unit/create_vm_spec.rb
```diff
@@ -427,7 +427,13 @@ def user_data(unique_name, network_spec, nameserver = nil, openssh = false)
...
 
   context "when cannot create an OpenStack server" do
+    let(:logger) { double }
+
     let(:cloud) do
+      allow(Bosh::Clouds::Config).to receive(:logger).and_return(logger)
+      allow(logger).to receive(:debug)
+      allow(logger).to receive(:info)
+
       c = mock_cloud do |openstack|
```
cppforlife: no need to keep local var `c`

xingzhou:
updated the PR based on @cppforlife 's comments

mariash:
@xingzhou we merged PR with small refactor here: https://github.com/cloudfoundry/bosh/commit/e0e1e05e780021affbcc115c73a6bfdcaa65455d 

Please note, when doing PRs they need to be done against `develop` branch, not `master`.

Thanks.

PR #841, opened 2015-06-10
Task 'create deployment' now includes deployment name

drnic:


cppforlife:
As of bosh-release 255.4 we now track deployment for each Director task (when it makes sense).

PR #842, opened 2015-06-10
Use security-group ID in addition to using security-group NAME

xingzhou:
In this patch, we provide the following changes for aws cpi:

- Support new deployment optional option 'default_security_group_ids'
- Support new deployment optional option 'security_group_ids'
- Usage document changes for these two new options

[#94035130]

### DIFF
--- a/bosh_aws_cpi/spec/unit/helpers_spec.rb
+++ b/bosh_aws_cpi/spec/unit/helpers_spec.rb
@@ -5,7 +5,7 @@ require "spec_helper"
 describe Bosh::AwsCloud::Helpers do
 
 
-  describe "#extract_security_group_names" do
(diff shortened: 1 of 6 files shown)


bosh_aws_cpi/spec/unit/helpers_spec.rb
```diff
@@ -22,5 +30,14 @@ class HelpersTester
       helpers_tester = HelpersTester.new
       expect(helpers_tester.extract_security_group_names(networks_spec)).to match_array(["numero uno", "two to tango"])
     end
+
+    it "should return a sorted list of unique security_group_ids" do
```
mariash: we can probably define HelpersTester above and use in both tests

mariash:
merged

Issue #845, opened 2015-06-11
[centos stemcell] /var/vcap/bosh/bin not in PATH for root user


drnic:
Unlike ubuntu stemcells, on a centos stemcell, `monit` is not in the path for root user.

Can we add `export PATH=$PATH:/var/vcap/bosh/bin` to the profile for the root user? Thanks!

cppforlife:
I've checked on the centos 7 stemcell and root user has /var/vcap/bosh/bin on the $PATH and I'm able to run `monit summary` without problems.

wayneeseguin:
FYI I just issued [Pull Request 849](https://github.com/cloudfoundry/bosh/pull/849) which addresses this issue. 

Also, this was likely an issue on not just CentOS stemcells because of the behavior however I did not verify other stemcell flavors.

voelzmo:
@wayneeseguin fixing that for all stemcells seems odd - one of the most common workflows I do is
```
bosh ssh <machine>
sudo su -
monit summary
```

which works perfectly fine on ubuntu stemcells for me.

cppforlife:
Fix was included in 2999.

PR #849, opened 2015-06-13
Fix: Issue #845

wayneeseguin:
@drnic @cppforlife in regards to issue #845 ::

If you sudo to root as using a login shell:
```sh
sudo su -
```
or
```sh
sudo su -l
```
then monit was not in the path.

If you sudo to root using a non-login shell:
```sh
sudo su
```
then monit was in the path.

The reason for this is that in bash for a login shell it only loads `$HOME/.bash_profile` and for a non-login shell it only loads `$HOME/.bashrc`.  The code only wrote a `/root/.bashrc` and not a `/root/.bash_profile`.

This patch adds the `/root/.bash_profile` file which sources `/root/.bashrc` which adds `$bosh-dir/bin` to `$PATH` and thus `monit` will now be in a root login shell's path.

  ~Wayne

### DIFF
--- a/bosh-stemcell/spec/support/os_image_shared_examples.rb
+++ b/bosh-stemcell/spec/support/os_image_shared_examples.rb
@@ -25,6 +25,10 @@ shared_examples_for 'every OS image' do
       it { should return_exit_status(0) }
     end
 
+    describe command("grep -q 'source /root/.bashrc\n' /root/.bash_profile") do
+      it { should return_exit_status(0) }
+    end
+
     describe command("stat -c %a ~vcap") do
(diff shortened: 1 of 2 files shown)


cppforlife:
Let's just do what Ubuntu does:

```
root@4f2e1302-9a4e-4996-6059-5519e7fa70e3:/home/vcap# cat /root/.profile
# ~/.profile: executed by Bourne-compatible login shells.

if [ "$BASH" ]; then
  if [ -f ~/.bashrc ]; then
    . ~/.bashrc
  fi
fi
```

mariash:
merged into develop

PR #855, opened 2015-06-18
Fix release tarball command to properly handle whitespace in file paths

benmoss:
- We used `!!` to coerce the nil that `system` can return


benmoss:
Looks like Travis was failing already, and the tests that are red don't seem to have anything to do with the changes here.

cppforlife:
In 1.2999.0 of the CLI. Thanks for the contribution!

Issue #803, opened 2015-04-16
Bosh AWS CPI tries to deregister Instances from all ELBs without check


owaism:
While terminating instances the AWS CPI iterates over all ELBs and tries to remove each of the terminating instances.

You can see this in the code:
https://github.com/cloudfoundry/bosh/blob/stable-2922/bosh_aws_cpi/lib/cloud/aws/instance.rb#L92-L99

So if there are many ELBs (more than 25) associated with that AWS Region; you soon end up getting the AWS **Rate Exceeded** error because of too many **mutating** API calls.

Increasing the exponential back off does not solve the problem because the deployment time increases by many folds. For example a deployment that should take only 25 minutes ends up taking 4 and half hours.

The simple fix to this would be to check if the instance is really associated with the ELB before attempting to deregister it. This will result in mostly **read only** API calls and very few **Mutating** API Calls.

cppforlife:
Good point. Referenced info: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html

cppforlife:
AWS CPI no longer deregisters instances from ELBs on termination but rather relies on AWS's behavior to automatically remove instances when they are terminated.

PR #858, opened 2015-06-19
Do not remove instance from ELBs when terminating instances

evanfarrar:
If you have an instance associated with ~25 ELBs this will time out on terminating the instance.

https://github.com/cloudfoundry/bosh/issues/803


evanfarrar:
I can't really reason about this failing build. Monit retry timeout test?

PR #863, opened 2015-06-26
Adding the ability to provide the StrictHostKeyChecking flag to ssh when starting interactive terminals.

jamiemonserrate:
This allows us to ignore the host key when we recreate vms.

Usage:
bosh ssh --strict_host_key_checking no


mariash:
merged into develop

PR #880, opened 2015-07-16
openstack domain (Keystone v3) support)

dkoper:
after applying this, build bosh-openstack-cpi gem, add to vendor cache in https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release/pull/8


cppforlife:
@dkoper Seems like there is failure in the unit tests for the Registry.

cppforlife:
Seems like unit tests are passing so it's time to merge this in.

voelzmo:
Nice, thanks!

PR #892, opened 2015-07-24
[Power Support] Add blobs that need updated config.guess file to S3 blobstore.

allomov:
Hey, all.

I work on porting BOSH to Power microprocessor architectures. The most common problem here is that some components that are compiled during deployment don’t have correct build configuration support for Power. 

This problem is solved easily by updating [config.guess](http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD) files to such packages. You can find out how I generated this blobs using [this script](https://github.com/Altoros/power-bosh-jumpbox-bootstrap/blob/config-updater/roles/binaries-builder/files/scripts/update-configs.sh) (this script uses this [helpers script](https://github.com/Altoros/power-bosh-jumpbox-bootstrap/blob/config-updater/roles/binaries-builder/files/scripts/helpers.sh) for some functions), inputs for this scripts you can find in [this yaml file](https://github.com/Altoros/power-bosh-jumpbox-bootstrap/blob/config-updater/group_vars/config-updater#L7).

Following [this answer in cf-dev](http://lists.cloudfoundry.org/pipermail/cf-dev/2015-July/000642.html) I need to put this blobs to S3 blobstore and post output of `bosh blobs` to this PR. I tried to run `bosh blobs` command and it shows `No blobs to upload` now. But I guess that contents of `release/config/blobs.yml` can demonstrate you updated blobs.

Thank you,
Alex L.

### DIFF
--- a/release/config/blobs.yml
+++ b/release/config/blobs.yml
@@ -8,9 +8,9 @@ postgres/postgres-9.0.3-1.i386.tar.gz:
   sha: b2bde31e720eb219bfdc9f1b784f985ecca2cbfb
   size: 15111387
 postgres/postgresql-9.0.3.tar.gz:
-  object_id: rest/objects/4e4e78bca61e121004e4e7d51d950e04fbd4f2ca2d89
-  sha: c8fb5d1d7930e78211e554bb5751183a8148d0ee
-  size: 18335858
+  object_id: 1ab87cef-91b0-4c37-be96-51f5a97758c6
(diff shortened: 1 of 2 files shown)


allomov:
I've got an answer from Matthew Sykes in cf-dev mailing list that suggested to make a patches in a build phase. I consider that solution with patching is a better choice. Will make PR soon.

PR #893, opened 2015-07-27
Resolve stig boot-loader relevant issues

xingzhou:
[#98978094]
[#98978300]
[#98978312]
[#98978246]

### DIFF
--- a/bosh-stemcell/spec/support/stemcell_centos_rhel_7_shared_examples.rb
+++ b/bosh-stemcell/spec/support/stemcell_centos_rhel_7_shared_examples.rb
@@ -18,6 +18,13 @@ shared_examples_for 'a CentOS 7 or RHEL 7 stemcell' do
       it { should contain 'plymouth.enable=0' }
       it { should_not contain 'xen_blkfront.sda_is_xvda=1'}
       it('single-user mode boot should be disabled (stig: V-38586)') { should_not contain 'single' }
+
+      it('should set the user name and password for grub menu (stig: V-38585)') { should contain 'set superusers=vcap' }
+      it('should set the user name and password for grub menu (stig: V-38585)') { should contain /^password_pbkdf2 vcap grub.pbkdf2.sha512.*/ }
+
+      it('should be of mode 600 (stig: V-38583)') { should be_mode('600') }
+      it('should be owned by root (stig: V-38579)') { should be_owned_by('root') }
+      it('should be grouped into root (stig: V-38581)') { should be_grouped_into('root') }
     end
 
     # GRUB 0.97 configuration (used only on Amazon PV hosts) must have same kernel params as GRUB 2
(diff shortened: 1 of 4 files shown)


stemcell_builder/stages/image_install_grub/apply.sh
```diff
@@ -78,9 +81,20 @@ then
 GRUB_CMDLINE_LINUX="vconsole.keymap=us net.ifnames=0 crashkernel=auto selinux=0 plymouth.enable=0"
 EOF
 
+  pbkdf2_password=`run_in_chroot ${image_mount_point} "echo -e '${random_password}\n${random_password}' | grub2-mkpasswd-pbkdf2 | grep -Eo 'grub.pbkdf2.sha512.*'"`
```
mariash: @xingzhou could you please put comments why we configure password for vcap user?

stemcell_builder/stages/image_install_grub/apply.sh
```diff
@@ -156,5 +170,12 @@ else
   exit 2
 fi
 
+md5_password=`run_in_chroot ${image_mount_point} "if which grub-md5-crypt > /dev/null;then echo -e '${random_password}\n${random_password}' | grub-md5-crypt | grep -e '^\\$1\\$.*'; fi"`
```
mariash: @xingzhou and also here:

1) why we need to check if grub-md5-crypt is present, can we just call it?
2) lets put password line when we writing template above (line  150,162)
3) always chmod/chown grub.conf

stemcell_builder/stages/image_install_grub/apply.sh
xingzhou: @mariash, at present, we can not prevent user from editing the startup menu. In this situation, we generated a random password which we do not even know what the password exactly is, under this situation, the start up menu will be protected.

cppforlife: @xingzhou Maria meant to put that comment in the code.

stemcell_builder/stages/image_install_grub/apply.sh
xingzhou: @mariash, 
(1). grub-md5-crypt is there when grub V1 is installed. in CentOS 7 system, we only install grub V2, which will use another method(edit the 00_header config file), so in CentOS 7, grub-md5-cyprt is not there.

(2). I don't catch you on the comment, would you please make some detail description? thanks

(3). for system installed grub V2, the grub config file is /boot/grub2/grub.cfg not /boot/grub/grub.conf, so in this case, we don't chmod/chown grub.conf

bosh-stemcell/spec/support/stemcell_centos_rhel_7_shared_examples.rb
```diff
@@ -13,16 +13,20 @@
       it { should contain 'selinux=0' }
       it { should contain 'plymouth.enable=0' }
       it { should_not contain 'xen_blkfront.sda_is_xvda=1'}
+
       # single-user mode boot should be disabled (stig: V-38586)
       it { should_not contain 'single' }
-    end
 
-    # GRUB 0.97 configuration (used only on Amazon PV hosts) must have same kernel params as GRUB 2
-    describe file('/boot/grub/grub.conf') do
```
mariash: why was this test removed?

xingzhou:
@mariash, for CentOS and RHEL, they are using grub2 as the grub system, which does not need grub.conf file, so we remove the test cases to test the content of grub.conf;(we removed the grub.conf generation code for CentOS and RHEL in the script)

cppforlife:
@xingzhou after looking over these changes I think there was a reason why grub.conf was being generated as comment states:

> GRUB 0.97 configuration (used only on Amazon PV hosts) must have same kernel params as GRUB 2

i think we should bring it back.

PR #895, opened 2015-07-27
[Power Support] Update build configuration files to support IBM Power platform (for ruby and libpq packages)

allomov:
This PR is made to provide IBM Power support to BOSH.

You can find details of the issue in this PR https://github.com/cloudfoundry/bosh/pull/892. I've come to the idea of this changes thanking to Matthew Sykes (and his post in [cf-dev](http://lists.cloudfoundry.org/pipermail/cf-dev/2015-July/000911.html)).

### DIFF
--- a/release/packages/libpq/pre_packaging
+++ b/release/packages/libpq/pre_packaging
@@ -3,13 +3,6 @@
 # This pre_packaging script runs only for IBM Power platform
 # It updates old versions of build configuration (config.guess and config.sub files)
 
-main(){
-  if [ "`uname -m`" == "ppc64le" ]; then
-    cd ${BUILD_DIR}
-    update_build_config postgres/postgresql-9.0.3.tar.gz
-  fi  
-}
-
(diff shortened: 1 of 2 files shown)


cppforlife:
Hey @allomov. We typically want to avoid usage of pre_packaging scripts (since they run in a deterministic environment) and requiring internet access. For this PR it should be really simple to resolve it. Just commit files being downloaded into the src directory and then replace config files in the appropriate `packaging` scripts if ppc64le.

allomov:
@cppforlife I fully agree with you about using `pre_packaging` scripts. I used `pre_packaging` script to  show you the source of `config.guess` file. Also I was not sure about where to place `config.guess` file. Is it ok to place it to `src/common/config.guess`?

cppforlife:
I would just create `src/config/config.guess` and `src/config/config.sub`.

allomov:
Ok. Thank you for advise.

PR #900, opened 2015-07-30
optional check for openstack volume service

ronakbanka:
Just a optional openstack volume service check , the requirement for this is because openstack deployment sometimes is without shared storage . 

So if there is no volume endpoint on openstack , bosh openstack cpi initialization will fail after retrying . 
By default this value is true so it won't affect normal openstack deployments with volume endpoint.

### DIFF
--- a/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
+++ b/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
@@ -47,7 +47,7 @@ module Bosh::OpenStackCloud
       @boot_from_volume = @openstack_properties["boot_from_volume"]
       @boot_volume_cloud_properties = @openstack_properties["boot_volume_cloud_properties"] || {}
       @use_dhcp = @openstack_properties.fetch('use_dhcp', true)
-      @use_openstack_volume_service = @openstack_properties["use_openstack_volume_service"]
+      @use_openstack_volume_service = @openstack_properties.fetch('use_openstack_volume_service', true)
 
       unless @openstack_properties['auth_url'].match(/\/tokens$/)
         @openstack_properties['auth_url'] = @openstack_properties['auth_url'] + '/tokens'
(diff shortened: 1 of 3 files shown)


cppforlife:
Unfortunetly currently 'bosh micro' does not use defaults from a spec. bosh-init does on the other hand. So to stay backwards compatible (also would fix the tests) please add a default also in code, just like we did for use_dhcp.

ronakbanka:
@cppforlife done

cppforlife:
Just to clarify: 

In your case OpenStack installation does not have volume management, so configure BOSH is only used to manage VMs without persistent disks? Are you using external DB and blobstore for your Director?

ronakbanka:
@cppforlife , we have a hybrid deployment where directors are on vsphere and managing vm on openstack.

ronakbanka:
@cppforlife When can we expect merge of this branch ?? Any timeline? 

If the issue is with the CLA at that time but now it is no more. You can refer below pull request.
https://github.com/cloudfoundry/loggregator/pull/56

cppforlife:
We will be trying to pull it in this week.

calebamiles:
We have attempted to address this issue by moving the connection to the volume service to CPI methods which require it. If 25c7b41ef6b9d36b17bd636ac71f8cf8e92fa302 does not address your issue, please let us know!

@calebamiles

BOSH CPI Team

ronakbanka:
@calebamiles I have tested this with bosh release and it works fine. 

In the commit history on the bosh-openstack-cpi-release 
https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release/commits/c96f7fc505755d11ab8fd52256afd678af4c1c4c

current release version is 10 on bosh.io
http://bosh.io/releases/github.com/cloudfoundry-incubator/bosh-openstack-cpi-release?version=10

which points to above branch, so can't test it as of now with bosh-init.

@cppforlife when the next release  for bosh-openstack-cpi will be there on bosh.io ??

cppforlife:
@ronakbanka we have a tiny problem with our openstack tenant right now (lifecycle tests are failing: https://main.bosh-ci.cf-app.com/pipelines/bosh-openstack-cpi). we will be cleaning out that env tomorrow, so I expect to have a new release soon.

cppforlife:
v11 of openstack-cpi-release includes the change.

PR #903, opened 2015-08-03
Add has_disk? to openstack CPI

xingzhou:
[#74166312]

### DIFF
--- a/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
+++ b/bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
@@ -491,6 +491,20 @@ module Bosh::OpenStackCloud
     end
 
     ##
+    # Check whether an OpenStack volume exists or not
+    #
+    # @param [String] disk_id OpenStack volume UUID
+    # @return [bool] whether the specific disk is there or not
+    def has_disk?(disk_id)
+      with_thread_name("has_disk?(#{disk_id})") do
+        @logger.info("Check the presence of disk with id `#{disk_id}'...")
+        volume = with_openstack { @openstack.volumes.get(disk_id) }
+
+        !volume.nil?
+      end
+    end
+
+    ##
     # Deletes an OpenStack volume
     #
     # @param [String] disk_id OpenStack volume UUID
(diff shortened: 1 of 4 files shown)


bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
```diff
@@ -550,7 +564,7 @@ def detach_disk(server_id, disk_id)
         cloud_error("Server `#{server_id}' not found") unless server
 
         volume = with_openstack { @openstack.volumes.get(disk_id) }
-        cloud_error("Volume `#{disk_id}' not found") unless volume
+        return if volume.nil?
```
ronakbanka: How about logging the cloud_error here, in case volume is not found?

bosh_openstack_cpi/spec/unit/has_disk_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
+
+describe Bosh::OpenStackCloud::Cloud do
+  describe 'has_disk?' do
+      context 'when disk is found' do
```
cppforlife: indention is funky.

bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
cppforlife: Agreed.

bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
```diff
@@ -550,7 +564,8 @@ def detach_disk(server_id, disk_id)
         cloud_error("Server `#{server_id}' not found") unless server
 
         volume = with_openstack { @openstack.volumes.get(disk_id) }
-        cloud_error("Volume `#{disk_id}' not found") unless volume
+        @logger.info("Disk `#{disk_id}' not found while trying to detach it from vm `#{server_id}'...")
```
ronakbanka: Logging before volume check ??

xingzhou: yeap, a bug, fixing it now

bosh_openstack_cpi/lib/cloud/openstack/cloud.rb
```diff
@@ -550,7 +564,10 @@ def detach_disk(server_id, disk_id)
         cloud_error("Server `#{server_id}' not found") unless server
 
         volume = with_openstack { @openstack.volumes.get(disk_id) }
-        cloud_error("Volume `#{disk_id}' not found") unless volume
+        if volume.nil?
+          @logger.info("Disk `#{disk_id}' not found while trying to detach it from vm `#{server_id}'...")
+          return
+        end
```
mariash: We should update agent settings anyway at the end. Something like: 

```
if volume.nil?
  @logger.info("Disk `#{disk_id}' not found while trying to detach it from vm `#{server_id}'...")
else
   detach_volume(server, volume)
end
```

PR #908, opened 2015-08-05
Add has_disk? to aws cpi

xingzhou:
[#74166188]

### DIFF
--- a/bosh_aws_cpi/lib/cloud/aws/cloud.rb
+++ b/bosh_aws_cpi/lib/cloud/aws/cloud.rb
@@ -227,6 +227,18 @@ module Bosh::AwsCloud
     end
 
     ##
+    # Check whether an OpenStack volume exists or not
+    #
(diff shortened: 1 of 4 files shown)


bosh_aws_cpi/lib/cloud/aws/cloud.rb
```diff
@@ -304,6 +316,11 @@ def detach_disk(instance_id, disk_id)
         instance = @ec2.instances[instance_id]
         volume = @ec2.volumes[disk_id]
 
+        unless volume.exists?
```
mariash: @xingzhou we should update agent settings anyway, lets move the check below it.

mariash: Lets instead update settings after detach to be consistent with openstack.

```
if volume.exists?
  detach_ebs_volume(instance, volume)
else
  @logger...
end

update_agent_settings...
```

PR #912, opened 2015-08-07
Adding logic to build Photon stemcell

dthaluru:


cppforlife:
Can you add a section similar to https://github.com/cloudfoundry/bosh/tree/master/bosh-stemcell#special-requirements-for-building-a-rhel-os-image about Photon ISO and where to get it from.

dthaluru:
Sure

PR #919, opened 2015-08-10
    Adding logic to build Photon stemcell

dthaluru:


dthaluru:
Signed-off-by: dthaluru@vmware.com

stemcell_builder/stages/bosh_users/apply.sh
```diff
@@ -18,8 +18,8 @@ run_in_chroot $chroot "
 groupadd --system admin
 useradd -m --comment 'BOSH System User' vcap
 chmod 755 ~vcap
-echo \"vcap:${bosh_users_password}\" | chpasswd
-echo \"root:${bosh_users_password}\" | chpasswd
+echo "vcap:${bosh_users_password}" | chpasswd
```
mariash: why escaping was removed? that seems like it might break this script

stemcell_builder/stages/system_grub/apply.sh
```diff
@@ -9,10 +9,10 @@ source $base_dir/lib/prelude_apply.bash
 
 # Install grub or grub2 (check existence of classic grub package first because Ubuntu Trusty has a transitional grub2 dummy package)
 
-if pkg_exists grub; then
-  pkg_mgr install grub
-elif pkg_exists grub2; then
+if pkg_exists grub2; then
```
mariash: based on the comment above, it might break Ubuntu Trusty stemcells

dthaluru: Okay I can revert this change. Previous version of Photon used to come with grub and grub2 packages. To avoid installing grub package I have changed the order. As latest version TP2 of Photon only comes with grub2, so it should not affect building Photon stemcell.

stemcell_builder/stages/bosh_users/apply.sh
dthaluru: With escape character it is not able to change password in Photon. I verified in Fedora and Ubuntu with out escape character and it works. Let me know if I have to revert this change.

dthaluru: I tried changing password manually with escape character on Ubuntu trusty and it did not work.  Looks like this code is never seemed to be working correctly.

dthaluru: I am wrong. I have tested again with escape character, it seems to be working. I am going to revert this change.

dthaluru:
@mariash  I have revereted the changes you suggested me.

PR #924, opened 2015-08-17
Update bosh-registry to support Azure without plugin settings.

AbelHu:


AbelHu:
Just a test and close it.

PR #932, opened 2015-08-19
fix the bug of not using $chroot/etc/group to check permission

zhang-hua:
we occasionally found this bug of not checking /etc/group file
permission under chroot after os image was built. The reason is
File.stat method won't get the chroot context in Rspec setup by
os_image.rb. It will use the /etc/group file of stemcell build
machine's OS, not the stemcell's. But because both of the files
has the same mode value 0644, the problem does not come out in
this case. it will fail the other cases when the files in
$chroot and OS have different mode value.


cppforlife:
@zhang-hua can we add a custom matcher to check for permissions. also since we know exact permissions of the file we should just check for equality.

zhang-hua:
@cppforlife , resubmitted this PR as you suggested. Thanks!

PR #937, opened 2015-08-21
Update monit to 5.14

CAFxX:
Monit 5.2.4 is over 4 years old and does not have support for some features we need (primary use case: running custom check scripts for monitoring clusters)


CAFxX:
Any chance oh having this pulled in?

tedsuo:
Sorry that we have taken so long to reply.  Monit is out of date.  But because we plan to move away from monit to a better cross platform interface, we want to keep the monit version stable to enable a smoother transition for Bosh community.  As a result, monit upgrade requests have to-date been declined.

Please see the closing of this pull request for more details: https://github.com/cloudfoundry/bosh/pull/590
Currently we are scheduled for an upgrade to monit 5.2.5.  You can track the progress of that story here: https://www.pivotaltracker.com/story/show/88572424

As an alternative, I suggest you enquire on the bosh or monit mailing list about the best way to emulate the behavior you need with the current version of monit.

Thank you,
Ted
@tedsuo

CAFxX:
@tedsuo understood, but then I have a few followup questions: what will the "better cross platform interface" be and when it is slated to be implemented? Also, assuming we emulate the behavior we need, what are the odds that our solution will keep working once this new interface comes around?

tedsuo:
We plan to fully support migrating from monit 5.2 to the new interface, so as long as you are targeting that as your api, you should be OK. @cppforlife can probably provide more details about the new interface. I don't believe there is a concrete proposal yet.  We just know that we don't want direct access to monit to be our permanent interface for process monitoring.  Possibly a pull request against https://github.com/cloudfoundry/bosh-notes would get things moving?  The necessary shleps are researching all of the api's we expect to support; deriving a universal api; and deriving a migration strategy from monit 5.2.

BTW, here is the pull-request for the current move to monit 5.2.5: https://github.com/cloudfoundry/bosh/pull/743

sykesm:
@tedsuo, @cppforlife: Our environment is hitting a number of bugs related to monit that appear to have been fixed in more recent levels. We have no interest in using new features but we do need to stop working around the bugs.

Right now, monit is at the core of what bosh does, so, without a proposal or replacement in hand, why should this pull request get stalled?

I'd also like to point out that PR #743 has been open for more than seven months without integration. That's quite troubling given we're talking about such a simple change.

cppforlife:
I've asked legal to let us know if it's ok to bump to a newer version. 

@sykesm which bugs (besides check process) are you running into and which versions of monit fix it? Would bumping to 5.2.5 solve your problems?

Code simplicity of this change does not reflect effort needed to confirm that update of monit does not break all kinds of other things, bosh-agent and releases that do non-pid checking come to expect of monit.

sykesm:
@cppforlife: I'm following up on this side to get the precise problems.

As to effort and complexity, I saw the story was a 1 point so I made an assumption.

sykesm:
@cppforlife: The problems we hit are either the "check process" bug that was fixed in 5.2.5 or the bug fixed in 5.5 that was described as:

> * In the case that the process start/restart execution failed, monit kept "Execution failed" flag even if the process was recovered later (for example it was starting slowly or manually recovered).

In either case, an update is needed to resolve the issue. Ideally we'd see something at 5.5 or better.

@CAFxX: Apologies for hijacking your PR.

drnic:
I started experiencing this quirk I think with the docker-boshrelease on bosh-lite/garden. The `docker` process is running just fine and there are no errors in the log - `monit restart docker` makes it work; but `monit summary` originally thought docker failed (and `bosh vms` thinks its failing).

cppforlife:
It sounds like for now we can only update to 5.2.5.

cppforlife:
@sykesm since we are not seeing this issue in our environments, it would be great if you guys can make a stemcell with 5.2.5 monit and confirm that it resolves your problem in your environments and does not show any other side effects. once confirmed i think we will be more comfortable bumping it for everyone.

/cc @maximilien

mattcui:
@CAFxX Did you verify if all jobs work well (like CC job) after upgrading to monit 5.14 or later version? Thanks.

fabianschwarzfritz:
It seems to like there are no plans to update monit to a more recent version in the near future. The version that's currently deployed is quite outdated.

In the current version I'm missing a feature provided in monit version 5.9.0 (https://mmonit.com/monit/changes/ --> 5.9.0). It is the *check program* feature that would enable us to do a more sophisticated check via bash scripts if our service is up and running.

What's the status of this ticket? Will there be any update of monit to a never version in the near future?

liuweichu:
Hello, it's been more than one year since the ticket is opened.
I wonder what is the status of the monit replacement?

@tedsuo @cppforlife @sykesm

CAFxX:
We keep hitting small quirks and bugs in the 5+ years old version of monit that is present in the stemcells. Finding relevant documentation and discussions is a pain. Even the source code from those versions is not in the monit git repo.

Wouldn't it make more sense, instead of keeping deferring the updates because we don't want to support newer features in this AFAICT-not-even-planned-yet process monitoring interface, to

1. define the subset of monit features and semantics that will be supported when the new interface comes around
2. update monit, describing in the docs that any feature outside of the list defined in 1. is unsupported and will stop working in the future

One of the risks of sticking with a specific version instead of a list of features is that when the time will come to define the process monitoring interface it will have to model also the quirks/bugs we have been working around.

I would also be interested in understanding the need to provide a OS-agnostic monitoring interface. Most of our monit scripts call bash _ctl scripts. What are the chances of a bash _ctl script running unmodified on Windows?

dpb587-pivotal:
Closing - stemcell building-related code has been moved to [cloudfoundry/boxh-linux-stemcell-builder](https://github.com/cloudfoundry/bosh-linux-stemcell-builder), and there are no upcoming plans to bump monit due to reasons mentioned here and in #1295.

PR #938, opened 2015-08-21
Add iftop and iotop to stemcells

CAFxX:


cppforlife:
I'm not sure we should include these by default in the stemcells, since we are trying to keep stemcells as strip down as possible. I would recommend putting them into a release and collocating it. Alternatively if env has internet access, apt-get install on the fly should also work.

CAFxX:
I guess the reason is the same one for having things like `mg`, `tcpdump` and `htop`. Those could go in a release as well.

drnic:
Perhaps install them via https://github.com/cloudfoundry-community/bosh-root-env

cppforlife:
The existing tools were already included long time ago and we cannot remove them for backwards compatibility. There are several security guidelines that different companies have about including debugging tooling (like nc, etc.) that do not allow adding more tools for production envs. Overtime we want to be completely passing even the strictest guidelines, hence do not want to pick up extra executables.

PR #941, opened 2015-08-27
Allow more AWS credential sources

drocamor:
This PR contains changes that allow BOSH to be configured to get AWS credentials from different places. The primary driver for this is to allow BOSH to use IAM Roles via EC2 instance profiles so it includes a change that enables BOSH users to set IAM instance profiles for their resource pools.

BOSH users can set different credential sources in their configuration. This is how it changes BOSH configuration:

* `credentials_source` (optional)
  Where to get AWS credentials. This can be set to `static` for to use an `access_key_id` and `secret_access_key` or `env_or_profile` to get the credentials from environment variables or an EC2 instance profile. Defaults to `static` if not set
* `access_key_id` (optional, required when `credentials_source` is `static`)
  AWS IAM user access key
* `secret_access_key` (optional, required when `credentials_source` is `static`)
  AWS IAM secret access key

The default behavior matches what BOSH does today.

### DIFF
--- a/blobstore_client/README.md
+++ b/blobstore_client/README.md
@@ -19,7 +19,7 @@ $ blobstore_client_console -p local -c config/local.yml.example
 => Welcome to BOSH blobstore client console
 You can use 'bsc' to access blobstore client methods
 > bsc.create("this is a test blob")
-=> "ef00746b-21ec-4473-a888-bf257cb7ea21" 
+=> "ef00746b-21ec-4473-a888-bf257cb7ea21"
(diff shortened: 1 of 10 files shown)


cppforlife:
@drocamor one rubocop test failed on travis.

drocamor:
Think I may have made a mistake here by branching from master rather than develop. Any tips on fixing this?

Thanks!

cppforlife:
It's fine we'll rebase off of develop.

PR #947, opened 2015-09-09
Add a --terse option to `bosh stemcells` list view

Jonty:
The current output of `bosh stemcells` is hard to accurately parse from
scripts that invoke it. This commit implements the `--terse` option
currently only found on `bosh properties`.

One line is outputted per stemcell with zero lines being output if no
stemcells are available. The terse view lacks the total size visible in
the table view.

### DIFF
--- a/bosh_cli/lib/cli/commands/stemcell.rb
+++ b/bosh_cli/lib/cli/commands/stemcell.rb
@@ -84,9 +84,10 @@ module Bosh::Cli
 
     usage 'stemcells'
     desc 'Show the list of available stemcells'
+    option "--terse", "easy to parse output"
     def list
+      terse = options[:terse]
       auth_required
-      show_current_state
 
       stemcells = director.list_stemcells.sort do |sc1, sc2|
         if sc1['name'] == sc2['name']
@@ -96,21 +97,29 @@ module Bosh::Cli
         end
       end
 
-      err('No stemcells') if stemcells.empty?
-
-      stemcells_table = table do |t|
-        t.headings = 'Name', 'OS', 'Version', 'CID'
-        stemcells.each do |sc|
-          t << get_stemcell_table_record(sc)
+      if terse
+        stemcells.each do |sc| 
+          row = get_stemcell_table_record(sc)
+          say(row.join("\t"))
         end
-      end
+      else
+        show_current_state
+        err('No stemcells') if stemcells.empty?
 
-      nl
-      say(stemcells_table)
-      nl
-      say('(*) Currently in-use')
-      nl
-      say('Stemcells total: %d' % stemcells.size)
+        stemcells_table = table do |t|
+          t.headings = 'Name', 'OS', 'Version', 'CID'
+          stemcells.each do |sc|
+            t << get_stemcell_table_record(sc)
+          end
+        end
+
+        nl
+        say(stemcells_table)
+        nl
+        say('(*) Currently in-use')
+        nl
+        say('Stemcells total: %d' % stemcells.size)
+      end
     end
 
     usage 'public stemcells'


Jonty:
The corporate CLA has now been correctly filed. Can this pull request be considered now?

cppforlife:
@Jonty to understand better, why do you parse the stemcells output of the CLI?

Jonty:
@cppforlife We are automating the removal of old stemcells from storage. Currently parsing ````bosh stemcells```` involves some horrendous grep and awk and we'd like to make this more robust.

bosh-ci-push-pull:
There is a `bosh cleanup` command that removes unused stemcells and
releases. Does that solve your use case?

Jonty:
@bosh-ci-push-pull No, that does not solve our specific use case.

If this pull request was merged I was planning to add machine-parsable output to many of the bosh commands as we have other instances where we need to process output from bosh commands in automated environments.

simonjohansson:
@Jonty 

There is a client that we use for these kind of tasks. (Uploading releases, stemcells, deleting releases, stemcells, deploying etc.)

```
require 'cli' # Super shitty name for a client :D
bosh_client = Bosh::Cli::Client::Director.new("https://hostname.com:25555")
bosh_client.login("admin", "password")

stemcells = bosh_client.list_stemcells
unused = stemcells.select {|stemcell| stemcell["deployments"] == []}

unused.sort_by {|stemcell| stemcell["version"].to_f}.each do |stemcell|
  bosh_client.delete_stemcell(stemcell['name'], stemcell['version'])
end
```

https://rubygems.org/gems/bosh_cli/versions/1.3071.0

Jonty:
@simonjohansson The machine-parsable output is generally used by people who are comfortable in bash, but not in ruby, hence this change.

Given ````--terse```` already exists for one of the other commands, is there a particular reason why replicating this functionality for other commands is not wanted?

voelzmo:
@simonjohansson that looks pretty much like what `bosh cleanup` is actually doing: https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/commands/maintenance.rb#L54-L81

@Jonty I can see that having machine-parseable output for many bosh commands makes sense - especially for cases in which you are building functionality that bosh doesn't have and doesn't want to include.

In this case, however, I don't really see how `bosh cleanup` doesn't achieve what you want. Working with the output of a CLI is always error-prone and if you could rather switch to a native CLI command, I'd rather do that.

This is just my 2cents, @cppforlife feel free to jump in if you have a different opinion here.

cppforlife:
I'd like to build out some higher level functionality in bosh CLI that most people need when operating bosh. `bosh cleanup` is a good example of that where most bosh operators need to clean up stemcells.

what functionality is missing from cleanup command?

simonjohansson:
@cppforlife 

I agree that functionality with the bosh CLI for most normal tasks is great, but sometimes there is a need to go outside of the "normal", and then ``--terse`` would be really nifty so one does not have to write awk's/sed's of doom.

Surely there is room for both?

Jonty:
@cppforlife As I have said this commit was purely the first in a series as I intended to add --terse as an option for all bosh commands - we parse the output of many of them, not just this one.

If you wish to talk about extending the cleanup command I'm happy to do so in a new issue, but continuing the discussion here derails from getting this change merged.

Are the CF maintainers actively against adding machine-readable output?

keymon:
I agree that a cleanup with additional options would help here. In this case a good option to the clean up is specify how many releases to keep (in this case `--all` is misleading, it keeps 2 releases). We could add `--keep NUM` to keep latest NUM releases.

But on the other hand, automating bosh with shell scripts a quite useful thing, for quick prototyping, etc. We all know the benefits of a quick shell script :). The current output of bosh commands makes it really difficult to handle in scripting. 

Other example of difficult to parse output is `bosh releases` output, and upload a release to finally find out that has been already uploaded is really slow.

The cloudfoundry client has a 'curl' option, which is extremely useful for debugging and also scripting.

I would say that +1 for a `--terse` option to ease the management.

cppforlife:
> I agree that a cleanup with additional options would help here. In this case a good option to the clean up is specify how many releases to keep (in this case --all is misleading, it keeps 2 releases). We could add --keep NUM to keep latest NUM releases.

`--all` removes all unused releases. without `--all` (which is a default) it only keeps last 2 unused releases/stemcells.

cppforlife:
> The current output of bosh commands makes it really difficult to handle in scripting.

that's somewhat purposeful to prevent people reinventing the wheel; or inventing it and not contributing it to bosh CLI itself.

> Other example of difficult to parse output is bosh releases output, and upload a release to finally find out that has been already uploaded is really slow.

This is an excellent example why we do not want to introduce --terse for this kind of thing =) Goal is for most of the commands to be idempotent including `bosh upload release`. We actually have a story coming up to improve upload release command to be faster in a noop case: https://www.pivotaltracker.com/story/show/101726210.

1stvamp:
> that's somewhat purposeful to prevent people reinventing the wheel; or inventing it and not contributing it to bosh CLI itself.

Making something difficult for users of a project is not the way to get them to contribute back.

cppforlife:
> Making something difficult for users of a project is not the way to get them to contribute back.

imho it's better to understand the underlying issue user is trying to solve (e.g. cleanup stemcells/releases) and work through it with them, than to encourage a way to write scripts to automate bosh.

cppforlife:
@Jonty 

> If you wish to talk about extending the cleanup command I'm happy to do so in a new issue, but continuing the discussion here derails from getting this change merged.

Sure, we would be interested in a better cleanup command. 

> Are the bosh maintainers actively against adding machine-readable output?

I'd like to first find out how you are planning to use these APIs, and which commands potentially can be improved to avoid extra wrappers, etc.

mtekel:
Hi,

having terse output on stemcells is useful whenever you want to automate some process around stemcell management. E.g. imagine you want to sync what stemcells you have in multiple bosh instances. And sync that against some remote stemcell repository (e.g. I add new stemcell to that repo, I want to have it added to all my boshes, I remove old stemcell from repo, I want it removed from all my bosh instances). 

Is bosh-cli planning to implement that kind of functionality? Or instead, is the intention of bosh-cli to never provide any machine parsable output, so that you can explore every single possible use-case? Or is the intention here to force people using ruby instead? In that case, you should remove --terse output from other CLI commands that currently have it...

keymon:
+ @mtekel 

To support the argumentation of parseable output: I think that BOSH cli, as a command line tool where its users are unix/linux operators, it should enable and empower them by subscribing the [Unix Command Line philosophy](http://www.tldp.org/LDP/GNU-Linux-Tools-Summary/html/c1089.htm).

I am not saying that we don't implement full out of the box use cases, but parseable output is desired and not so difficult to implement.

cppforlife:
@mtekel 

> E.g. imagine you want to sync what stemcells you have in multiple bosh instances.

To address common scenarios like your example, we are trying to make all commands idempotent so that you can just write bosh upload stemcell X --skip-if-exists (in future skip-if-exists flag wouldn't even be necessary) in a loop. You shouldnt have to check against a list of stemcells that already exist on a director.

>  ...so that you can explore every single possible use-case?

It's ideal to know what people are trying to do because I have seen so many duplicate bash scripts doing same exact thing. Most likely if you are trying to write a script around bosh cli, at least few other people already done it.

> ...Or is the intention here to force people using ruby instead?

nope.

> In that case, you should remove --terse output from other CLI commands that currently have it...

There is one command doing it and it was changed in 2012.

Let me reiterate from above: My goal here is not to make your life harder than necessary, but to learn about use-cases people write scripts for and make them a product feature if that is something a lot of people are trying to do.

mtekel:
Hi,

thanks for your perspective. This switch is specifically useful to enable lots of these different, too uncommon to be implemented in client, scenarios. For example, in my case I would also check if the stemcells are in sync between environments (e.g. I'd implement a nagios check to do this - as stemcell uploads can fail, cloud providers can delete stemcells, bosh DBs can get corrupt/deleted by accident on different environments etc.; default --skip-if-exists won't help me here). 

When are you going to implement nagios support in the CLI? Do we need to get enough people to vote for it? Would you ever consider implementing a feature, that's not a final end use case/scenario, but enables lots of people implement their own specific, but very different niche use cases?

cppforlife:
Given the demand for this feature let's pull this (and similar) PRs in. I still hope that other use cases will be shared with the community even if there is an easy way to automate through some bash scripting. 

We have also recently documented the Director API in its basic form: https://bosh.io/docs/director-api-v1.html.

Jonty:
It's been a year since this PR was opened, and almost as long since it was marked as `ready-to-pull-in`. Is it likely to be merged, or should I close it?

cppforlife:
Apologies for the delay on making this happen. We are on the verge of releasing rewritten bosh CLI (https://github.com/cloudfoundry/bosh-cli) that specifically addresses variety of problems collected from community such as this one. It might be best to close this PR at this point - we probably won't backport this sort of functionality to ruby CLI.

We've made table formatting a first class option so that all commands are script friendly (json and cut/grep). Here is an example of it in action:

```
$ bosh stemcells|cat
bosh-aws-xen-centos-7-go_agent         	3232.custom*	centos-7     	ami-3849af55
bosh-aws-xen-hvm-centos-7-go_agent     	3232        	centos-7     	ami-ab3822c1
bosh-aws-xen-hvm-ubuntu-trusty-go_agent	3280*       	ubuntu-trusty	ami-d9a6c3ce light
bosh-aws-xen-hvm-ubuntu-trusty-go_agent	3262.7*     	ubuntu-trusty	ami-3b77ee2c light
bosh-aws-xen-hvm-ubuntu-trusty-go_agent	3232        	ubuntu-trusty	ami-c63b21ac
bosh-aws-xen-hvm-ubuntu-trusty-go_agent	3231        	ubuntu-trusty	ami-ec140e86
bosh-aws-xen-ubuntu-trusty-go_agent    	3290*       	ubuntu-trusty	ami-4e5a2c59 light
bosh-aws-xen-ubuntu-trusty-go_agent    	3262.12     	ubuntu-trusty	ami-96dca881 light
bosh-aws-xen-ubuntu-trusty-go_agent    	3262.9*     	ubuntu-trusty	ami-5e6c0349 light
```

Here is the pipeline that produces CLI binaries if you want to try it out and let us know if it fits in your workflows or if you have suggestions on the formatting of the new CLI: https://main.bosh-ci.cf-app.com/pipelines/bosh:cli. (Few differences in the new CLI: https://github.com/cloudfoundry/docs-bosh/blob/master/cli-v2.html.md.erb).

Issue #939, opened 2015-08-24
Keystone v3 support not working for domains other than 'Default'


voelzmo:
**Problem**
Deploying a bosh director with bosh-init to a keystone v3 enabled openstack environment with a domain name different than 'Default' fails during stemcell upload. The reason is that the used version of fog doesn't implement passing the necessary domain name argument to openstack in the `Fog::Compute` component - so any calls involving nova will fail.

**Analysis**
There is a bug in fog 1.31.0 which leads to [Fog::Compute not knowing about the `openstack_domain_name` parameter](https://github.com/fog/fog/blob/v1.31.0/lib/fog/openstack/compute.rb#L7-L14)
Therefore, the output in the console is `STDERR: '[fog][WARNING] Unrecognized arguments: openstack_domain_name` followed by an HTTP 404
```
D, [2015-08-21T09:15:52.852051 #3870] DEBUG -- : excon.error {:error=>#<Excon::Errors::NotFound: Expected([201]) <=> Actual(404 Not Found)
excon.error.response
  :body          => "<html>\r\n<head><title>404 Not Found</title></head>\r\n<body bgcolor=\"white\">\r\n<center><h1>404 Not Found</h1></center>\r\n<hr>
<center>nginx/1.4.4</center>\r\n</body>\r\n</html>\r\n"
  :headers       => {
    "Connection"     => "keep-alive"
    "Content-Length" => "168"
    "Content-Type"   => "text/html"
    "Date"           => "Fri, 21 Aug 2015 09:15:53 GMT"
    "Server"         => "nginx/1.4.4"
  }
  :local_address => "192.168.0.104"
  :local_port    => 41757
  :reason_phrase => "Not Found"
  :remote_ip     => "147.204.8.32"
  :status        => 404
  :status_line   => "HTTP/1.1 404 Not Found\r\n"
>}
E, [2015-08-21T09:15:52.852453 #3870] ERROR -- : Unable to connect to the OpenStack Compute API. Check task debug log for details.
[Cmd Runner] 2015/08/21 09:15:52 DEBUG - Successful: true (0)
[cpiCmdRunner] 2015/08/21 09:15:52 DEBUG - Exit Code 0 when executing external CPI command '/home/vcap/.bosh_init/installations/5ead9388-6a94-45c6-50b1
-75c84a703849/jobs/cpi/bin/cpi'
```

[...]

PR #950, opened 2015-09-15
Fix Openstack Keystone V3 API

voelzmo:
Bug: https://github.com/cloudfoundry/bosh/issues/939

Modify registry, director, and swift blobstore client:
* introduce new parameter 'project', which is mandatory for Keystone V3
* mark parameter 'tenant' as optional: only necessary for Keystone V2


voelzmo:
Pulled in with https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release/commit/8dc83d427e61ff1f53f22f5e604439dfb5aa0cad

PR #952, opened 2015-09-17
Bosh delete deployment should succeed if rendered templates archives …

guoger:
…are missing

[#101724116]

### DIFF
--- a/bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
+++ b/bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
@@ -1,20 +1,31 @@
 module Bosh::Director
   class RenderedJobTemplatesCleaner
-    def initialize(instance_model, blobstore)
+    def initialize(instance_model, blobstore, logger)
       @instance_model = instance_model
       @blobstore = blobstore
+      @logger = logger
     end
 
     def clean
       @instance_model.stale_rendered_templates_archives.each do |archive|
-        @blobstore.delete(archive.blobstore_id)
+        begin
+          @blobstore.delete(archive.blobstore_id)
+        rescue Bosh::Blobstore::NotFound => e
+          @logger.debug("Blobstore#delete error: #{e.message}, will ignore this error and delete the db record")
+        end
+
         archive.delete
       end
     end
 
     def clean_all
       @instance_model.rendered_templates_archives.each do |archive|
-        @blobstore.delete(archive.blobstore_id)
+        begin
+          @blobstore.delete(archive.blobstore_id)
+        rescue Bosh::Blobstore::NotFound => e
+          @logger.debug("Blobstore#delete error: #{e.message}, will ignore this error and delete the db record")
+        end
+
         archive.delete
       end
     end
--- a/blobstore_client/lib/blobstore_client/dav_blobstore_client.rb
+++ b/blobstore_client/lib/blobstore_client/dav_blobstore_client.rb
@@ -54,6 +54,7 @@ module Bosh
       def delete_object(id)
         response = @client.delete(url(id), header: @headers)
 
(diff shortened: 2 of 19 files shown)


bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
```diff
@@ -1,20 +1,33 @@
...
+    def initialize(instance_model, blobstore, logger)
       @instance_model = instance_model
       @blobstore = blobstore
+      @logger = logger
     end
 
     def clean
       @instance_model.stale_rendered_templates_archives.each do |archive|
-        @blobstore.delete(archive.blobstore_id)
+        if @blobstore.exists? archive.blobstore_id
```
cppforlife: Instead of making an extra guaranteed call to the blobstore (`exists?`), let's just catch not found error from `delete` and log then.

bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
```diff
@@ -1,20 +1,33 @@
...
+        end
+
         archive.delete
       end
     end
 
     def clean_all
       @instance_model.rendered_templates_archives.each do |archive|
-        @blobstore.delete(archive.blobstore_id)
+        if @blobstore.exists? archive.blobstore_id
```
cppforlife: same here.

bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
xingzhou: @dk, the blobstore client like dav-blobstore-client will not throw NotFound exception, instead, it will generally throw out a BlobstoreError exception and saying that "Could not delete object", from this generic error, we can not tell if this error is because of object not found or object deletion errors

bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
xingzhou: @dk, please see my comments above,thx

bosh-director/lib/bosh/director/rendered_job_templates_cleaner.rb
cppforlife: @xingzhou can we check on response code (404) inside the delete_object and raise NotFound?

xingzhou: @dk, yes, but one thing is we need to change all kinds of blobstore clients(dav, local, s3) to raise NotFound error instead of BlobstoreError, this may introduce more changes, if you think is ok, we will change the Blobstore client then

cppforlife: Let's do that:
- local seems to already handle that
- s3 is almost there except it does not raise NotFound (how come delete_object does not use AWS::S3::Errors::NoSuchKey error just like get_object?)

xingzhou: sure, will resubmit this then

cppforlife:
Can you also update `blobstore_client/spec/functional/dav_spec.rb` with tests about deletion of not found objects?

mariash:
@guoger could you look at unit test failures on travis, they seem to be related. 

Could you please also add a functional tests in blobstore_client/spec/functional/?

guoger:
@mariash Sure, we'll look into that!

xingzhou:
@mariash and @cppforlife, have resubmitted the PR, for functional test, we found the s3_spec is out of date, we made some additional changes to this spec to make the test cases pass, FYI.

xingzhou:
To @mariash , there is an rubocop violation:

are you going to fix it?

mariash:
@xingzhou should be fixed now, sorry about that :)

xingzhou:
Thanks @mariash , we have rebased and resubmitted the PR

PR #963, opened 2015-09-25
Add Azure support.

AbelHu:


AbelHu:
We have signed the Corporate CLA. Could you help to merge it?

cppforlife:
Planning to merge it in this story: https://www.pivotaltracker.com/story/show/104454938

allomov:
@cppforlife could you comment on your plans to support solution with registry?

Do you plan to make it IaaS independent or remove from bosh at all?

uzzz:
Looks like it won't work. InstanceManager class has template method instance_ips – https://github.com/cloudfoundry/bosh/blob/master/bosh-registry/lib/bosh/registry/instance_manager.rb#L45 which is defined only in subclasses.

AbelHu:
Please reference https://github.com/cloudfoundry/bosh/blob/master/bosh-registry/lib/bosh/registry/api_controller.rb#L21-L22

AbelHu:
Please reference https://github.com/cloudfoundry/bosh/blob/master/bosh-registry/lib/bosh/registry/api_controller.rb#L21-L22

uzzz:
I see. So maybe it will be useful to add default method implementation to `InstanceManager` class, something like that:

```ruby
def instance_ips(instance_id)
  raise NotImplemenedError, "Default implementation of InstanceManager does not support " \ 
    "IPs retrieval. Create IaaS-specific subclass and override this method"
end
```

It would be nicer to get such explanation in case if someone will call `read_settings` with second argument specified instead of just NoMethodError.

AbelHu:
From my understanding, only when the registry client connects to the server without username/password, the registry server will check the client's IP address.

AbelHu:
I think your suggestion is good. I will add the default instance_ips in next submit.

stemcell_builder/stages/bosh_azure_agent_settings/apply.sh
```diff
@@ -0,0 +1,31 @@
...
+# Set SettingsPath but never use it because file_meta_service is avaliable only when the settings file exists.
+cat > $chroot/var/vcap/bosh/agent.json <<JSON
+{
+  "Platform": {
+    "Linux": {
+      "DevicePathResolutionType": "virtio"
+    }
+  },
+  "Infrastructure": {
+    "NetworkingType": "dhcp",
```
cppforlife: i dont think agent cares about `NetworkingType` anymore.

stemcell_builder/stages/image_install_grub/apply.sh
```diff
@@ -68,18 +68,25 @@ then
...
   echo "(hd0) ${device}" > ${image_mount_point}/device.map
 
   # install bootsector into disk image file
   run_in_chroot ${image_mount_point} "grub2-install -v --no-floppy --grub-mkdevicemap=/device.map ${device}"
 
-  cat >${image_mount_point}/etc/default/grub <<EOF
+  # Append "console=ttyS0,115200n8 earlyprintk=ttyS0 rootdelay=300" for azure
+  if [ "${stemcell_infrastructure}" == "azure" ]; then
+    cat >${image_mount_point}/etc/default/grub <<EOF
+GRUB_CMDLINE_LINUX="vconsole.keymap=us net.ifnames=0 crashkernel=auto selinux=0 plymouth.enable=0 console=ttyS0,115200n8 earlyprintk=ttyS0 rootdelay=300"
```
cppforlife: why do we need rootdelay? what about earlyprintk?

i believe we set console=ttyS0,115200n8 configuration for openstack stemcells. i think we can always configure it for all stemcells.

stemcell_builder/stages/bosh_azure_agent_settings/apply.sh
AbelHu: It is because that Windows Azure Linux agent needs to access network before generating CustomData. bosh-agent will read CustomData before configuring network. So we have to set this.

cppforlife: If you grep bosh-agent source code, there is no NetworkingType. May be you want to set use_dhcp flag on each network: https://github.com/cloudfoundry/bosh-agent/blob/master/settings/settings.go#L124 like AWS CPI does: https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/blob/ec62d78c16c857f1a6538a9365b4c8f3204d0a57/src/bosh_aws_cpi/lib/cloud/aws/cloud.rb#L784

AbelHu: We have set use_dhcp in Azure CPI https://github.com/Azure/bosh-azure-cpi-release/blob/master/src/bosh_azure_cpi/lib/cloud/azure/cloud.rb#L416. I will remove this from agent.json as AWS did https://github.com/cloudfoundry/bosh/commit/2f95fd89b8bbb1edad8e9e52d6f80d5996acd629.

stemcell_builder/stages/system_azure_wala/apply.sh
```diff
@@ -0,0 +1,35 @@
...
+
+base_dir=$(readlink -nf $(dirname $0)/../..)
+source $base_dir/lib/prelude_apply.bash
+
+packages="python python-pyasn1"
+pkg_mgr install $packages
+
+wala_release=2.0.14
+run_in_chroot $chroot "
+  curl -L https://raw.githubusercontent.com/Azure/WALinuxAgent/WALinuxAgent-${wala_release}/waagent > /tmp/waagent
```
cppforlife: need to add sha sum verification

stemcell_builder/stages/system_azure_wala/apply.sh
```diff
@@ -0,0 +1,35 @@
...
+  chmod 0755 /tmp/waagent
+  cp -a /tmp/waagent /usr/sbin/waagent
```
cppforlife: where do we delete tmp file for the agent?

AbelHu: Done

cppforlife:
Rebased as new PR: https://github.com/cloudfoundry/bosh/pull/998. Closing this one.

cppforlife:
@AbelHu registry changes were merged. i believe we can now use unmodified bosh release when deploying the director. you should be able to update CI scripts to use official bosh release.

AbelHu:
@cppforlife Thanks for the reminder. We will do it.

PR #973, opened 2015-10-11
Clarify `bosh scp` and `bosh ssh` documentation

### LINKED ISSUE #884: Unclear `bosh scp` command documentation
I want to download file from bosh-managed VM. `bosh --help` is not clear about `scp` command usage.

```
# Wrong way; my first guess after reading docs
$ bosh scp --download /tmp/cf-logs.logs ingester/0

# Right way; just guessed
$ bosh scp ingester/0 --download /tmp/cf-logs.logs /tmp
```

Help:
```
scp [--download] [--upload] [--public_key FILE] [--gateway_host HOST]
[--gateway_user USER] [--gateway_identity_file FILE]
    upload/download the source file to the given job. Note: for download /path/to/destination is a
    directory
..
```

1. Tell that `job/index` must go right after scp, than options
2. Tell that  `/path/to/destination` is required for download and it is placed at the end of string without any options.
3. Maybe some symmetrical syntax like --from/--to ?

jhunt:
- Add positional arguments to the usage string
- Be more explicit re: what `scp --upload` and `--download` options do

Fixes #884

### DIFF
--- a/bosh_cli/lib/cli/commands/ssh.rb
+++ b/bosh_cli/lib/cli/commands/ssh.rb
@@ -6,7 +6,7 @@ module Bosh::Cli
     class Ssh < Base
 
       # bosh ssh
-      usage 'ssh'
+      usage 'ssh <job>/<index>'
       desc 'Execute command or start an interactive session'
       option '--gateway_host HOST', 'Gateway host'
       option '--gateway_user USER', 'Gateway user'
@@ -38,11 +38,11 @@ module Bosh::Cli
       end
 
       # bosh scp
-      usage 'scp'
-      desc 'upload/download the source file to the given job. ' +
-             'Note: for download /path/to/destination is a directory'
-      option '--download', 'Download file'
-      option '--upload', 'Upload file'
+      usage 'scp <job>/<index> <source> <destination>'
+      desc "Transfer files to (--upload) or from (--download) a job.\n" +
+             'Note: for --download, <destination> is treated as a directory'
+      option '--download', 'Download <source> file from the job'
+      option '--upload', 'Upload <source> file to the job'
       option '--gateway_host HOST', 'Gateway host'
       option '--gateway_user USER', 'Gateway user'
       option '--gateway_identity_file FILE', 'Gateway identity file'


bosh_cli/lib/cli/commands/ssh.rb
```diff
@@ -6,7 +6,7 @@ module Command
     class Ssh < Base
 
       # bosh ssh
-      usage 'ssh'
+      usage 'ssh <job>/<index>'
```
cppforlife: Does this break command look up?

filefrog: I'm not 100% sure, but with the changes in place, things like `bosh help ssh` seems to work...

Is there a way I can test that?

cppforlife: Does `bosh ssh` actually work after this change?

filefrog: Yes.  `bosh ssh jobname/0`, `bosh ssh jobname/0 ls /tmp` and `bosh ssh jobname 0` all work

filefrog: Also, various invocations of `bosh scp` also still work.

voelzmo: My guess at what you are trying to achieve here is to have `bosh help scp` print out the parameters that can be passed into that command. `usage` is probably not meant to be used this way. Other commands, such as `bosh deployment` or `bosh login` print their parameters, the printing even considers if the parameters are optional or not. To achieve this, they use some funky logic hidden in https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/commands/help.rb#L62-L74

The parsing of method parameters happens in https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/command_handler.rb#L66-L74.

So the problem seems to be that `bosh ssh` and `bosh scp` accept `*args` instead of specifying what might go in there – if these methods were taking specific arguments, you'd get the documentation for free.

Note that there is some funky comparisons going on with `cmd.usage` in https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/commands/help.rb#L40 and https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/commands/help.rb#L51 so your change might add some unexpected sideeffects.

The configuration adds commands with `usage` as key: https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/config.rb#L39 and this is used in parsing the commandline to find out which command to run: https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/runner.rb#L214-L218

So I'd rather not have the `usage` contain parameters and rather have some real arguments in the method itself.

filefrog: Currently, `bosh scp` allows the following calling conventions:

    bosh scp box_z1/3 /a /b --upload
    bosh scp box_z1 3 /a /b --upload     # space-separated job and index number

Changing the call signature of `scp(*args)` to either `scp(job, index, source, destination)` or `scp(job_and_index, source, destination)` would seem to force one and drop support for the other.

I suppose we could do something like this:

```
def scp(job, index, source, destination)
  if destination.nil? && job.match('/')
    destination = source
    source = index
    job, index = job.split('/', 2)
  end
  # ...
end
```

I find that a bit more awkward than the current `*args` approach, but if that is preferred to (ab)using the usage string, I don't mind.  I really just want the help screen to reflect accurate usage of `bosh scp`, since I was also initially confused about how to invoke it.

filefrog: I'm not actually even sure that the above code would work for the /-separated job/index form, since that's only three arguments and the function *requires* four.  And making the `destination` flag optional leads to bad documentation (since it's not really optional)

jhunt:
Any thoughts on this?

geofffranks:
@voelzmo @cppforlife What needs to happen to get this in a merge-able state?

dpb587-pivotal:
fwiw... the incomplete scp command help also confuses the bosh devs :)

jhunt:
@dpb587-pivotal - can we get this PR merged then?

dpb587-pivotal:
Successfully tested local merge of 78bf608...

 * scp - tested upload and download; tested older `<job> <index> ...` syntax
 * ssh - tested interactive and command argument
 * units are [green](https://main.bosh-ci.cf-app.com/builds/108621)
 * integrations are [green](https://main.bosh-ci.cf-app.com/builds/108504)

```
$ bosh help scp
scp <job>/<index> <source> <destination> [--download] [--upload]
[--gateway_host HOST] [--gateway_user USER] [--gateway_identity_file
FILE] [--no_gateway]
    Transfer files to (--upload) or from (--download) a job.
    Note: for --download, <destination> is treated as a directory
    --download                                                Download <source> file from the job
    --upload                                                  Upload <source> file to the job
    --gateway_host HOST                                       Gateway host
    --gateway_user USER                                       Gateway user
    --gateway_identity_file FILE                              Gateway identity file
    --no_gateway                                              Ignore gateway provided by the director

$ bosh help ssh
ssh <job>/<index> [--gateway_host HOST] [--gateway_user USER]
[--gateway_identity_file FILE] [--default_password PASSWORD]
[--strict_host_key_checking <yes/no>] [--no_gateway]
    Execute command or start an interactive session
    --gateway_host HOST                                                     Gateway host
    --gateway_user USER                                                     Gateway user
    --gateway_identity_file FILE                                            Gateway identity file
    --default_password PASSWORD                                             Use default ssh password (NOT RECOMMENDED)
    --strict_host_key_checking <yes/no>                                     Can use this flag to skip host key checking (NOT RECOMMENDED)
    --no_gateway                                                            Ignore gateway provided by the director
```

dpb587-pivotal:
Merged - thanks!

Issue #769, opened 2015-03-04
[bosh_cli] RFE: Set "--no-color" by default by configuration


cyrille-leclerc:
It would be very nice to be able to specify "`--no-color`" by configuration, if possible with an environment variable, as we can already define `BOSH_USER` and `BOSH_PASSWORD `.

drnic:
If you're interested in adding support, I think
https://github.com/cloudfoundry/bosh/blob/master/bosh_cli/lib/cli/config.rb
is a good place to start (+ an update to the specs + README)

cyrille-leclerc:
Thanks @drnic, I'll try but I am completely new to Ruby and the learning curve has been stiff fore te moment :-)

Enhancing error message of the CLI may be an easier entry point for me.

Anyway, I'll try, thanks!

cppforlife:
makes sense to make an ENV for it. e.g. BOSH_COLOR=false to make it consistent with cf cli?

cyrille-leclerc:
@cppforlife this looks perfect.

filefrog:
I put in pull request #975 for this.

cppforlife:
In CLI 1.3122.0

PR #975, opened 2015-10-11
Support BOSH_COLOR env var for implicit --no-color

filefrog:
If BOSH_COLOR is set to the string "false", then the `bosh' utility will
behave as if the `--no-color` option has been set.  This behavior can be
overridden by also passing `--color` (which takes precedence)

If BOSH_COLOR is not set, or has any other value, the previous default
behavior holds (only do color if we are hooked up to a terminal)

### DIFF
--- a/bosh_cli/lib/cli/runner.rb
+++ b/bosh_cli/lib/cli/runner.rb
@@ -25,8 +25,8 @@ module Bosh::Cli
       @option_parser = OptionParser.new(banner)
 
       Config.colorize = nil
-      if ENV.has_key? "BOSH_COLOR"
-        Config.colorize = ENV["BOSH_COLOR"] == "false"
+      if ENV.has_key?("BOSH_COLOR") && ENV["BOSH_COLOR"] == "false"
+        Config.colorize = false
       end
       Config.output ||= STDOUT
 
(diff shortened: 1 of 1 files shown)


filefrog:
There are no tests for this, since I could find no pre-existing tests for `--no-color` / `--color` options to the bosh utility.

If anyone has any suggestions on the best way to architect such tests, I'm willing to implement them.

filefrog:
Also, this implements #769

bosh_cli/lib/cli/runner.rb
```diff
@@ -25,6 +25,9 @@ def initialize(args, options = {})
       @option_parser = OptionParser.new(banner)
 
       Config.colorize = nil
+      if ENV.has_key? "BOSH_COLOR"
+        Config.colorize = ENV["BOSH_COLOR"] == "false"
```
cppforlife: should colorize be set to true when BOSH_COLOR==false?

filefrog: Hah, yeah that's wrong.  Tests sure would be helpful here :grin:

cppforlife:
In CLI 1.3120.0

PR #978, opened 2015-10-13
bosh instances should filter non-failing instances if --failing flag …

zhang-hua:
…is given

[#100344958](https://www.pivotaltracker.com/story/show/100344958)


cppforlife:
Story was rejected: https://www.pivotaltracker.com/story/show/100344958.

Issue #981, opened 2015-10-13
Bosh create release fails when there is a missing directory/file in the jobs dir 


fraenkel:
I was hitting this lovely error
```.../.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/resources/job.rb:12:in `block in discover': undefined method `<<' for nil:NilClass (NoMethodError)```

The bug is in line 8
        next unless File.directory?(job_base)

It needs to return jobs otherwise you end up with a nil array.

**Comment by allomov** [comment]
did you change sources of `lib/cli/resources/job.rb`? I am confused because error message show that error happen on line 12 and there is no any [code invocation there](https://github.com/cloudfoundry/bosh/blob/stable-3093/bosh_cli/lib/cli/resources/job.rb#L12).

**Comment by allomov** [comment]
@fraenkel really "lovely" behaviour.

I fixed it in this PR https://github.com/cloudfoundry/bosh/pull/983

**Comment by fraenkel** [comment]
Nope, here is a more complete stack
```
/Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/resources/job.rb:12:in `block in discover': undefined method `<<' for nil:NilClass (NoMethodError)
	from /Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/resources/job.rb:7:in `each'
	from /Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/resources/job.rb:7:in `inject'
	from /Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/resources/job.rb:7:in `discover'
	from /Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/commands/release/create_release.rb:194:in `build_jobs'
	from /Users/fraenkel/.gem/ruby/2.1.7/gems/bosh_cli-1.3093.0/lib/cli/commands/release/create_release.rb:98:in `create_from_spec'
```

cppforlife:
In CLI 1.3120.0

PR #983, opened 2015-10-14
Fix problem in CLI when jobs folder contain files

allomov:
This PR fixes problem described in this issue https://github.com/cloudfoundry/bosh/issues/981


cppforlife:
In CLI 1.3120.0

PR #992, opened 2015-10-22
Fix the issue that /etc/sudoers.d is not included correctly.

AbelHu:
Correct one issue which is introduced in this commit: https://github.com/cloudfoundry/bosh/commit/650ef135c85bb8b0e1a51eb5d5a413980de217f9

No blank line as the last line in /etc/sudoers so that the last line will be as below.
%bosh_sudoers ALL=(ALL) NOPASSWD: ALL#includedir /etc/sudoers.d

### DIFF
--- a/stemcell_builder/stages/bosh_sudoers/apply.sh
+++ b/stemcell_builder/stages/bosh_sudoers/apply.sh
@@ -10,6 +10,7 @@ source $base_dir/lib/prelude_bosh.bash
 
 # setup sudoers to use includedir, and make sure we don't break anything
 cp -p $chroot/etc/sudoers $chroot/etc/sudoers.save
+echo '' >> $chroot/etc/sudoers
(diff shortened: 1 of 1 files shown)


pivotal-nader-ziada:
Why is the empty line needed in the sudoers file? is it causing a problem?

AbelHu:
If no empty line, the last line in the final /etc/sudoers will be "sudoers ALL=(ALL) NOPASSWD: ALL#includedir /etc/sudoers.d"

It will prompts you need to input the password for the user vcap when using 'sudo su'.

Best Regards,
Abel HU

? 2015?10?23??00:21?Nader Ziada <notifications@github.com<mailto:notifications@github.com>> ???


Why is the empty line needed in the sudoers file? is it causing a problem?

-

cppforlife:
Fix was made: https://github.com/cloudfoundry/bosh/commit/9f2c37ae207325e6ab499485701a235800dfd0d7.

PR #994, opened 2015-10-23
Convert director.yml.erb to use top level JSON.dump

hashmap:
JSON.dump doesn't work for strings in ruby 1.9.3, also escaping is
required for each property. This fix constructs ruby hash and then
convert it to JSON/YAML by a single JSON.dump call.

Also additional test for vsphere was introduced to catch conversion
errors. Verified on AWS, verification on the rest of CPIs is required.

[#104923910] (https://www.pivotaltracker.com/story/show/104923910)

### DIFF
--- a/release/jobs/director/templates/director.yml.erb.erb
+++ b/release/jobs/director/templates/director.yml.erb.erb
@@ -1,438 +1,534 @@
----
-name: <%= p('director.name') %>
-port: <%= p('director.backend_port') %>
-encryption: <%= p('director.encryption') %>
+<%=
+params = {
+  'name' => p('director.name'),
+  'port' => p('director.backend_port'),
+  'encryption' => p('director.encryption'),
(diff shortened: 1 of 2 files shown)


allomov:
Thank you @hashmap. Looks much more better for me. 

It seems it will fix this issue https://github.com/cloudfoundry/bosh-init/pull/46 with bosh-init.

hashmap:
@allomov yes, I was able to deploy bosh using bosh-init with ubuntu 14.04 default ruby (1.9.3)

allomov:
I think it would be better to add 1.9.3 ruby version to [this list](https://github.com/cloudfoundry/bosh/blob/master/.travis.yml#L13) in `.travis.yml` to be sure nothing is broken by upcoming changes.

cppforlife:
Merged.

PR #998, opened 2015-10-29
Rebased azure support for the stemcell builder

cppforlife:


cppforlife:
I believe this was merged.

PR #1004, opened 2015-11-03
verify blobstore contents and replace with correct versions with --fi…

xingzhou:
…x option

[#101003512](https://www.pivotaltracker.com/n/projects/956238/stories/101003512)

Signed-off-by: Xing Zhou <xingzhou@cn.ibm.com>

### DIFF
--- a/bosh-director/lib/bosh/director/blob_util.rb
+++ b/bosh-director/lib/bosh/director/blob_util.rb
@@ -22,6 +22,15 @@ module Bosh::Director
       blobstore_id
     end
 
+    def self.replace_blob(old_blobstore_id, path)
+      blobstore.delete(old_blobstore_id) if blobstore.exists?(old_blobstore_id)
+      return create_blob path
+    end
+
+    def self.verify_blob(blobstore_id, sha1)
+      blobstore.exists?(blobstore_id) && sha1 == Digest::SHA1.hexdigest(blobstore.get(blobstore_id))
+    end
+
     def self.save_to_global_cache(compiled_package, cache_key)
       global_cache_filename = [compiled_package.package.name, cache_key].join('-')
       Dir.mktmpdir do |path|
--- a/bosh-director/lib/bosh/director/jobs/update_release.rb
+++ b/bosh-director/lib/bosh/director/jobs/update_release.rb
@@ -33,6 +33,7 @@ module Bosh::Director
         @release_model, @release_version_model, @manifest, @name, @version = nil, nil, nil, nil, nil
 
         @rebase = !!options['rebase']
+        @fix = !!options['fix']
       end
 
       # Extracts release tarball, verifies release manifest and saves release in DB
@@ -328,7 +329,7 @@ module Bosh::Director
               end
             end
 
-            if source_release && package.blobstore_id.nil?
+            if source_release && (package.blobstore_id.nil? || @fix)
               save_package_source_blob(package, package_meta, release_dir)
               package.save
             end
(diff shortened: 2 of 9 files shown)


bosh-director/lib/bosh/director/jobs/update_release.rb
```diff
@@ -484,25 +485,43 @@ def create_package(package_meta, release_dir)
...
-
-        name, version, existing_blob = package_meta['name'], package_meta['version'], package_meta['blobstore_id']
+        name, version, existing_blob, sha1 = package_meta['name'], package_meta['version'], package_meta['blobstore_id'], package_meta['sha1']
         desc = "package '#{name}/#{version}'"
+        package_tgz = File.join(release_dir, 'packages', "#{name}.tgz")
+
+        if @fix
+          if package.blobstore_id != nil
+            logger.info("Verifying package #{desc} with blobstore_id: #{package.blobstore_id}")
+            return false if BlobUtil.verify_blob(package.blobstore_id, package.sha1)
```
mariash: Seems like we are downloading blob for sha verification anyway, so what is the point of checking this (instead of just uploading every package)?

@loewenstein & @mariash

bosh_cli/lib/cli/commands/release/upload_release.rb
```diff
@@ -125,7 +133,7 @@ def upload_tarball(tarball_path, upload_options = {})
 
             # if the director is an older version that doesn't support optimized unpack we
             # have to upload everything
-            if tarball.upload_packages?(package_matches) || !director_supports_fast_unpack?
+            if tarball.upload_packages?(package_matches) || !director_supports_fast_unpack? || upload_options[:fix]
```
mariash: unneeded or

bosh_cli/spec/unit/commands/release/upload_release_spec.rb
```diff
@@ -253,6 +253,32 @@ module Bosh::Cli::Command::Release
...
+
+            it 'should upload the release archive along with fix option' do
+              expect(command).to receive(:upload_tarball)
+                .with(release_archive, hash_including(:fix => true))
+                .and_call_original
+              expect(director).to receive(:upload_release).with(release_archive, hash_including(:fix => true))
+              command.upload(release_archive)
+            end
+
+            it 'should ignore --name and --version' do
```
mariash: does not validate if release already uploaded

bosh-director/lib/bosh/director/blob_util.rb
```diff
@@ -22,6 +22,15 @@ def self.copy_blob(blobstore_id)
       blobstore_id
     end
 
+    def self.replace_blob(old_blobstore_id, path)
+      blobstore.delete(old_blobstore_id) if blobstore.exists?(old_blobstore_id)
```
cppforlife: let's just call delete and ignore blob not found call.

bosh-director/lib/bosh/director/blob_util.rb
```diff
@@ -22,6 +22,15 @@ def self.copy_blob(blobstore_id)
...
+      return create_blob path
+    end
+
+    def self.verify_blob(blobstore_id, sha1)
+      blobstore.exists?(blobstore_id) && sha1 == Digest::SHA1.hexdigest(blobstore.get(blobstore_id))
```
cppforlife: what cleans up file created by blobstore.get(blobstore_id)?

bosh-director/lib/bosh/director/jobs/update_release.rb
xingzhou: @mariash, if we just upload the package, might introduce other issue, e.g. for the healthy package in the blobstore, if any exception happens during the uploading of the package, this may cause we use a damaged package to replace the original healthy one. and also, if we just upload the package every time, we need to update the db record and clean the old package file in the blobstore, which is also a more 'longer' process to take.

bosh-director/lib/bosh/director/blob_util.rb
xingzhou: we will fix this and submit a PR later

bosh_cli/lib/cli/commands/release/upload_release.rb
xingzhou: we will fix this and submit a new PR later

bosh-director/lib/bosh/director/blob_util.rb
xingzhou: here we are using the #get method of the blobstore client, and if there is no file name provided, the content will be written into a temp file and will be cleaned up by blobstore client after we get the content of the file, so here we don't manually cleanup this file

bosh_cli/spec/unit/commands/release/upload_release_spec.rb
xingzhou: @mariash, based on our understanding, if we provide --fix option together with --name & --version, we will ignore the check of whether the release is already uploaded, so we don't check this in the test

PR #1006, opened 2015-11-03
bosh cli should present the full details of the vm state

zhang-hua:
[#100344890](https://www.pivotaltracker.com/story/show/100344890)

### DIFF
--- a/bosh_cli/lib/cli/commands/instances.rb
+++ b/bosh_cli/lib/cli/commands/instances.rb
@@ -48,6 +48,8 @@ module Bosh::Cli::Command
 
       row_count = 0
       has_disk_cid = instances[0].has_key?('disk_cid')
+      has_uptime = instances[0]['processes'] && instances[0]['processes'].size > 0 && instances[0]['processes'][0].has_key?('uptime')
+      has_cpu = instances[0]['processes'] && instances[0]['processes'].size > 0 && instances[0]['processes'][0].has_key?('cpu')
 
       instances_table = table do |t|
         headings = ['Instance', 'State', 'Resource Pool', 'IPs']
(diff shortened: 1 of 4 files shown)


cppforlife:
I've updated the story with some changes to how CLI should behave.

spec/integration/cli_instances_spec.rb
```diff
@@ -50,12 +48,10 @@
         |   process-1 | running |               |             |
         |   process-2 | running |               |             |
         |   process-3 | failing |               |             |
-        +-------------+---------+---------------+-------------+
```
cppforlife: why was the separator removed? we still want to the separator between instances if --ps is specified.

bosh_cli/lib/cli/commands/instances.rb
```diff
@@ -48,6 +48,8 @@ def show_deployment(name, options={})
 
       row_count = 0
       has_disk_cid = instances[0].has_key?('disk_cid')
+      has_uptime = instances[0]['processes'] && instances[0]['processes'][0].has_key?('uptime')
```
cppforlife: what if there are not processes?

bosh_cli/lib/cli/commands/instances.rb
```diff
@@ -48,6 +48,8 @@ def show_deployment(name, options={})
...
+      has_cpu = instances[0]['processes'] && instances[0]['processes'][0].has_key?('cpu')
```
cppforlife: what if there are not processes?

spec/integration/cli_instances_spec.rb
zhang-hua: @cppforlife , ok, misunderstand it just for --ps. will update it.

bosh_cli/lib/cli/commands/instances.rb
zhang-hua: @cppforlife , good point, will fix it.

bosh_cli/lib/cli/commands/instances.rb
zhang-hua: @cppforlife , will fix it.

PR #1007, opened 2015-11-04
Refine replace_blob method to raise error while deleting missing blob,

xingzhou:
rather than doing extra existence check.

[#101003512](https://www.pivotaltracker.com/story/show/101003512)

Signed-off-by: Xing Zhou <xingzhou@cn.ibm.com>

### DIFF
--- a/bosh-director/lib/bosh/director/blob_util.rb
+++ b/bosh-director/lib/bosh/director/blob_util.rb
@@ -22,9 +22,8 @@ module Bosh::Director
       blobstore_id
     end
 
-    def self.replace_blob(old_blobstore_id, path)
-      blobstore.delete(old_blobstore_id) if blobstore.exists?(old_blobstore_id)
-      return create_blob path
+    def self.delete_blob(blobstore_id)
+      blobstore.delete(blobstore_id)
     end
 
     def self.save_to_global_cache(compiled_package, cache_key)
--- a/bosh-director/lib/bosh/director/jobs/update_release.rb
+++ b/bosh-director/lib/bosh/director/jobs/update_release.rb
@@ -490,22 +490,20 @@ module Bosh::Director
         package_tgz = File.join(release_dir, 'packages', "#{name}.tgz")
 
         if @fix
+          package.sha1 = package_meta['sha1']
+
           if package.blobstore_id != nil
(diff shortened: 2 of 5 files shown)


bosh-director/lib/bosh/director/blob_util.rb
```diff
@@ -23,7 +23,11 @@ def self.copy_blob(blobstore_id)
     end
 
     def self.replace_blob(old_blobstore_id, path)
-      blobstore.delete(old_blobstore_id) if blobstore.exists?(old_blobstore_id)
+      begin
+        blobstore.delete(old_blobstore_id)
+      rescue Bosh::Blobstore::BlobstoreError => e
```
cppforlife: oops forgot to mention. let's make sure to log to debug about failure ignore. i believe we do similar thing in other places.

bosh-director/lib/bosh/director/jobs/update_release.rb
```diff
@@ -638,10 +635,15 @@ def next_release_version
...
 
       def fix_package(package, package_tgz)
-        single_step_stage("Fixing package '#{package.name}/#{package.version}'") do
-          logger.info("Fixing package '#{package.name}/#{package.version}'")
-          package.blobstore_id = BlobUtil.replace_blob(package.blobstore_id, package_tgz)
+        begin
+          logger.info("Deleting package '#{package.name}/#{package.version}'")
+          BlobUtil.delete_blob(package.blobstore_id)
+        rescue Bosh::Blobstore::BlobstoreError => e
+          logger.info("Error deleting '#{package.name}/#{package.version}' #{e.message}")
```
cppforlife: Please include blobstore_id and e.inspect in the error log.

guoger:
@cppforlife , just to comment on your rejection in the tracker, package.sha1 should not be affected since we are replacing corrupted blobs with correct ones, which should already match the sha1 in database, since they are identical packages.

Thanks!
/Jay

cppforlife:
@guoger they may be identical in content (fingerprint is the same) but because tar generates different file for the same content sha1 may not match.

guoger:
@cppforlife That's true, unlike simple tar, gzip (-z) indeed produces different bits out of same content. Didn't know that and thanks for pointing it out.

PR #1012, opened 2015-11-06
Initial support for db migration unit tests

loewenstein:
With this a test has fine grained control over Sequel migrations.

Just do

```ruby
during_migration(<namespace> (1)) do | migration, db |
    stop_before <migration name>

   # prepare the database

   stop_after <migration name>

   # assert migration result (2)
end
```

You can see a running example at https://github.com/loewenstein/bosh/blob/db-migration-unit-tests/bosh-director/spec/unit/db_migrations/20151031001039_add_spec_to_instance_spec.rb

Assertions can be made either using Sequel::Models. You have to ```reset_models``` the model classes though to make them compliant to the current version (the current migration step) of the database schema.
Alternatively you can use raw Sequel expressions like the following (details are in the history of https://github.com/loewenstein/bosh/blob/db-migration-unit-tests/bosh-director/spec/unit/db_migrations/20151031001039_add_spec_to_instance_spec.rb)

```ruby
#preparation
db['INSERT INTO deployments (id, name) VALUES (789, "deployment")'].insert

#assertions
expect(db.fetch('SELECT * FROM vms').first[:cid]).to eq('123')
```

Use ```assert { | db | # assertions }``` to get access to the ```db``` object from outside of ```during_migration```.

(1) Currently only the namespace 'director' is supported.
(2) You can run ```during_migration``` in a ```before``` block and put your expectations in separate ```it``` blocks

### DIFF
--- a/bosh-director/spec/spec_helper.rb
+++ b/bosh-director/spec/spec_helper.rb
@@ -28,7 +28,7 @@ module SpecHelper
   class << self
     include BufferedLogger
 
-    attr_accessor :temp_dir
+    attr_accessor :temp_dir, :db
 
     def init
       ENV["RACK_ENV"] = "test"
@@ -84,8 +84,6 @@ module SpecHelper
       Sequel.extension :migration
 
       connect_database(@temp_dir)
-
-      run_migrations
     end
 
     def connect_database(path)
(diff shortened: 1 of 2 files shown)


cppforlife:
I believe we can close this now. I think one or more tests were added for migrations to the Director.

PR #1015, opened 2015-11-09
Allow bosh_cli to stop a job or all jobs in deployment

barthy1:
- User can stop all instances of the job just skipping index in the stop command.
- User can stop all jobs of the deployment skipping job name in the stop command.

--hard option means that corresponding vm(s) should be deleted
and disappear from 'bosh vms'.

Signed-off-by: Alexey Miroshkin <alexey.miroshkin@ru.ibm.com>


bosh-director/lib/bosh/director/instance_updater/vm_updater.rb
```diff
@@ -55,8 +55,6 @@ def detach
 
       vm_deleter = VmDeleter.new(@instance, @vm_model, @cloud, @logger)
       vm_deleter.delete
-
-      @instance.job.resource_pool.add_idle_vm
```
mariash: why was this removed?

barthy1: It was mentioned in the description of task: "bosh stop --hard is useful for releasing all of the compute resources without losing any persistent data". Idle vm keeps resources to use.
Additionally, the task was "bosh stop --hard job1 should no longer show that vm in bosh vms"  - this idle vm is showed in bosh vms.

mariash:
merged here: https://github.com/cloudfoundry/bosh/commit/43b2ab5848ef0cf614604fee1a49363205559e5b Changed :all to '*' to be consistent with other commands.

PR #1016, opened 2015-11-09
User can cancel while drain action is running

hashmap:
This change makes drain action cancellable in bosh. There is a separate
change for bosh-agent.

[#106642152] (https://www.pivotaltracker.com/story/show/106642152)

### DIFF
--- a/bosh-director/lib/bosh/director/instance_updater/stopper.rb
+++ b/bosh-director/lib/bosh/director/instance_updater/stopper.rb
@@ -48,9 +48,6 @@ module Bosh::Director
       drain_time = initial_drain_time
 
       loop do
-        # This could go on forever if drain script is broken, canceling the task is a way out.
-        @config.task_checkpoint
-
         wait_time = drain_time.abs
         if wait_time > 0
           @logger.info("`#{@instance}' is draining: checking back in #{wait_time}s")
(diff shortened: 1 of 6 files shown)


bosh-director/lib/bosh/director/instance_updater/stopper.rb
```diff
@@ -48,9 +48,6 @@ def wait_for_dynamic_drain(initial_drain_time)
       drain_time = initial_drain_time
 
       loop do
-        # This could go on forever if drain script is broken, canceling the task is a way out.
-        @config.task_checkpoint
-
```
cppforlife: still want to keep this for older agents.

hashmap: @cppforlife my idea is to have this check in agent_client.drain only, why to keep this one? there is a spec for this change (stopper_spec.rb, the last one on this page). not sure it's related to agent version, it will be cancelled in bosh anyway and gets an error "cancel is not supported for this action" from older agents.

cppforlife: fair enough.

hashmap: @cppforlife thanks for raising it, found i forgot to include a change in 'instance_deleter' to my commit.

PR #1020, opened 2015-11-11
update s3_blobstore_client to support use AWS China S3 bucket

leizhu:
As you know, network speed is very low to access s3 blobstore based on US region from AWS China region, so it's a nightmare when uploading release in AWS China.

We suggest to use AWS China S3 buckets as the blobstores when doing bosh deployments in AWS China, which will speed up a lot. 

So we make a PR to make s3_blobstore_client support China S3 bucket, when uploading bosh release, we just make below change in config/final.yml: 
```
---
final_name: test
blobstore:
  provider: s3
  options:
    bucket_name: S3_BUCKET_IN_CHINA
    s3_endpoint: https://s3.cn-north-1.amazonaws.com.cn
```

Signed-off-by: Ning Fu <nfu@pivotal.io>

### DIFF
--- a/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
+++ b/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
@@ -31,7 +31,7 @@ module Bosh
         aws_options = {
           use_ssl: @options.fetch(:use_ssl, true),
           s3_port: @options.fetch(:port, 443),
-          s3_endpoint: @options.fetch(:host, URI.parse(S3BlobstoreClient::ENDPOINT).host),
+          s3_endpoint: @options.fetch(:host, URI.parse(@options[:s3_endpoint] || S3BlobstoreClient::ENDPOINT).host),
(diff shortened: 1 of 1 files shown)


cppforlife:
There is already option called `host` (same line you modified) that allows to set S3 endpoint. Is that not sufficient?

leizhu:
@cppforlife yes, I tried 'host' option, but it's not sufficient for s3 blob read_only mode..

cppforlife:
How about `endpoint` if it's for readonly?

leizhu:
@cppforlife I just tried the 'endpoint', yes, it's sufficient for readonly mode.
So that we could not need to add this PR, just use below config for both read_only and none_read_only mode:
```
---
final_name: cf-mysql
blobstore:
  provider: s3
  options:
    bucket_name: mysql-service-blobs
    host: s3.cn-north-1.amazonaws.com.cn
    endpoint: https://s3.cn-north-1.amazonaws.com.cn
```

dpb587-pivotal:
Closing since there are existing workarounds.

PR #1025, opened 2015-11-16
Prohibit the reuse of passwords with 24 iterations

xingzhou:
[#98978100]


cppforlife:
@xingzhou Does this prevent agent restarts after 24 times since agent sets passwords?

PR #1026, opened 2015-11-16
Changes for building a ppc64le stemcell for Power8 Ubuntu LE

hashmap:
This PR introduces support of stemcell creation on ppc64le platform on Ubuntu. Main changes:

- stemcell name includes arch name on ppc
- spec exclusion tag 'exclude_on_ppc64le' 
- 2 partitions on ppc, first is PReP (required for boot)
- bosh_micro_go agent step is disabled
- system_ixgbevf module is disabled 
- ppc64le ubuntu repos are added
- news assets config.guess and config.sub, also code which patches some packages with outdated config.guess and config.sub, otherwise 'configure' tool can't recognize ppc64le hardware platform
- changes in GRUB configuration
- rsyslog 8 is not supported yet


hashmap:
Re-sending this pr against develop

PR #1032, opened 2015-11-18
Bump to latest S3 CLI

calebamiles:
Users should not have to specify region or host when using AWS S3 for buckets located in the us standard region (us-east-1)

[#108364206](https://www.pivotaltracker.com/story/show/108364206)


calebamiles:
accidentally opened against master

PR #1040, opened 2015-11-25
Allow bosh_cli to start/restart/recreate a job or all jobs in deployment

barthy1:
[#108399300](https://www.pivotaltracker.com/story/show/108399300)

Signed-off-by: Alexey Miroshkin <alexey.miroshkin@ru.ibm.com>


bosh-ci-push-pull:
it would be great to also include a stronger assertion in integration tests on the output to see if instance was updated or not in the task output. currently these tests dont check that director did anything.

bosh-ci-push-pull:
well missed the checks for vm cid and state changes. may be then just add a check that only one cid changed when doing recreate of a single vm? and check that starting a single vm doesnt start them all?

PR #1046, opened 2015-11-30
Bosh uses db instead of redis to keep locks, task state

hashmap:
This patch implements DB backend for locks, includes:

- New Sequel Model and migration script
- Changes in lock.rb and tests
- Changes in Lock Controller and tests

Also this patch replaces Resque.enqueue and Resque workers with DJ, includes:
- migration script to support DJ Sequel model
- changes in director_service.rb to support monitoring worker state in DB
- changes in bosh-director-drain-workers and new bosh-director-worker-dj
- DBJob class as a layer between Director jobs and DJ. DJ requires to enqueue real job instances,
      Director job runner works with job class names. DBJob allows to solve this problem initially,
      having job class name as instance variable. JobQueue and tests were changed accordingly
- change in bosh release building logic was required (GemComponents),
      because link to delayed_job_sequel as git source in Gemfile won't lead to generate automatically gem.
      In process of copying gems if any gem doesn't exist, it tries to get gem_path and build this gem from source

Also all references to Redis were removed from release and config files.

[#106964214] (https://www.pivotaltracker.com/story/show/106964214)


bosh-director/lib/bosh/director/job_queue.rb
```diff
@@ -2,13 +2,15 @@
...
 
-  # Abstracts the resque system.
+  # Abstracts the delayed jobs system.
 
   class JobQueue
     def enqueue(username, job_class, description, params)
       task = Api::TaskHelper.new.create_task(username, job_class.job_type, description)
 
-      Resque.enqueue(job_class, task.id, *params)
+      Delayed::Worker.backend = :sequel
```
cppforlife: should this be set everytime we enqueue?

PR #1057, opened 2015-12-04
X Windows must not be enabled unless required.

zhang-hua:
[#98978178]

Signed-off-by: Zhou Xing <xingzhou@cn.ibm.com>


cppforlife:
@zhang-hua instead of configuring files let's just check that x server is not installed.

zhang-hua:
@cppforlife , okay! will update it.

PR #1062, opened 2015-12-10
Fixing problem with skip_drain option for changing job(deployment) state

barthy1:
Earlier it worked only for changing job instance state.

See PR https://github.com/cloudfoundry/bosh/pull/1015

[#100244046](https://www.pivotaltracker.com/story/show/100244046)


cppforlife:
@barthy1 Any tests that can be added to catch this?

PR #1065, opened 2015-12-11
Show error message if static ip is not found.

fabianschwarzfritz:
When a use selects a static ip that is not within the network's subnets
we show an error message.

### DIFF
--- a/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
@@ -45,7 +45,11 @@ module Bosh::Director
       def reserve(reservation)
         reservation.reserved = false
         if reservation.ip
-          find_subnet(reservation.ip) do |subnet|
+          subnet = find_subnet(reservation.ip)
+          puts subnet.class
+          if subnet.nil?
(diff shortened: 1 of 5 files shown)


fabianschwarzfritz:
Hi cf folks,
I cannot see the codeclimate result, how can I see what the issue is?
Cheers,
Fabian

cppforlife:
You cannot see codeclimate result after logging in? (I wouldnt bother doing much about codeclimate result as it is something that we use as a global metric).

We are also about to merge change to networking which makes this error show up as:

```
Are you sure you want to deploy? (type 'yes' to continue): yes

Director task 217
  Started preparing deployment > Preparing deployment. Failed: Job 'db3' with network 'private' declares static ip '11.10.64.124', which belongs to no subnet (00:00:00)

Error 150002: Job 'db3' with network 'private' declares static ip '11.10.64.124', which belongs to no subnet

Task 217 error
```

Is that acceptable error message in your opinion?

fabianschwarzfritz:
Hi @cppforlife ,
thanks for the response! I think the error message is all right :smile: . 
I'll close this pull request then.
Cheers,
Fabian

PR #1081, opened 2015-12-29
user can restore director db from a backup

xingzhou:
[#107648464]


bosh_cli/lib/cli/client/director.rb
```diff
@@ -602,6 +632,14 @@ def upload_and_track(method, uri, filename, options = {})
           file.stop_progress_bar if file
         end
 
+        def upload_without_track(uri, filename, options = {})
+          file = FileWithProgressBar.open(filename, 'r')
+          status, _ = post(uri, options[:content_type] || {}, file, options[:headers] || {}, options)
```
cunnie: We notice that you allow for `options[:headers]`, but, based on the two cases where the method `upload_without_track` is called, it is called without options (i.e. `options[:headers]` will always be nil)

bosh_cli/lib/cli/client/director.rb
```diff
@@ -554,6 +554,36 @@ def fetch_backup
           path
         end
 
+        def restore_db(filename, options={})
```
cunnie: We also noticed that although options are allowed here, they are not passed in either existing use case, always defaulting to an empty hash.

cunnie:
Hi @xingzhou & @zhang-hua ,

Thanks for the pull request!

* we liked how you adhered to the standards (e.g. marking messages red "DESTRUCTIVE OPERATIONS")
* we also liked your tests
* we noticed that you allowed for additional options for the post request. we added inline comments to this PR and we will merge in additional commits to iterate over these points.
* you don't need to do more work—we'll take it from here!

Thanks,

@wcamarao & @cunnie

xingzhou:
@wcamarao & @cunnie

Long time no see:), this is Tom from IBM, thanks for your review:)

for the options comment, you will help to modify the code and merge the PR finally, am I understanding correctly? hehe, thanks for your time:)

wcamarao:
Hi Tom, ltns.

Yes that's correct, we're taking it from here.

Thanks

release/jobs/director/templates/restore-db
```diff
@@ -0,0 +1,76 @@
...
+    if [ -z "$(monit summary | grep '^Process.*registry.*not monitored$')" ];then
+       continue
+    fi
+
+    break
+done
+
+if [ "$adapter" = 'postgres' ];then
+    export PGPASSWORD=$pass
+    monit restart postgres
```
wlindner: Why is it necessary to restart postgres before restoring the db?

PR #1089, opened 2016-01-11
director automatically removes compilers from the stemcell by default…

xingzhou:
… and keeps them on compilation vms

[#111025994]


cunnie:
Hi xingzhou, 

Thanks for sending this PR. 
Question: How did you generate the centos_dev_tools_file_list and ubuntu_dev_tools_file_list? We're concerned this list will become outdated.

Brian & Tyler

tylerschultz:
@xingzhou and @guoger, this commit shows up as merged, but that is not actually the case (none of the changes ever made it to develop due to a bad merge). Please see @cunnie's comment above, and submit a new pull request with the generator commands (rather than the generated file).

Sorry about that!

dpb587-pivotal:
# ubuntu

## cleaner

    dpkg-query -L $package_name | xargs file | grep -Ev ':\s+directory\s+$' | awk -F ':' '{ print $1 }'


## derived package list

from the oriringally committed files

 * binutils
 * bison
 * build-essential
 * cmake
 * cpp
 * cpp-4.8
 * debhelper
 * dkms
 * dpkg-dev
 * flex
 * g++
 * g++-4.8
 * gcc
 * gcc-4.8
 * gettext
 * intltool-debian
 * libmpc3
 * make
 * patch
 * po-debconf

based on

 	( while read file ; do dpkg -S $file | awk -F ':' '{ print $1 }' ; done ) < /tmp/ubuntu_dev_tools_file_list | sort | uniq

re-generating the list from this package list results in the same size list


# centos

## cleaner

    rpm -ql $package_name | xargs file | grep -Ev ':\s+directory\s+$' | awk -F ':' '{ print $1 }'

## derived package list

from the oriringally committed files

 * autoconf
 * bison
 * cmake
 * cpp
 * flex
 * gcc
 * gcc-c++
 * gcc-gfortran
 * gettext
 * intltool
 * libmpc
 * libquadmath-devel
 * libstdc++-devel
 * make
 * patch

based on

    ( while read file ; do for package in $( rpm -qf --qf "%{NAME}\n" $file ) ; do if rpm -q $package > /dev/null 2>&1 ; then echo $package ; fi ; done ; done ) < /tmp/centos_dev_tools_file_list | sort | uniq

re-generating the list from this package list results in 824 more files (diff in https://gist.github.com/dpb587-pivotal/ea6114fd4d3cd1c06ed6; additional files seem sane, though)

PR #1096, opened 2016-01-13
Force utf-8 when JSON-encoding CPI results / errors

filefrog:
This fixes a problem noticed in a production vSphere environment where
snapshot tasks were being met with HTTP responses including some sort of
arbitrary binary (believed to be Unicode) data.

See https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/issues/5

### DIFF
--- a/bosh_cpi/lib/bosh/cpi/cli.rb
+++ b/bosh_cpi/lib/bosh/cpi/cli.rb
@@ -99,7 +99,7 @@ class Bosh::Cpi::Cli
         message: message,
         ok_to_retry: ok_to_retry,
       },
-      log: @logs_string_io.string,
+      log: @logs_string_io.string.force_encoding(Encoding::UTF_8),
     }
     @result_io.print(JSON.dump(hash)); nil
   end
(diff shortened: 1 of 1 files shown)


filefrog:
CLA: should be signed as a member of Stark & Wayne (also, I think I signed a personal one myself)

Updates: Looks like the affected vCenter has every possible integration (including some 3rd-party SAN solution) enabled.  Per VMware, those plugins and integrations are essentially microservices that vCenter combines in the API that BOSH queries, so any one of these (including those not developed inside of VMWare proper) could be returning output with mixed encodings, leading to this bug.  Unfortunately, this has made it difficult to reproduce in environments outside of the original failing one, due to licensing and hardware constraints.

I will keep looking for a way to reproduce this.

Possibly related: http://kb.vmware.com/selfservice/microsites/search.do?language=en_US&cmd=displayKC&externalId=1791

rnandi:
Merged to develop https://github.com/cloudfoundry/bosh/commit/af16716928e28ef05f644e1b323bdbd40448bdc5

Thanks for the pull request!

cppforlife:
Fixed in v248 of bosh-release.

Issue #1098, opened 2016-01-14
make EM.threadpool_size configurable


leizhu:
Due to network latency, the sending speed of Bosh metrics is slow from China region to Datadog server.
After a period,  the time lag is so big  that Datadog drops any new incoming metric.

The possible fix is to change the following file to accept a config param for EM.threadpool_size
bosh_monitor/lib/bosh/monitor/runner.rb

dpb587-pivotal:
Closing - thanks for PR #1102 which was merged and fixed this issue!

PR #1101, opened 2016-01-20
add support for em threadpool_size property

leizhu:
This is the pull request for the issue https://github.com/cloudfoundry/bosh/issues/1098

The changes will allow a configurable EM threadpool_size so that users can set large value to increase throughput. 

Signed-off-by: Simon Gao <sgao@pivotal.io>


leizhu:
We're using a new branch, so close this PR

Issue #1133, opened 2016-02-16
Urgent issue: BOSH migration always triggers disk migration


mattcui:
We are trying to migrate the old bosh (1.2751.0) to the new version (1.3153.0), but we found there always triggered disk migration even though the disk setting is not changed in deployment yaml. We got the root casue that in the old bosh director database, there isn't "cloud_properties_json" column in persistent_disks table, for examples:

bosh=> select \* from persistent_disks;
 id  | instance_id | disk_cid |  size  | active
-----+-------------+----------+--------+--------
 198 |         118 | 6320765  |  20480 | t
 199 |         134 | 6321235  |  20480 | t

After migration to the new BOSH director, the table schema of persistent_disks is change, cloud_properties_json column is added and the value is nil, for examples: 

bosh=> select \* from persistent_disks;
 id | instance_id | disk_cid | size  | active | cloud_properties_json
----+-------------+----------+-------+--------+-----------------------
 198|         118 | 6320765  | 20480 | t      | 
 199|         134 | 6321235  | 20480 | t      | 

But the value is an empty list {} (marshaled to a string) if there is no property specified in disk_pool. So during the migration of CF deployment, the director tried to compare the value ({}) defined in deployment yaml and the value (nil) in persistent_disks table, since {} is not equal to nil, the disk migration is triggered. it's not acceptable to us since we have lots of deployment to be migrated, it really waste much time and would bring long outage especially if the disk migration fails on some core nodes, like CCDB, UAADB and etc. I think one solution to fix this issue is to set "{}" as the default value when initializing (changing the table schema) persistent_disks table. It's an urgent issue to us, we hope to get it solved ASAP before we move forward with migration on other deployments. Thanks.

@cppforlife  @maximilien

maximilien:
@mattcui can we submit a PR for this? Should be relatively easy and since you are already testing this.

Let me know if you need help. Best,

max

mattcui:
@maximilien We "hacked" (updated) persistent_disks table, adding {} to cloud_properties_json column when we migrated our deployment. I am trying to solve the problem in BOSH director code, but hope BOSH community to confirm if it's a problem or they have better solution.

maximilien:
@mattcui submit the PR (include unit tests) and that will start the discussion. Thanks.

mattcui:
Submitted the [PR 1137](https://github.com/cloudfoundry/bosh/pull/1137)

maximilien:
[...]

PR #1134, opened 2016-02-16
bosh events command should show deployment events

barthy1:
Event records contain:
-  name of the deployment
-  action (create, update, delete)
-  state of the action started/done/error/cancelled
-  error message if exists
-  task id
-  timestamp

[#112797557](https://www.pivotaltracker.com/story/show/112797557)

Signed-off-by: Peter Kalambet <peter.kalambet@ru.ibm.com>

PR #1137, opened 2016-02-18
Set default value {} in cloud_properties_json column

mattcui:
Submitted this PR to address [issue 1133](https://github.com/cloudfoundry/bosh/issues/1133), I already verified this change, I can see the default value "{}" set in the  newly added column cloud_properties_json.

 id | instance_id | disk_cid |  size  | active | cloud_properties_json
----+-------------+----------+--------+--------+-----------------------
  1 |           2 | 8091253  | 524288 | t      | {}
  2 |           4 | 8091423  | 153600 | t      | {}


mattcui:
Travis-ci failed, it should not be a problem in my change, the same failure occurred in the previous [PR
1134](https://github.com/cloudfoundry/bosh/pull/1134)
https://travis-ci.org/cloudfoundry/bosh/builds/109647679

bosh-ci-push-pull:
you ll have create a new migration since once migration is published it cannot be modified. 

please also add a test for the migration to make sure we dont override existing data (eg https://github.com/cloudfoundry/bosh/tree/master/bosh-director/spec/unit/db)

mattcui:
@bosh-ci-push-pull 
I just need to replicate the content of 20141204234517_add_cloud_properties_to_persistent_disk.rb to a new file, or I should assume the column cloud_properties_json is already added by 20141204234517_add_cloud_properties_to_persistent_disk.rb, in my code, I just need to set {} if the value is "" (nil) in the column? Thanks.

cppforlife:
@mattcui it's already added. this is a new migration from the current (latest) db schema to whatever the new default is.

mattcui:
@cppforlife Seems test cases operating on the same database are polluted by each other, after adding my db migration and unit test, the unit test for "AttachDisk" always failed, you can see the error in the [travis log](https://travis-ci.org/cloudfoundry/bosh/builds/113596086). I remember it failed even I removed my test case (still retain migration db script). Do you have any idea on how to fix that? Thanks.

@maximilien

cppforlife:
@mattcui Update looks good. We'll pull it in for the next release.

cppforlife:
Story: https://www.pivotaltracker.com/story/show/115721221

wlindner:
Thanks for the PR - we gave it a look and started to merge, but noticed a couple issues when running the full test suites...

* The travis failure above is due to unit test environment. We fixed it with the following if you want to apply the patch:
```
diff --git a/bosh-director/spec/unit/jobs/attach_disk_spec.rb b/bosh-director/spec/unit/jobs/attach_disk_spec.rb
index a01a363..adf21b6 100644
--- a/bosh-director/spec/unit/jobs/attach_disk_spec.rb
+++ b/bosh-director/spec/unit/jobs/attach_disk_spec.rb
@@ -62,7 +62,7 @@ module Bosh::Director

         it 'marks the pre existing active persistent disk as inactive and orphans it' do
           attach_disk_job.perform
-          expect(instance_model.persistent_disks).to_not include(original_disk)
+          expect(instance_model.persistent_disks[0].disk_cid).to eq('fake_disk_cid')
           original_persistent_disk = Models::PersistentDisk[disk_cid: 'original-disk-cid']
           expect(original_persistent_disk).to be(nil)
```

* This new defaulting of fields within the database does not work in MySQL (it doesn't seem to support defaults of `TEXT` fields which the `cloud_properties_json` uses). We get the following error.
```
BLOB/TEXT column 'cloud_properties_json' can't have a default value (Sequel::DatabaseError)
```

If you're still interested in this PR being merged, please have a look and resolve those issues and then let us know once they have been pushed. Then we can take another look at merging. Thanks!

dpb587-pivotal:
Closing - if you're still interested in seeing these changes, please feel free to take another look at the feedback and reopen with changes. I'm guessing that by now most migrations have happened and this is much less of an issue.

maximilien:
@dpb587 are you saying this was never pulled in?

dpb587-pivotal:
Correct; and I don't see any similar or equivalent changes in the repo.

maximilien:
@mattcui can you chime in here and confirm we are good on this front. Thx

mattcui:
@maximilien We added it in our forked BOSH v1 release, I don't think there is the problem when migrating from BOSH v1 to v2 without this commit, not sure about the case from legacy BOSH to BOSH v2, but we should have no such case. So, I think it's OK to close it with unmerged.

PR #1142, opened 2016-02-24
upgrade datadog api to 1.21

yuanzhao:
Currently DadaDog Ruby API gem dogapi-1.6.0 is used by bosh-monitor. However, this version has a drawback in APIService class(in common.rb): the timeout value for HTTP session is not configurable:

    # Manages the HTTP connection
    def connect
      uri = URI.parse(@api_host)
      session = Net::HTTP.new(uri.host, uri.port)
      if 'https' == uri.scheme
        session.use_ssl = true
        # FIXME - turn off SSL verification for now until we can spend
        # some time figuring out how to find certs across platforms.
        # - matt 10/06/2011
        session.verify_mode = OpenSSL::SSL::VERIFY_NONE
      end
      session.start do |conn|
        conn.read_timeout = 5
        yield(conn)
      end
    end
This is needed as when sending from a remote place to Datadog, the connection read timeout value should be bigger, like 60. It is supported in the newer version of dogapi package. 

Signed-off-by: Yuan Zhao <yuzhao@pivotal.io>


yuanzhao:
Hi,

Please find the signed CLA in the attachment. Thank you!
image1
image2

chipchilders:
@yuanzhao if you work for Pivotal, can you please join and make public your membership in one of the following GitHub orgs? You should be covered by the Pivotal CLA.

pivotal
pivotal-cf
Pivotal-Field-Engineering
pivotal-golang

cppforlife:
Story: https://www.pivotaltracker.com/story/show/115721441

chipchilders:
CLA received. This is cleared for merge if technically accepted. cc/ @cppforlife

PR #1153, opened 2016-02-29
Update WALA to 2.1.3 for Azure.

AbelHu:


cppforlife:
@AbelHu what's in the update? why do we need to bump?

AbelHu:
@cppforlife New WALA supports new protocols on Azure. It also supports AzureStack.

cppforlife:
Does that mean older stemcells will stop working because they dont support newer protocol?

AbelHu:
The old stemcell still works now but in future it can not work after the protocols are switched.

Best Regards,
Abel HU

? 2016?3?3??08:59?Dmitriy Kalinin <notifications@github.com<mailto:notifications@github.com>> ???


Does that mean older stemcells will stop working because they dont support newer protocol?

-

bosh-ci-push-pull:
Do you have the timeline when protocol may change?

AbelHu:
@cppforlife Azure still supports the old protocols in future. But AzureStack only supports new protocols. In order to use the same stemcell both on Azure and AzureStack, we need to update WALA to 2.1.3.

cppforlife:
Got it. Story: https://www.pivotaltracker.com/story/show/115712317

PR #1159, opened 2016-03-03
Enable trim mode (<% -%>) for templating

liuweichu:
### Issue
This PR is for https://github.com/cloudfoundry/bosh/issues/1086

### Purpose
Enable trim mode for bosh ERB templating, for more cosmetic templating output.
No backward compatibility break.

PR #1164, opened 2016-03-04
Support trim characters when rendering templates

### LINKED ISSUE #1086: Bosh template should enable trim_mode
Hello,

For now, bosh template does not set trim_mode [(erb doc)](http://ruby-doc.org/stdlib-2.3.0/libdoc/erb/rdoc/ERB.html#method-c-new), which makes it hard to make a properly formatted output.

We wish to have bosh template supporting `<% -%>` modifier by setting `trim_mode = "-"` in `lib/bosh/template/renderer.rb` file of `bosh-template` lib.

This change should not break backward compatibility.

geofffranks:
BOSH job templates now support using the '-' character
to specify trimming  of newlines in the rendered ERB
templates.

Fixes #1086

### DIFF
--- a/bosh-template/lib/bosh/template/renderer.rb
+++ b/bosh-template/lib/bosh/template/renderer.rb
@@ -12,7 +12,7 @@ module Bosh
       def render(template_name)
         spec = JSON.parse(@context)
         evaluation_context = EvaluationContext.new(spec)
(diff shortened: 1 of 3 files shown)


geofffranks:
Apparently this is a duplicate of #1159

PR #1166, opened 2016-03-07
Update rsyslog_logrotate.conf

ChaosEternal:
turn rotation of syslog and messages really hourly.
on heavy load system , ha-proxy and other component will generate huge logs , only true hourly rotated syslog can avoid big disk usage.

### DIFF
--- a/stemcell_builder/stages/rsyslog_config/assets/rsyslog_logrotate.conf
+++ b/stemcell_builder/stages/rsyslog_config/assets/rsyslog_logrotate.conf
@@ -1,6 +1,7 @@
 /var/log/syslog
 {
 	rotate 7
+	hourly
 	size 5M
 	missingok
 	notifempty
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Hi @ChaosEternal - have you had a chance to verify this works as intended? The config also specifies `size 5M` which, if I recall correctly, takes precedence over time-based rotations.

Also, can you maybe explain your usage which is causing so many logs to end up in `/var/log`? Typically BOSH releases are configured to log everything in `/var/vcap/sys/log` (with a separate [default logrotate setting](https://github.com/cloudfoundry/bosh-agent/blob/9619b2579b79ee7c84b1665a8dba798256a5f4a3/agent/applier/applyspec/v1_apply_spec.go#L74) of 50M), which means the `/var/log` logs are fairly quiet.

cppforlife:
@ChaosEternal bump. thoughts on what @dpb587-pivotal said?

ChaosEternal:
I am arranging a detailed test to prove/disprove my idea.

dpb587-pivotal:
Closing due to inactivity and conflicts. Please feel free to reopen this PR once you have more details!

Issue #1169, opened 2016-03-14
Parallel uploads of stemcells


omazhary:
When uploading the same stemcell from different clients **at the same time**, a race condition occurs whereby only 1 client wins, and the others abort their entire process and give out the following error:

```
Stemcell `bosh-openstack-kvm-ubuntu-trusty-go_agent/3196' already exists. Increment the version if it has changed.
```

This is especially evident using bosh-workspaces. When multiple workspaces execute `bosh prepare deployment` at the same time, this race condition occurs.

The effect can be most felt when the workspaces are prepared and deployed using concourse pipelines. This error halts the pipeline's execution, thereby neutralizing its automated properties.

omazhary:
Hello everyone,

I've tried to get around to the problem myself, and I actually believe I've solved it. You can find it [here](https://github.com/omazhary/bosh/commit/dd6faf87b4a9b42a6abda7a827e49d5b16d84e86).

The problem is, it breaks the unit-tests that check for stemcell duplicates.

How would you propose to proceed?

Regards,
Omar

cppforlife:
We can change the unit tests for that too. We always wanted to change default behavior anyway. One problem I see though with parallel uploads is that it still will error most likely at the director side. We'll have to check if update_stemcell.rb background job deals with this properly.

omazhary:
Hi Dmitriy,

thanks for your response.

I've done a little checking. You have a point. I assume [this block](https://github.com/omazhary/bosh/blob/master/bosh-director/lib/bosh/director/jobs/update_stemcell.rb#L79-L91) will cause cause a problem.

Regards,
Omar

omazhary:
Hi Dmitriy,

I've submitted the pull request. I still, however, have not fixed the unit tests. Should I give it a shot as well, or is this something that should be done by the bosh team?

Regards,
Omar

jfmyers9:
Hi @omazhary,

Sorry that we haven't tackled this yet. You can find the corresponding story [here](https://www.pivotaltracker.com/story/show/118979067). Hopefully we will get to it soon.

Best,
@jfmyers9

s4heid:
Hi @omazhary,

thanks for submitting the issue. [This commit](https://github.com/cloudfoundry/bosh/commit/5e7131549937aba82a471a0516c5e92d4e8e87f5) should fix the problem.

Best regards,
@dpb587-pivotal && @s4heid

PR #1175, opened 2016-03-18
Parallel Stemcell Uploads

omazhary:
These commits should allow the client and director to receive duplicate stemcells, but ignore them without throwing exceptions and stopping the execution's flow. See [this issue](https://github.com/cloudfoundry/bosh/issues/1169) for more details.

### DIFF
--- a/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
+++ b/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
@@ -77,10 +77,12 @@ module Bosh::Director
         end
 
         stemcell = nil
+	      stemcell_exists = false
         track_and_log("Checking if this stemcell already exists") do
           begin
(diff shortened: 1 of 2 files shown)


bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -77,10 +77,12 @@ def perform
         end
 
         stemcell = nil
+	      stemcell_exists = false
         track_and_log("Checking if this stemcell already exists") do
           begin
             stemcell = @stemcell_manager.find_by_name_and_version @name, @version
-            raise StemcellAlreadyExists, "Stemcell '#{@name}/#{@version}' already exists" unless @fix
+            #raise StemcellAlreadyExists, "Stemcell `#{@name}/#{@version}' already exists" unless @fix
```
allomov: I think you can simply remove this line.

PR #1181, opened 2016-03-23
Changed job_state for not existing VMs

Kiemes:
- job_state actually reflects agent_state,
  therefore null instead of 'missing vm' is
  more appropriate

[#114100869](https://www.pivotaltracker.com/story/show/114100869)


Kiemes:
@beyhan

PR #1183, opened 2016-03-24
Parallel Stemcell Upload

omazhary:
- Implemented a fix on the bosh-cli side.
- Implemented a fix on the bosh-director side.
- Modified tests to accommodate fixes. This is however, not a clean solution, and will require a more permanent fix.

### DIFF
--- a/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
+++ b/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
@@ -77,10 +77,12 @@ module Bosh::Director
         end
 
         stemcell = nil
+	      stemcell_exists = false
         track_and_log("Checking if this stemcell already exists") do
           begin
             stemcell = @stemcell_manager.find_by_name_and_version @name, @version
-            raise StemcellAlreadyExists, "Stemcell '#{@name}/#{@version}' already exists" unless @fix
+            #raise StemcellAlreadyExists, "Stemcell `#{@name}/#{@version}' already exists" unless @fix
+            stemcell_exists = true
           rescue StemcellNotFound => e
             stemcell = Models::Stemcell.new
             stemcell.name = @name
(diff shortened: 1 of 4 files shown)


bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -77,10 +77,12 @@ def perform
         end
 
         stemcell = nil
+	      stemcell_exists = false
         track_and_log("Checking if this stemcell already exists") do
           begin
             stemcell = @stemcell_manager.find_by_name_and_version @name, @version
-            raise StemcellAlreadyExists, "Stemcell '#{@name}/#{@version}' already exists" unless @fix
+            #raise StemcellAlreadyExists, "Stemcell `#{@name}/#{@version}' already exists" unless @fix
```
allomov: I think you can simply remove this line.

omazhary:
Thanks for the suggestion!! Removed the commented out line.

cunnie:
Hi Omar,

Thanks for your pull request — being able to upload stemcells in parallel could be a useful feature.

Some notes:
- it appears that your pull request removes the ability to upload a stemcell with the `--fix` option (an option necessary to overwrite a corrupted stemcell). We'd like to preserve the existing behavior (i.e. `--fix` replaces the existing stemcell).
- commenting-out tests, as tempting as it may be, is not something we encourage. If a test is stale or no longer relevant, delete it. In this case, it appears that the tests are still relevant (i.e. uploading an existing stemcell will raise an error, uploading an existing stemcell with `--fix` replaces an existing stemcell)
- We'd like to see more coverage on the new feature, specifically corner cases (e.g. what happens when the same stemcell is uploaded simultaneously?). Admittedly writing such a test might not be trivial to write.
- We prefer fewer commit messages, rather than many. I use `git rebase -i ...` to squash my [often many] commits before submitting a PR.

Thank you,

—Brian

cppforlife:
We do also have a story for to tackle this: https://www.pivotaltracker.com/story/show/118979067

omazhary:
@cppforlife If you have a story for that, should I close this pull request?

dpb587-pivotal:
Closing - I believe this was fixed via [#130570721](https://www.pivotaltracker.com/story/show/130570721) and released in [v260](https://github.com/cloudfoundry/bosh/releases/tag/v260).

PR #1184, opened 2016-03-24
Make sure the progress bar goes to 100%

evanphx:
No one likes it when an operation ends and the progress bar never went all the way to 100%. @zachgersh was showing me some concourseci output and the progress bars always ended at 96%. Rather than sort out exactly why the chunks never match up exactly, just set the progress bar to 100% when we're done.

### DIFF
--- a/bosh_cli/lib/cli/file_with_progress_bar.rb
+++ b/bosh_cli/lib/cli/file_with_progress_bar.rb
@@ -32,6 +32,7 @@ module Bosh
         if result && result.size > 0
           progress_bar.inc(result.size)
         else
+          progress_bar.set(size)
           progress_bar.finish
         end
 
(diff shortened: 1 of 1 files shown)


zachgersh:
@cppforlife this is a major plus one and huge improvement.

evanphx:
@zachgersh I emailed in the CLA. The CLA was probably 3 to 4 times more text input that the actual fix.

ryanmoran:
:100:

cppforlife:
Story: https://www.pivotaltracker.com/story/show/116487229

dpb587-pivotal:
Merged to develop f6114d9 - thanks!

All active development on BOSH happens on the develop branch. The master branch is our way of tracking what has already been released and published on bosh.io. Please make future pull requests to develop. :)

Thanks!

PR #1199, opened 2016-04-08
Add a new feature flag which could switch the virtual delete_vm function when instance uprade

jianqiu:
When instance upgrade, bosh director will first delete vm from infrastructure then recreate the vm. Under  some special conditions, for example, in SoftLayer, we don`t want to *real* delete vm from Softlayer, bosh softlayer cpi would leverage the os_reload provided by Softlayer to upgrade the stemcell/instance vms.
So, I just submit this PR which including a new feature flag to enable virtual delete vm when instance upgrade.


jianqiu:
@cppforlife
Could you please review this PR which including a new feature flag to enable virtual deletion when doing instance upgrade? Thanks!

bosh-director/lib/bosh/director/vm_deleter.rb
```diff
@@ -17,7 +18,11 @@ def delete_for_instance(instance)
 
     def delete_vm(vm_cid)
       @logger.info('Deleting VM')
-      @error_ignorer.with_force_check { @cloud.delete_vm(vm_cid) }
+      @error_ignorer.with_force_check {
+        unless @enable_virtual_delete_vm
+          @cloud.delete_vm(vm_cid)
```
cppforlife: wouldnt this result in never deleting any vms? did you run this in your env?

jianqiu: The default value of @enable_virtual_delete_vm is false, so if you did not set @enable_virtual_delete_vm = true explicitly, bosh director will invoke @cloud.delete_vm to delete vm from infrastructure, otherwise, bosh will keep it.

allomov:
It seems to be CPI dependent feature. I think possible solution can be providing CPI with additional information about reason of VM removal to avoid bosh-director from having CPI dependent options.

jianqiu:
@allomov 
I quite agree with you, the os_reload or virtual deletion is specific for SoftLayer, however, CPI could not make a distinction between delete_vm of delete deployment and instance upgrade if only CID was passed to CPI. Do you have some good ideas to implement it in CPI side? Thanks!

allomov:
@jianqiu the idea is that bosh-director should provide context to CPI with the reason of VM removal.

jianqiu:
@allomov 
 It should be a big change to bosh-director to involve the reason of vm removal, but it seems that could be a general solution to this kind of requirements. Our solution might be a temporary solution that does not affect the normal function of other CPIs.
@cppforlife 
Do you think it is possible to make bosh-director provide the context of VM removal? Thanks!

jianqiu:
@cppforlife
We ever tested the PR in our pipeline, and passed the bats. So, could you please look at it? Thanks!

bosh-director/lib/bosh/director/vm_deleter.rb
cppforlife: Right so if enable_virtual_delete_vm=true then there is no way to delete VM ever. `bosh delete deployment` command will not delete VMs. I thought you actually wanted that.

jianqiu: @cppforlife 

Not quite correct, the PR is including not only the changes of vm_deleter.rb but also instance_updater.rb. So there are 3 scenarios:
1. For deployment upgrade / instance update, if enable_virtual_delete_vm=true, bosh director will not invoke delete_vm action of CPI, that is what we want, because we can use os_reload to achieve instance upgrade.
2. For bosh delete deployment command, even if enable_virtual_delete_vm=true, the command will also delete vms.
3. For delete_vm resolution by bosh cck,  even if enable_virtual_delete_vm=true, the solution will delete that vm.
As a whole, the change will cause no impact to other CPI/infrastructure, such as, aws, openstack, vsphere. For Softlayer, if we want to enable the virtual delete action during instance upgrade, we will set enable_virtual_delete_vms = true in director.yml

jianqiu:
@mattcui, fyi

jianqiu:
@cppforlife @mattcui @maxim
The codes were polluted by the codes from master branch, so closing it and file another PR

allomov:
@cppforlife sorry for delay in response.

About the idea of how CPI can get context of operation that is currently happen: I can think of 2 ways, and both of them are debatable:

1. CPI can make request to BOSH director API and get the latest event that occurred with disk which has specific CID. This will require extend events endpoint to be able to filter events by resources like disks. Also this will require CPI developers to implement simple bosh director client. Both options don't seem to be hard.
1. CPI interface can be extended. In order to extend CPI interface easier, the way to get version of supported CPI interface should be implemented. This seems to be trickier and less at this moment, not so required (I mean there won't be lots of CPI interface changes).

I would vote for 1st option. What do you think?

PR #1203, opened 2016-04-12
Changes for proper support `micro_bosh_eanbled` variable for `ppc64le`

kalambet:
This PR contains two types of changes separated in two commits.

The first one related to stemcell build process. Disabling the `micro_bosh` in stemcell build leads to problem with local blobstore because proper path does not exist in resulting stemcell so the `micro_bosh_go` stage just creates the path explicitly. Plus previously added excluding of `bosh_micro_go` and `bosh_ruby` build stages reverted back in favour of using `bosh_micro_enabled` environment variable.

The second commit contain changes in `bosh-release` build process. It update configuration for old PostgreSQL sources to support `ppc64le` architecture and also update bundler configuration for using system `xslt` and `xml2` libs in case of `ppc64le` build.

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
@@ -37,7 +37,7 @@ module Bosh::Stemcell
         :aws_cli,
         :logrotate_config,
         :dev_tools_config,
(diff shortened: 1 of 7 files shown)


voelzmo:
@kalambet If it is two unrelated changes, can we have those as two separate PRs? This makes reasoning and discussing easier. Thanks!

kalambet:
@voelzmo sure, I'll separate them. thanks!

PR #1205, opened 2016-04-12
Update bundler configuration in bosh-release build for `nokogiri` build for `ppc64le`

kalambet:
Update bundler configuration in bosh-release build for `nokogiri` build for `ppc64le`
Update postgresql config with the new `config.guess` and `config.sub` for `ppc64le` support

### DIFF
--- a/release/packages/director/packaging
+++ b/release/packages/director/packaging
@@ -20,6 +20,10 @@ EOF
 
 bundle_cmd="/var/vcap/packages/ruby/bin/bundle"
 
+if [ "`uname -m`" == "ppc64le" ]; then
+    $bundle_cmd config build.nokogiri '--use-system-libraries'
+fi
+
 $bundle_cmd config build.mysql2 \
   --with-mysql-config=$mysqlclient_dir/bin/mariadb_config \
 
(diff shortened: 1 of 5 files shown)


kalambet:
This PR is a first part of separated https://github.com/cloudfoundry/bosh/pull/1203

cppforlife:
Would BOSH_MICRO_ENABLED env variable make last 2-3 PRs unnecessary?

kalambet:
@cppforlife not sure about that. Last 2 PRs aimed on propper BOSH_MICRO_ENABLED support for ppc64le stemcell build. 

Before this changes in case this variable is set to 'no' the cache dir for local blobstore is not created, so bosh-agent fails on adding blob to the blobstore. That is the reason for mkdir.

Plus since this var is a propper way to disable 'bosh-ruby' and 'bosh_micro_go' micro stages I removed exceptions form the stages lists.

So current build of ppc64le supported stemcell executes with this var set to 'no'.

But all those descriptions are for other bosh PR. In this PR we are changing the way bosh release created for ppc64le support, like nokogiri build and build postgresql.

kalambet:
Second commit removes `curl` for getting `config.sub` and `config.guess` for `posrgres` package and use locally stored versions.

kalambet:
As discussed with @cppforlife added changes for `rsyslog` execution in chroot.

PR #1207, opened 2016-04-12
Remove mistakenly repeated require statement in director's api.rb file

allomov:
The PR is described in the title ^^


allomov:
The error in CI does not seem to be connected with this PR.

PR #1208, opened 2016-04-12
Update eventmachine gem - again 

allomov:
Hey, all.

I have a problem with installation eventmachine 1.0.3 on Mac OS X 10.11.4 and xcode 7.3 with version of Ruby > 2.2.x (including 2.3.0).

You can find the whole error message in [this gist](https://gist.github.com/allomov-altoros/ebfe3a4ff0f80fa075d6021762c1895d).

Discussion of this error in https://github.com/eventmachine/eventmachine/issues/553 shows updating to eventmachine 1.0.4 as one of solutions of this problem. At the same time Gemfile [contains warning](https://github.com/cloudfoundry/bosh/blob/master/Gemfile#L64) not to update eventmachine before this issue https://github.com/eventmachine/eventmachine/issues/633 will be solved. Since this issue says that problem appears on eventmachine version 1.0.7 and higher, updating to using eventmachine 1.0.4 sounds reasonable for me.

What do you think on this?


dpb587-pivotal:
Thanks @allomov!

Assuming all the integration tests also pass with this, I think this should be merged. Our current story about upgrading to a newer version of Ruby blocks on eventmachine upgrade anyway and 1.0.4 looks like it at least compiles on ruby 2.3.0.

:+1: for integration testing and merge...

Issue #1211, opened 2016-04-13
bosh-director : Showing tasks for given deployment doesn't work if the deployment has been deleted


holgerkoser:
Hi,

the `bosh-director` allows to filter the list of recent tasks by different critiria (limit, state and deployment_name). If I for example list the latest 3 recents tasks I get the following result: 

``` bash
$ bosh tasks recent 3
Acting as user 'admin' on 'Bosh Lite Director'

+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
| #   | State | Timestamp               | User      | Deployment  | Description           | Result           |
+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
| 930 | done  | 2016-04-12 13:19:08 UTC | admin     | bar         | delete deployment bar | /deployments/bar |
| 927 | done  | 2016-04-12 13:13:44 UTC | admin     | bar         | create deployment     | /deployments/bar |
| 925 | done  | 2016-04-12 12:57:08 UTC | admin     | foo         | create deployment     | /deployments/foo |
+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
Showing 3 recent tasks
```

I see there are two deployments `foo` and `bar`. The deployment foo has been created but not deleted and the deployment `bar` which has been created and deleted afterwards. If I list the tasks for deployment `foo` I get the following result as expected:

``` bash
$ bosh tasks --deployment foo
Acting as user 'admin' on 'Bosh Lite Director'

+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
| #   | State | Timestamp               | User      | Deployment  | Description           | Result           |
+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
| 925 | done  | 2016-04-12 12:57:08 UTC | admin     | foo         | create deployment     | /deployments/foo |
+-----+-------+-------------------------+-----------+-------------+-----------------------+------------------+
Showing 1 recent tasks
```

But if I list the tasks for deployment `bar` I get an error that the deployment doesn't exist which is true but not the correct behaviour from my point of view. I would expect to see a list with the "create deployment" and "delete deployment bar" tasks.

``` bash
$ bosh tasks --deployment bar
Acting as user 'admin' on 'Bosh Lite Director'
Error 70000: Deployment `bar' doesn't exist
```

Best regards, Holger

allomov:
[...]

PR #1212, opened 2016-04-13
Fix task deployment filter when deployment is removed

allomov:
This PR is created to solve the this issue https://github.com/cloudfoundry/bosh/issues/1211

I am sure that user should have ability to filter tasks by deployment name, even if deployment is removed. If this is not so, then what is the reason to show this tasks in the list of all tasks.


cppforlife:
@dpb587-pivotal can you please update @allomov on the changes we are planning to make regarding this issue.

dpb587-pivotal:
Last week we had to do some refactorings on code which will affect this PR. In [#117557245](https://www.pivotaltracker.com/story/show/117557245) we started adding the deployment's teams onto the task's db record to allow deleted deployments to still be followed. That also means we can more cleanly handle the listings endpoint for deleted deployments like your PR deals with. Story [#117845011](https://www.pivotaltracker.com/story/show/117845011) has been added to the backlog which deals with fixing this listing/deleted deployment issue in light of the new tasks.teams field.

allomov:
@dpb587-pivotal thank you for the note. It seems that it will not be hard to follow this changes in this PR. 

One thing that bothered me was that tasks from removed deployment can be seen by everyone who has read access to director (even if this user was not in the scope of this project). The possible solution here could be soft deletion, which seems to lead to a lot of code changes.

cppforlife:
@allomov yup deletion always causes more problems then it solves imho. we are planning to add first class team support tasks -- each task belongs to set of teams -- which should eliminate this problem.

cppforlife:
I believe this was fixed in either v256 or v257.

PR #1219, opened 2016-04-18
Add a new feature flag which could switch the virtual delete_vm function when instance uprade

jianqiu:
When instance upgrade, bosh director will first delete vm from infrastructure then recreate the vm. Under some special conditions, for example, in SoftLayer, we don`t want to real delete vm from Softlayer, bosh softlayer cpi would leverage the os_reload provided by Softlayer to upgrade the stemcell/instance vms.
So, I just submit this PR which including a new feature flag to enable virtual delete vm when instance upgrade.


jianqiu:
@cppforlife @mattcui @maxim
A new PR to trace this enhancement.

PR #1221, opened 2016-04-18
Information about new/old releases and stemcells of the deployment is…

barthy1:
… stored in event context

Create/update deployment state is fixed.

[#117345977] (https://www.pivotaltracker.com/story/show/117345977)

Signed-off-by: Peter Kalambet <peter.kalambet@ru.ibm.com>


cppforlife:
@barthy1 after looking at the content, let's change it to two keys `before` and `after`. before/after keys should include stemcells/releases as they were/are. that way i think it will be a little easier to understand state transitions.

PR #1226, opened 2016-04-19
Customize BOSH agent settings for GCE stemcell

evandbrown:
**Important**: This PR depends on https://github.com/cloudfoundry/bosh-agent/pull/73 and should not be merged until that `bosh-agent` PR is merged to master.

This customizes *agent.json* on Google Compute Engine stemcells to use GCE instance metadata to retrieve BOSH settings, eliminating the need for the BOSH registry. `UseRegistry` is set to false, `UserDataPath` is removed (no need to bootstrap the registry connection info), and a default `SettingsPath` is provided where the agent can expect to find BOSH settings JSON.

### DIFF
--- a/stemcell_builder/stages/bosh_google_agent_settings/apply.sh
+++ b/stemcell_builder/stages/bosh_google_agent_settings/apply.sh
@@ -16,9 +16,9 @@ cat > $chroot/var/vcap/bosh/agent.json <<JSON
     "Settings": {
       "Sources": [
         {
-          "Type": "HTTP",
+          "Type": "InstanceMetadata",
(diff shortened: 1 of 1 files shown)


voelzmo:
@evandbrown out of curiosity: That means the GCE metadata service lets you change settings during runtime of a VM and not just once during boot time? That sounds awesome!

evandbrown:
@voelzmo That's correct! Metadata can be updated on running VMs. This allows us to eliminate the BOSH registry for the Google Compute Engine CPI.

voelzmo:
@evandbrown that's pretty sweet! Finally some reasonable decisions for IaaS metadata :)

PR #1251, opened 2016-05-04
Add a new feature flag enable_virtual_delete_vms, when it was enabled…

jianqiu:
When instance upgrade, bosh director will first delete vm from infrastructure then recreate the vm. Under some special conditions, for example, in SoftLayer, we don`t want to real delete vm from Softlayer, bosh softlayer cpi would leverage the os_reload provided by Softlayer to upgrade the stemcell/instance vms.
So, I just submit this PR which including a new feature flag to enable virtual delete vm when instance upgrade.

PR #1254, opened 2016-05-06
Check no less frequently than 15s during watch_time

CAFxX:
Especially stateful components require sometimes a pretty high max_*_watch_time (10-20 minutes) causing 1-2 minutes intervals between checks.
This change makes sure that checks are done at most every 15s, regardless of min_watch_time and max_watch_time.


CAFxX:
The travis CI seems broken as already noted in https://github.com/cloudfoundry/bosh/pull/1251#issuecomment-216736513

cppforlife:
thanks @CAFxX

dpb587-pivotal:
Merged - thanks!

PR #1255, opened 2016-05-06
Allow job and index in cli

mdelillo:
Most of the other bosh cli commands allow for either of the two formats:

```
bosh command job 0
bosh command job/0
```

This PR makes the rest of the commands support both formats:

```
bosh attach disk
bosh logs
bosh take snapshot
bosh snapshots
```

Paired with @mkenyon

### DIFF
--- a/bosh_cli/lib/cli/commands/disks.rb
+++ b/bosh_cli/lib/cli/commands/disks.rb
@@ -43,9 +43,12 @@ module Bosh::Cli::Command
 
     usage 'attach disk'
     desc 'Attaches an disk to an instance and replaces the current disk'
-    def attach(job_name, id, disk_cid)
+    def attach(job_name, id, disk_cid = nil)
       auth_required
       manifest = prepare_deployment_manifest
+
+      job_name, id, disk_cid = split_job⁄id(job_name, id, disk_cid)
+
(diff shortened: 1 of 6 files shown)


cppforlife:
@mdelillo unfortunate we didnt have a discussion before making this pr. we are trying to deprecate index from the ui of bosh and that's why we are not accepting it for new commands.

tomoe:
@cppforlife 
> we are trying to deprecate index from the ui of bosh 
Is there a discussion thread or design doc for that change so one can get more details on where it's going?

cppforlife:
UI deprecation of index is a result of https://github.com/cloudfoundry/bosh-notes/blob/master/finished/availability-zones.md

tomoe:
Seems to me that the change would be still useful after the change to uuid from numeric indexing. 
Could this patch be slightly changed and resurrected?

mkenyon:
@cppforlife This patch actually makes it nicer for unique id. It's not specific to indexes. There are a few commands that require `bosh subcommand job index_or_id` and they do not except `bosh subcommand job/index_or_id`. This patch fixes that inconsistency. The patch itself doesn't care whether the user uses the index or the id. I think that you should merge this.

PR #1257, opened 2016-05-09
Update misleading description for bosh stop hard command

ronakbanka:
``` Bosh stop job --hard ``` deletes the VM instead of power off .

### DIFF
--- a/bosh_cli/lib/cli/commands/job_management.rb
+++ b/bosh_cli/lib/cli/commands/job_management.rb
@@ -19,7 +19,7 @@ module Bosh::Cli
       usage 'stop'
       desc 'Stop all jobs/job/instance'
       option '--soft', 'Stop process only'
-      option '--hard', 'Power off VM'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bosh_cli/lib/cli/commands/job_management.rb [hunk 1]
```diff
@@ -19,7 +19,7 @@ def start_job(job = '*', index_or_id = nil)
       usage 'stop'
       desc 'Stop all jobs/job/instance'
       option '--soft', 'Stop process only'
-      option '--hard', 'Power off VM'
+      option '--hard', 'Deletes the VM'
       option '--force', FORCE
```
tomoe: May want to use imperative form (Delete without s at the end) to be consistent with the other messages.

ronakbanka: Fixed

tomoe:
LGTM

dpb587-pivotal:
Merged - thanks!

PR #1260, opened 2016-05-10
Fix failing Google stemcell spec

ljfranklin:
- GCE CPI recently switched to using a custom InstanceMetadata endpoint
  rather than the registry, but this test was not updated


dpb587-pivotal:
Caused by https://github.com/cloudfoundry/bosh/pull/1226

dpb587-pivotal:
Merged - thanks!

PR #1275, opened 2016-05-19
Updates for Bracket Computing

krishnanrs:
The main changes are:
1. Adding a sleep statement to workaround the aggressive disk attach timeout
2. Add gdisk package to support GPT paritioned root volumes
3. Add cloud-init to support cloud package automation

### DIFF
--- a/bosh-director/lib/bosh/director/disk_manager.rb
+++ b/bosh-director/lib/bosh/director/disk_manager.rb
@@ -19,6 +19,7 @@ module Bosh::Director
       disk = nil
       if instance_plan.needs_disk?
         disk = create_and_attach_disk(instance_plan)
+        sleep(30)
         mount_and_migrate_disk(instance, disk, old_disk)
       end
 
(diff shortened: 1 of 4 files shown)


allomov:
Hey, @krishnanrs.

This changes seem to be very cloud specific, so the best way to provide them will be developing your  external CPI for specifically for Bracket Computing infrastructure.

You can move delay for 30 seconds to `attach_disk` method of CPI and it will be executed before agent will get command to mount the disk.

Talking about changes in stemcell builder, they also should be decoupled in the other way, because this changes can fail on other IaaSes. For instance, you can build you pipeline that builds BRKT stemcell on base of current pipelines for stemcell building (which are [here](https://github.com/cloudfoundry/bosh/tree/master/ci/pipelines/stemcells)), you can patch stemcell builder inside this pipeline.

cppforlife:
sleeps should be replaced with agent.wait_until_ready if this is done due to rebooting of the machine. this could be useful to cpis that reboot machines after disk attachment. though agent will not properly startup after disk migration if this is the case. also unit tests should be added for such change.

regarding added pkgs please elaborate in detail why they are required and how they will be used.

bosh-agent conflicts with cloud init so im not sure we can accept this change as is. if you describe necessary functionality we may find a better way.

krishnanrs:
1. I'm submitting another change in the bosh-agent project, if approved would eliminate the need for the sleep statements.
2. Wasn't aware that cloud-init conflicts with bosh-agent. Will remove that package.
Will re-submit a new pull request while just including the gdisk package for ubuntu and centos.

PR #1278, opened 2016-05-20
Pass upload options to upload file method.

jmcarp:
This allows users to pass upload options to the s3 blobstore client via an `upload_options` key. I'm specifically looking to pass the `server_side_encryption` option for compliance purposes, but any options from http://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Object.html#upload_file-instance_method can be passed. This also adds a unit test asserting that the expected arguments are passed to `upload_file`.

cc @sharms @rogeruiz

### DIFF
--- a/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
+++ b/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
@@ -128,8 +128,12 @@ module Bosh
         raise BlobstoreError, "object id #{oid} is already in use" if object_exists?(oid)
 
         s3_object = Aws::S3::Object.new({:key => oid}.merge(@aws_options))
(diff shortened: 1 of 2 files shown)


cppforlife:
@jmcarp we have actually switched director to use s3cli (https://github.com/pivotal-golang/s3cli) so that it's more consistent with what agent uses to download upload/download blobs. if your intent is to make the director use this configuration, i think you will have to add this to s3cli.

jmcarp:
@cppforlife: thanks, I'll file an issue on s3cli. If s3cli eventually supports these options, would you be open to passing them through in `S3cliBlobstoreClient`?

jmcarp:
Replaced by #1299.

PR #1279, opened 2016-05-25
Separate worker for bosh vms and bosh ssh

barthy1:
[#119515921](https://www.pivotaltracker.com/story/show/119515921)

### DIFF
--- a/release/jobs/director/templates/worker_ctl.erb
+++ b/release/jobs/director/templates/worker_ctl.erb
@@ -25,7 +25,13 @@ export GEM_HOME=/var/vcap/packages/director/gem_home/ruby/2.3.0
 
 export TMPDIR=/var/vcap/data/tmp/director
 
-export QUEUE=normal
+export QUEUE="normal,urgent"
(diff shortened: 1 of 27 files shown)


release/jobs/director/templates/worker_ctl.erb
```diff
@@ -25,7 +25,13 @@ export GEM_HOME=/var/vcap/packages/director/gem_home/ruby/2.3.0
 
 export TMPDIR=/var/vcap/data/tmp/director
 
-export QUEUE=normal
+export QUEUE="normal,urgent"
+<% if_p('director.enable_dedicated_status_worker') do |enable_dedicated_status_worker| %><% if (enable_dedicated_status_worker && (p('director.workers', 3) > 1)) %>
```
dpb587: Minor feedback/FYI... since this property is defaulted to a value in the job spec file, we don't need to duplicate the default value of `3` here (no second param needed) – helps avoid duplicate defaults which could get out of sync and cause confusion over time.

barthy1:
@dpb587 thanks! The code is updated.

PR #1281, opened 2016-05-27
Namespace bosh_cli files under bosh/cli

alext:
With the way things are currently layed out, it's not possible to
install this and cf-uaac alongside each other. This is becaise both
have a `cli/config.rb` file, and therefore when bosh-cli does a 
`require 'cli/config'` it gets the uaac version instead of its own.

Fix this by namespacing all the files under `bosh/` to avoid this
collision. This also means that the filesystem layout matches the module
structure, and makes this gem better adhere to the recommended
rubygems patterns - http://guides.rubygems.org/patterns/#loading-code


alext:
The Travis failure looks to be unrelated to this change (the installed version of bundler doesn't match the `Gemfile`).

cppforlife:
We are trying to avoid making any non-small changes to the CLI as we are moving towards https://github.com/cloudfoundry/bosh-cli.

Issue #1267, opened 2016-05-16
Ambiguous error when using a non-existant VM extension


dpb587:
If you typo a `vm_extensions` value of an instance group to something which is not defined in cloud config, the deploy task errors with the following ambiguous error after it has finished compiling packages...

```
Error 100: undefined method `cloud_properties' for nil:NilClass
```

And the following backtrace in the task debug...

```
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance.rb:288:in `block in cloud_properties'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance.rb:287:in `each'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance.rb:287:in `cloud_properties'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance.rb:204:in `cloud_properties_changed?'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance_plan.rb:36:in `changes'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/instance_plan.rb:24:in `changed?'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/job_updater.rb:21:in `block in update'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/job_updater.rb:20:in `select'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/job_updater.rb:20:in `update'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/deployment_plan/multi_job_updater.rb:34:in `block (3 levels) in run'
[...]

PR #1287, opened 2016-06-05
Raise meaningful error if non-existant vm_extension is requested

dpb587:
I ran into [this error](https://github.com/cloudfoundry/bosh/issues/1267) again and was briefly confused before recognizing it. Here's a fix so I don't have to think about it next time. Instead of failing with the following during deployment...

    Error 100: undefined method `cloud_properties' for nil:NilClass

It now shows...

    Error 100: The vm_extension 'wrongname' has not been configured in cloud-config.

Also a commit to allow using the `fly:*` tasks with a Concourse that doesn't have a `fly-integration`-tagged worker. Pre-existing behavior is still default.


allomov:
@dpb587-pivotal hey. could you also check this PR https://github.com/cloudfoundry/bosh-init/pull/73 for additional validations in bosh-init? it covers error type that happens rather frequently.

PR #1296, opened 2016-06-10
Event ids should be timestamp instead of Integer

barthy1:
- id and parent_id have Time type
- timestamp column is removed
- fractional_seconds connection option is added for mysql2
- if duplicate/unique db error for event id appears, timestamp id should be slightly increased
- previous information from event table is deleted

[#116887061](https://www.pivotaltracker.com/story/show/116887061)


medvedzver:
Had to revert commit. Mysql unit tests are failing because mysql can't store fractional datetime.

This is the version of mysql we are using 
mysql  Ver 14.14 Distrib 5.5.49, for debian-linux-gnu (x86_64) using readline 6.3

https://dev.mysql.com/doc/refman/5.5/en/fractional-seconds.html
`However, when MySQL stores a value into a column of any temporal data type, it discards any fractional part and does not store it.`

PR #1299, opened 2016-06-14
Pass encryption options to s3cli.

jmcarp:
Note: these options won't have any effect unless s3cli is bumped to
>=0.0.34.

Replaces #1278.

### DIFF
--- a/blobstore_client/lib/blobstore_client/s3cli_blobstore_client.rb
+++ b/blobstore_client/lib/blobstore_client/s3cli_blobstore_client.rb
@@ -39,7 +39,9 @@ module Bosh
           credentials_source: @options.fetch(:credentials_source, 'none'),
           access_key_id: @options[:access_key_id],
           secret_access_key: @options[:secret_access_key],
-          signature_version: @options[:signature_version]
+          signature_version: @options[:signature_version],
+          server_side_encryption: @options[:server_side_encryption],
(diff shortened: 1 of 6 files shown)


cppforlife:
@jmcarp I think that properties also have to be added to the bosh release: in the spec (https://github.com/cloudfoundry/bosh/blob/master/release/jobs/director/spec#L260) and ERB template (https://github.com/cloudfoundry/bosh/blob/master/release/jobs/director/templates/director.yml.erb.erb#L135)

jmcarp:
Thanks for the tip @cppforlife. I added the properties to the files you mentioned, and bumped s3cli to the latest version.

jmcarp:
I just noticed that @coreyti added these options in https://github.com/cloudfoundry/bosh/pull/1307, so now I'm not sure if this is needed. I was guessing that we would need to pass encryption-related options in `s3cli_blobstore_client.rb`, which I did here, but if that's not necessary, then I can close this.

cc @cppforlife @dpb587-pivotal

cppforlife:
@jmcarp hmm. we did miss passing this into s3 cli. we'll pull this in. thank you.

dpb587-pivotal:
Merged in cec8c74e1be7f8730afe96c3ae669641ce612672 on `develop`. Thanks for your PR!

All active development on bosh happens on the `develop` branch. The `master` branch is used as a record of the latest release of bosh.

Thanks again! :)

PR #1304, opened 2016-06-15
Separate dj worker pool for each bosh director instance

barthy1:
[#119516383](https://www.pivotaltracker.com/story/show/119516383)

### DIFF
--- a/bosh-director/lib/bosh/director/jobs/db_job.rb
+++ b/bosh-director/lib/bosh/director/jobs/db_job.rb
@@ -31,8 +31,13 @@ module Bosh::Director
       end
 
       def queue_name
-        @job_class.instance_variable_get(:@queue) ||
-            (@job_class.respond_to?(:queue) and @job_class.queue)
+        if (@job_class.instance_variable_get(:@local_fs) ||
+            (@job_class.respond_to?(:local_fs) and @job_class.local_fs)) && !Config.director_pool.nil?
+          Config.director_pool
+        else
+          @job_class.instance_variable_get(:@queue) ||
+              (@job_class.respond_to?(:queue) and @job_class.queue)
+        end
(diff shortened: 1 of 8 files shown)


dpb587-pivotal:
This adds a fair bit of extra complexity that we have to maintain. Couldn't we just rely on the existing DJ queues mechanism, switching the local-dependent jobs to `@queue = :local` and append `local` when the workers see a director job colocated?

dpb587-pivotal:
Ah, I misunderstood there would be multiple directors running against the same database. I don't think my idea is relevant there. Sorry. Ignore!

bosh-director/lib/bosh/director/jobs/db_job.rb
```diff
@@ -31,8 +31,13 @@ def perform
       end
 
       def queue_name
-        @job_class.instance_variable_get(:@queue) ||
-            (@job_class.respond_to?(:queue) and @job_class.queue)
+        if (@job_class.instance_variable_get(:@local_fs) ||
+            (@job_class.respond_to?(:local_fs) and @job_class.local_fs)) && !Config.director_pool.nil?
```
medvedzver: May we use `&&` instead of `and`?
There is a difference: http://stackoverflow.com/questions/1426826/difference-between-and-and-in-ruby

bosh-director/lib/bosh/director/jobs/db_job.rb
```diff
@@ -31,8 +31,13 @@ def perform
...
+          Config.director_pool
+        else
+          @job_class.instance_variable_get(:@queue) ||
+              (@job_class.respond_to?(:queue) and @job_class.queue)
```
medvedzver: same here

PR #1309, opened 2016-06-20
Updates for Bracket Computing

krishnanrs:
Added gdisk package to support GPT partitioned root volumes.

This is a new pull request in lieu of a previously closed one (#1275).

### DIFF
--- a/stemcell_builder/stages/base_centos_packages/apply.sh
+++ b/stemcell_builder/stages/base_centos_packages/apply.sh
@@ -33,7 +33,7 @@ nfs-common flex psmisc apparmor-utils iptables sysstat \
 rsync openssh-server traceroute libncurses5-dev quota \
 libaio1 gdb libcap2-bin libcap-devel bzip2-devel \
 cmake sudo libuuid-devel parted NetworkManager e2fsprogs cloud-utils-growpart \
-xfsprogs"
+xfsprogs gdisk"
 pkg_mgr install ${packages} ${version_specific_packages}
 
 # Install runit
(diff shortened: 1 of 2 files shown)


cppforlife:
Can you explain a bit more about why we need gdisk? Who uses it? I dont believe bosh-agent requires it.

krishnanrs:
The bosh-agent runs growpart (https://github.com/cloudfoundry/bosh-agent/blob/master/platform/linux_platform.go#L283) on the root volume as part of the bootstrap process. On GPT partitioned volumes, gdisk is required for growpart to succeed. The problem is that on linux hosts where growpart exists but fails, then bosh-agent fails the bootstrap process.

dpb587-pivotal:
Merged in e77bcb226fc492e4d3847d18291638796e1b013a on `develop`. Thanks for your PR!

All active development on bosh happens on the develop branch. The master branch is used as a record of the latest release of bosh.

Thanks again! :)

krishnanrs:
@dpb587-pivotal when will this PR be merged into master or will be picked up in a bosh release?

dpb587-pivotal:
It'll be merged to master at our next major stemcell release. I don't currently have an ETA - we've been doing minor releases of 257 lately while some other feature work settles in `develop`. I'm sure you've already found it, but here's the [doc](https://github.com/cloudfoundry/bosh/blob/develop/bosh-stemcell/README.md) with some pointers for building stemcells if you wanted to build a local one with your changes and other recent commits for testing.

PR #1315, opened 2016-06-22
Provide manual network type to external cpi

sykesm:
Looks like bosh-init provides the network type to the e-CPI but the bosh director does not. Seems both should do the same thing.

Per quick chat with @cppforlife on slack, opening a PR.

Units and integration were run but I did not execute the BATS.

### DIFF
--- a/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
@@ -51,6 +51,7 @@ module Bosh::Director
         end
 
         config = {
+          "type" => "manual",
           "ip" => ip.ip,
           "netmask" => subnet.netmask,
           "cloud_properties" => subnet.cloud_properties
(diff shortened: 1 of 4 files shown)


dpb587-pivotal:
@cppforlife, I'd like to see this merged. Do you have any objections, barring me rerunning latest units/integrations? Provides better consistency in the external CPI API.

sykesm:
@cppforlife I'm hitting this again and it's sort of causing issues since the network type matters for what I'm doing now.

I opened it per our discussion in slack (where you asked me to create a PR) and it's never moved forward. If this is rebased, will it ever be merged? I don't want to spend time on it if it's not going anywhere.

Thanks.

cppforlife:
@sykesm fell off the truck. we'll rebase it.

cppforlife:
started merging... hunting down integration tests.

dpb587-pivotal:
Merged via 2a626af7b30f44289c0363000be16405c355525e

PR #1317, opened 2016-06-24
Task can be cancelled when director is waiting for the vm to be in sp…

barthy1:
…ecific state

[#121535555](https://www.pivotaltracker.com/story/show/121535555)


dpb587-pivotal:
units: https://main.bosh-ci.cf-app.com/builds/111067
integration: https://main.bosh-ci.cf-app.com/builds/111068

PR #1324, opened 2016-06-29
stemcell_builder/stages: Set correct file permissions on system_google_packages

evandbrown:
Google heavy stemcell tests fail without this.


dpb587-pivotal:
Thanks for switching to `develop`!

Issue #1234, opened 2016-04-21
registry hangs and is unresponsive: bosh deploy fails with execution expired


voelzmo:
We saw recently a few deployments failing with the error of `execution expired` during writing the settings for a VM into the registry. The registry stopped responding at all, there was nothing in the logs.

Even when being on the Director itself, calling the registry didn't work. Something like
```
# curl  http://localhost:25777/instances/bla/settings
```
should fail with an error saying `instance 'bla' not found`, but just hangs until it runs into a read timeout.

So we attached a `gdb` and looked at the backtrace, seeing that excon [couldn't open the `ssl_socket` and blocked forever in the rescue](https://github.com/excon/excon/blob/v0.45.4/lib/excon/ssl_socket.rb#L118-L123). Note that the documentation of [`IO.select`](http://ruby-doc.org/core-2.1.7/IO.html#method-c-select) states this piece of code as an explicit example on how to implement a **blocking read**

The `gdb` stacktracke:
```
(gdb) call (void) rb_backtrace()
	from /var/vcap/packages/registry/bin/bosh-registry:16:in `<main>'
	from /var/vcap/packages/registry/bin/bosh-registry:16:in `load'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/bosh-registry-1.3202.0/bin/bosh-registry:28:in `<top (required)>'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/bosh-registry-1.3202.0/lib/bosh/registry/runner.rb:18:in `run'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/bosh-registry-1.3202.0/lib/bosh/registry/runner.rb:34:in `start_http_server'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/thin-1.5.1/lib/thin/server.rb:159:in `start'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/thin-1.5.1/lib/thin/backends/base.rb:63:in `start'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/eventmachine-1.0.3/lib/eventmachine.rb:187:in `run'
	from /var/vcap/packages/registry/gem_home/ruby/2.1.0/gems/eventmachine-1.0.3/lib/eventmachine.rb:187:in `run_machine'
[...]

PR #1326, opened 2016-06-30
Bump bosh_openstack_cpi and switch to fog-openstack

beyhan:
In order to fix https://github.com/cloudfoundry/bosh/issues/1234 we needed to bump bosh_openstack_cpi and with that excon.

We have run the BATs with this change. We have seen three failures that seem to be unrelated, i.e. DNS problems and we had our sec groups misconfigured for DNS.

Otherwise things were looking good.

However as there are gem deps changes we would prefer one of you guys have a look and merge it.


tylerschultz:
LGTM
I like that only the needed fog gems are now required! (If there are problems, we'll have to sort them out when they're encountered.)

dpb587-pivotal:
Merged - thanks!

Also, it looks like the mime-types gem bump caused issues in our Ruby 1.9.3 tests, so we reverted *just the mime-types gem* in 91b3c4922f26c89fd7999d01afb301527c94c27d. It didn't look essential for the main fog/excon bumps, so this is just an FYI. Let us know if we need to discuss the mime-type revert further.

PR #1327, opened 2016-06-30
Blob_manager was opening files in text mode instead of binary mode.  …

dickeyf:
…This leads to blob's file corruption when blobs are retrieve when using bosh-cli from windows.

A fix that allowed by to use bosh create release from Windows.

I couldn't download blobs from S3 as blob_manager was corrupting the files by treating them as text file.  These files are binaries, hence, they should be opened as such.

### DIFF
--- a/bosh_cli/lib/cli/blob_manager.rb
+++ b/bosh_cli/lib/cli/blob_manager.rb
@@ -266,7 +266,7 @@ module Bosh::Cli
       checksum = file_checksum(blob_path)
 
       @progress_renderer.start(path, "uploading...")
-      object_id = @blobstore.create(File.open(blob_path, "r"))
+      object_id = @blobstore.create(File.open(blob_path, "rb"))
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Thank you for your pull request! There's currently some work and refactoring happening on the CLI side of things, so we'll be a bit delayed in taking a closer look at this. We'll update this PR once we're unblocked from those other CLI changes and are better able to consider testing and merging this.

dpb587-pivotal:
Ruby CLI is now frozen and we aren't planning on publishing any new versions in favor of further developing the new cli, so I'm going to close this. You can find the new CLI in the [bosh-cli](https://github.com/cloudfoundry/bosh-cli) repo - feel free to give it a try and provide feedback there. There's a Windows-related issue about producing binaries - https://github.com/cloudfoundry/bosh-cli/issues/53.

PR #1339, opened 2016-07-07
Change events id and parent_id type from Integer to Bignum

barthy1:
[#116887061](https://www.pivotaltracker.com/story/show/116887061)


dpb587-pivotal:
We ran units locally against mysql and postgresql to double check.

PR #1341, opened 2016-07-07
Avoid exception when invalid UTF-8 character is present in CPI Resp

ljfranklin:
- Using `force_encoding` will throw an exception if an invalid char is present,
  while encode + replace will replace all invalid chars with '�'

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>

[#124949631](https://www.pivotaltracker.com/story/show/124949631)

### DIFF
--- a/bosh_cpi/lib/bosh/cpi/cli.rb
+++ b/bosh_cpi/lib/bosh/cpi/cli.rb
@@ -99,7 +99,7 @@ class Bosh::Cpi::Cli
         message: message,
         ok_to_retry: ok_to_retry,
       },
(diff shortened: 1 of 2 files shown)


ljfranklin:
Once this is merged, we should publish a new CPI gem for the Ruby CPIs to consume.

medvedzver:
Test fails on Ruby 1.9.3:
https://main.bosh-ci.cf-app.com/pipelines/bosh/jobs/unit-1.9/builds/1020

Looks like the fix for this would be to add "# encoding: utf-8" at the beginning of a cli_spec.rb.
But then, this test fails with different error:
https://main.bosh-ci.cf-app.com/builds/131315

PR #1346, opened 2016-07-08
Document health monitor manifest testing process.

jmcarp:
New plugins should update existing tests in https://github.com/cloudfoundry/bosh/blob/master/release/spec/health_monitor.yml.erb_spec.rb and add new ones. Otherwise, existing tests will fail.


dpb587-pivotal:
Merged - thanks!

PR #1347, opened 2016-07-12
Expose `calculate_vm_cloud_properties` in BOSH CPI

forrestsill:
Allows users to specify VM resources (cpu, ram, ephemeral_disk_size) in an
IaaS-agnostic manner in their manifests

For example, instead of specifying `m3.large` in the manifest, one could
have the following `cloud_properties` stanza:

```yml
{
  "ram": 4096,
  "cpu": 4,
  "ephemeral_disk_size": 2048
}
```

[#126023253](https://www.pivotaltracker.com/story/show/126023253)

Signed-off-by: Brian Cunnie <bcunnie@pivotal.io>

### DIFF
--- a/bosh_cpi/lib/bosh/cpi/cli.rb
+++ b/bosh_cpi/lib/bosh/cpi/cli.rb
@@ -19,6 +19,7 @@ class Bosh::Cpi::Cli
     delete_snapshot
     get_disks
     ping
+    calculate_vm_cloud_properties
   ).freeze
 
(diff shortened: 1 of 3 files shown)


ljfranklin:
Once this is merged, we need to publish a new version of [the gem](https://rubygems.org/gems/bosh_cli/versions/1.3262.0). Not sure what the process is for that.

Also if you close and reopen the PR, DreddBot will show that Forrest has signed the CLA.

PR #1348, opened 2016-07-13
Support percentages in canaries and max in flight configurations

barthy1:
[#119159691](https://www.pivotaltracker.com/story/show/119159691)


cppforlife:
any reason to have 'size = nil' instead of just 'size'?

barthy1:
It's to be able to support `default_update_config` 
https://github.com/cloudfoundry/bosh/blob/7d727d83ce66b4b6c4902f9fbfdac1eb7ab35f15/bosh-director/lib/bosh/director/deployment_plan/update_config.rb#L52-L61

bosh-director/lib/bosh/director/deployment_plan/update_config.rb
```diff
@@ -95,6 +100,19 @@ def parse_watch_times(value)
...
+        return value if size.nil?
+        case value
+          when /^\d+%$/
+            [((/\d+/.match(value)[0].to_i * size) / 100).round, size].min
+          when /\A[-+]?[0-9]+\z/
+            value.to_i
+          else
+            raise 'cannot be converted'
+        end
+      end
```
zaksoup: This method has two possible return types, string or int, depending on the context. This seems like a big recipe for possible future confusion. We noticed lower down that there are tests that explicitly call out this strange behavior, but we're unsure why it's necessary in the first place. Can we make sure that all return values from `UpdateConfig.canaries(size)` are a single type?

bosh-director/lib/bosh/director/deployment_plan/planner_factory.rb
```diff
@@ -229,16 +229,16 @@ def update_mapped_properties(mapped_properties, property_path, value)
           return mapped_properties
         end
 
-        def parse_numerical_arguments arg
+        def parse_arguments arg
```
zaksoup: It looks like this method is public. Are there any existing unit tests for it? This functionality change looks like it would be good to have a unit test. This method also has two possible return types, string in the "%" case and int in the digit case. Can we refactor this to only have one return type and make sure that consumers of this method always properly handle strings as output?

zaksoup:
Hi @barthy1 

Thanks for this pull request. This functionality is very useful and will improve a lot of operators lives, I'm sure!

We've noticed a few inconsistencies with return types in some methods that had functionality added. We've commented where applicable for this issue. We're also concerned that changing all of the relevant unit tests to be strings actually hides the additional complication of multiple return types for the canaries and max_in_flight methods.

We'd love to pull this in once you apply these changes

Thanks!
Zak and @monkeyherder

zaksoup:
Thanks for the updates!

PR #1355, opened 2016-07-15
Fix "docker run" command to properly bind mount the bosh directory

aduffeck:
The --volume option takes the local source directory first and the
destination directory in the container last.


dpb587-pivotal:
:+1: - good fix. Team typically use these scripts from inside a vagrant from `ci/docker/Vagrantfile` where both vagrant and container will have `/opt/bosh`, so we overlooked this.

dpb587-pivotal:
Dev merge note - manually make it to develop.

PR #1357, opened 2016-07-19
Fixed nil exception for bosh events

knm3000:
[#122893411](https://www.pivotaltracker.com/story/show/122893411)

### DIFF
--- a/bosh-director/lib/bosh/director/models/event.rb
+++ b/bosh-director/lib/bosh/director/models/event.rb
@@ -7,12 +7,15 @@ module Bosh::Director::Models
     end
 
     def context
-      return {} if context_json.nil?
-      JSON.parse(context_json)
+      if context_json == 'null' || context_json == nil
+        {}
+      else
+        JSON.parse(context_json)
+      end
     end
 
     def context=(data)
-      self.context_json = JSON.generate(data)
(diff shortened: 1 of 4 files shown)


bosh-director/lib/bosh/director/models/event.rb
```diff
@@ -7,12 +7,15 @@ def validate
...
+      if context_json == 'null' || context_json == nil
+        {}
+      else
+        JSON.parse(context_json)
+      end
     end
 
     def context=(data)
-      self.context_json = JSON.generate(data)
+      self.context_json = data.nil? ? 'null' : JSON.generate(data)
```
dpb587-pivotal: Would it be simpler if we were just storing `{}` rather than `null`/`"null"`?

bosh_cli/lib/cli/commands/events.rb
```diff
@@ -41,6 +41,9 @@ def show_events
           row << event.fetch('task', '-')
           row << event.fetch('deployment', '-')
           row << event.fetch('instance', '-')
+          if !event.key?('context') || event['context'] == 'null' || event['context'] == nil
```
dpb587-pivotal: With the 4f5ef6a change, this could be simplified since the value will no longer be `null`.

zaksoup:
lgtm

PR #1367, opened 2016-07-27
fix parameter for deploy and recreate commands to recreate vms with u…

barthy1:
…nresponsive agents

[#119312571](https://www.pivotaltracker.com/story/show/119312571)


bosh_cli/lib/cli/commands/job_management.rb
```diff
@@ -104,6 +110,10 @@ def parse_manifest(operation)
           err('Cannot handle both --hard and --soft options, please choose one')
         end
 
+        if !fix_option_allowed?(operation) && fix?
+          err("--fix option only makes sense for 'recreate' operation")
```
dpb587-pivotal: If we control which commands have the `--fix` option, do we need to be performing this check as well - I thought the regular opts/args parser would error?

bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
```diff
@@ -80,6 +81,9 @@ def needs_recreate?
           if @recreate_deployment
             @logger.debug("#{__method__} job deployment is configured with \"recreate\" state")
             true
+          elsif @need_to_fix
```
dpb587-pivotal: I think it would be better for InstancePlan to check the state of the existing instance directly, similar to how it does all the other checks on the raw models/objects - this way we keep the logic centralized in this class and based on state. Otherwise we start relying on external, InstancePlan instantiators to perform the correct logic and correctly influence how instances are potentially going to be changing later. Maybe others have some thoughts though...

barthy1:
@dpb587-pivotal Information about unresponsive agent is necessary not only for `desired existing instance`, but also for `obsolete instance`, because attempt to stop it via agent will cause an exception. In order to be able to check the state of the obsolete instances I have to update obsolete instance plans with `instance` value.

bosh-director/spec/unit/api/controllers/deployments_controller_spec.rb
```diff
@@ -175,6 +175,17 @@ def manifest_with_errand(deployment_name='errand')
             end
           end
 
+          context 'with the "fix" param' do
+            it 'passes the parameter' do
+              allow_any_instance_of(DeploymentManager)
```
zaksoup: Why is this not an `expect` rather than `allow`? Looks like we are actually confirming that the `fix` param is passed into the `DeploymentManager` object.

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
```diff
@@ -359,6 +360,40 @@ module Bosh::Director::DeploymentPlan
...
+      context 'when instance is ok' do
+        let(:instance_plan) { InstancePlan.new(existing_instance: existing_instance, desired_instance: desired_instance, instance: instance, network_plans: network_plans, recreate_deployment: true) }
+
+        it 'should return false' do
+          expect(instance_plan.needs_to_fix?).to be_falsey
+        end
+      end
+
+      context 'when instance is nill' do
+        let(:job_state) { 'running' }
```
zaksoup: Is this the right `let` for this context? `job_state` returning a value of 'running' doesn't seem like it implies that the instance is nil.

barthy1:
@zaksoup Thank you for good catch. I've updated the code

mingxiao:
merged into develop branch 
https://github.com/cloudfoundry/bosh/commit/915d33c7ef082523fd71bc81c70de57f1a7b065c

PR #1372, opened 2016-07-28
No longer create `pre_packaging` file when running `generate package`

monkeyherder:
`pre_packaging` file functionality is no longer a suggested workflow, so running `bosh generate package` should no longer create the `pre_packaging` file

### DIFF
--- a/bosh_cli/lib/cli/commands/package.rb
+++ b/bosh_cli/lib/cli/commands/package.rb
@@ -26,11 +26,6 @@ module Bosh::Cli::Command
         "with a non zero value\nset -e\n"
       end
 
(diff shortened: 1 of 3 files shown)


bosh_cli/lib/cli/commands/package.rb
```diff
@@ -26,11 +26,6 @@ def generate(name)
         "with a non zero value\nset -e\n"
       end
 
-      generate_file(package_dir, "pre_packaging") do
-        "# abort script on any command that exits " +
-        "with a non zero value\nset -e\n"
-      end
-
```
zaksoup: this is not idiomatic go

zaksoup:
lgtm

PR #1375, opened 2016-08-02
Allow to use root partition as ephemeral disk on Azure

AbelHu:


zaksoup:
hey @AbelHu while we know you're an azure CPI member, our automated systems won't know it unless you make your membership of the relevant microsoft github orgs public.

Can you also please add a test for this change in `bosh-stemcell/spec/stemcells/azure_spec.rb`. You can find similar tests of the `agent.json` in `bosh-stemcell/spec/stemcells/centos_7_spec.rb`.

AbelHu:
@zaksoup I have changed my profile to set my email address to public. I will add the verification for '**DevicePathResolutionType**' and '**CreatePartitionIfNoEphemeralDisk**' in the case '**installed by bosh_azure_agent_settings**' in bosh-stemcell/spec/stemcells/ubuntu_trusty_spec.rb.

monkeyherder:
Great, thanks for making your address public. It should help @cfdreddbot realize that you've signed a CLA. Did you also make your membership in the appropriate Microsoft Github Orgs public?

Also, I don't think that the spec change cango into `bosh-stemcell/spec/stemcells/ubuntu_trusty_spec.rb`, as this change will only happen for Azure stemcells. So, if the additional test was there, we would end up failing the tests for stemcells on other infrastructures. It would probably be better to have the tests in `bosh-stemcell/spec/stemcells/azure_spec.rb`, which is infrastructure specific, and will only run when we're building Azure stemcells.

But if you can get the tests moved over to `bosh-stemcell/spec/stemcells/azure_spec.rb`, we should be able to pull this change in.

Thanks!

AbelHu:
@monkeyherder I have made my membership in Microsoft Azure public. I also updated the code as you said. Move the test case to bosh-stemcell/spec/stemcells/azure_spec.rb.

dsboulder:
Can we get this merged soon?

dsboulder:
Thanks!

PR #1376, opened 2016-08-03
Bump s3cli to 0.0.45

ljfranklin:
- Includes fix for Region being silently overridden if it didn't
  match the provided AWS Hostname

[#127485941](https://www.pivotaltracker.com/story/show/127485941)


ljfranklin:
As part of merging this in, someone will need to do `bosh add blob` on https://s3.amazonaws.com/s3cli-artifacts/s3cli-0.0.45-linux-amd64 and remove the old blob from the blobs.yml.

PR #1379, opened 2016-08-05
HM creates alerts for instances with missing VMs

beyhan:
such that the resurrector is triggered

- HM uses instances endpoint to get information about
   instances and if they expect a VM
- add new analyze_instances step in addition to analyze_agents
- produce alert for an instance that expects a VM but doesn't have one
- new interval for periodic analyze_instances
- IT added which covers the case where resurrection first fails and then succeeds
- this fixed issue https://github.com/cloudfoundry/bosh/issues/1223
   which prevented resurrection of VMs in this case

[#114101317](https://www.pivotaltracker.com/story/show/114101317)


zaksoup:
Hi @beyhan, Thanks for this PR!

We've definitely recognized the need for this functionality and are really glad to see you contribute.

After reviewing the code we would like to suggest that you refactor this pull request to avoid such heavy use of hashes. Instead we'd like you to fully implement model objects for Deployment, Agent, and Instance. Right now the relationships between these objects are unclear and difficult to discover. Hashes like `agent_id_to_agent` obfuscate the fact that the Agent as a "concept" has one ID and is generally referenced by the director via that ID. In other places, hashes like `deployment_name_to_agent_ids` obfuscate the relationship between Deployments and Instances. Conceptually a Deployment has many Instances, each which have one Agent. Every one of these concepts is backed by an ID (usually a UUID) in the Director/HM databases, but that's not always how we'd like to see them manipulated in code.

Instead, if we have a full ruby object for Deployment, which had many Instances as data members, we can reason about the code and logic with more intuition as people who collectively understand the general concepts of bosh.

Overall, this PR solves the problem well, and we're excited to pull it in. Many places in the BOSH codebase are riddled with what we jokingly refer to as "hash oriented programming". Moving forward, we, as a team, would like to focus on avoiding this pattern when writing new code and refactor old code when we come across it.

Thanks again,
Zak and @monkeyherder!

beyhan:
@zaksoup and @monkeyherder: Thanks for the feedback! We actually saw the need for the proposed refactoring but didn't want to make the change big. It makes absolutely sense to make the proposed refactoring. We are  working on it and will update the pull request.  

Thanks,
Beyhan

zaksoup:
@beyhan thanks! We definitely appreciate trying to keep PRs in readable chunks. We're looking forward to seeing your next few commits 😄

beyhan:
@zaksoup and @monkeyherder: We worked on the refactoring and introduced a new `Deployment` class. 

We didn't rebase this pull-request to the current state of `develop` branch because we wanted to avoid force push and it can still be merged into `develop`. But we did the rebase locally and ran the BOSH ITs which were all green.

bosh-monitor/lib/bosh/monitor/deployment.rb
```diff
@@ -0,0 +1,116 @@
+module Bosh::Monitor
+  class Deployment
+
+    attr_reader :name
+    attr_reader :agent_id_to_agent
+
+    def initialize(instance_data)
```
voelzmo: `instance_data` should probably be `deployment_data`

bosh-monitor/lib/bosh/monitor/deployment.rb
```diff
@@ -0,0 +1,116 @@
...
+        agent = Agent.new(agent_id)
+        @agent_id_to_agent[agent_id] = agent
+      end
+
+      agent.deployment = name
+      agent.job = instance.job
+      agent.index = instance.index
+      agent.cid = instance.cid
+      agent.instance_id = instance.id
+
```
voelzmo: Couldn't all of this `agent.` business go into `Agent.new(name, instance)`?

bosh-monitor/lib/bosh/monitor/instance.rb
```diff
@@ -0,0 +1,56 @@
+module Bosh::Monitor
+  class Instance
+
+    attr_reader :id
```
voelzmo: why is `id` special here?

bosh-monitor/lib/bosh/monitor/instance.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      Instance.new(instance_data)
+    end
+
+    def name
+      if @job
+        name = "#{@deployment}: #{@job}(#{@id}) ["
+        name = name + "agent_id=#{@agent_id}, " if @agent_id
+        name = name + "index=#{@index}, " if @index
+        name + "cid=#{@cid}]"
```
voelzmo: maybe use the `attr_reader` instead of instance variables here?

bosh-monitor/lib/bosh/monitor/instance.rb
```diff
@@ -33,24 +29,37 @@ def self.create(instance_data)
...
     def name
+
       if @job
-        name = "#{@deployment}: #{@job}(#{@id}) ["
-        name = name + "agent_id=#{@agent_id}, " if @agent_id
-        name = name + "index=#{@index}, " if @index
-        name + "cid=#{@cid}]"
+        identifier = "#{@job}(#{@id})"
+        attributes = create_optional_attributes([:agent_id, :index])
+        attributes += create_mandatory_attributes([:cid])
```
voelzmo: What are you using 'identifier' for in the `if` branch?

beyhan:
Hi @zaksoup and @monkeyherder. We are done with the pull-request. Will be great if you can look into it.

Thanks
Beyhan

bosh-monitor/lib/bosh/monitor/director.rb
```diff
@@ -19,8 +19,8 @@ def get_deployments
       parse_json(body, Array)
     end
 
-    def get_deployment_vms(name)
-      http = perform_request(:get, "/deployments/#{name}/vms")
+    def get_deployment_instances(name)
+      http = perform_request(:get, "/deployments/#{name}/instances")
```
mingxiao: endpoint update?

bosh-monitor/lib/bosh/monitor/instance_manager.rb
```diff
@@ -0,0 +1,359 @@
...
+      @rogue_agents.delete(agent_id)
+      @deployment_name_to_deployments.values.each { |deployment| deployment.remove_agent(agent_id) }
+    end
+
+    def get_instances_for_deployment(deployment_name)
+      @deployment_name_to_deployments[deployment_name].instances
+    end
+
+    def analyze_agents
+      @logger.info "Analyzing agents..."
```
mingxiao: We'll clean this up when we merge but we prefer to use parenthesis whenever possible for clarity and consistency. Some exceptions might be in RSpec's idiomatic syntax (`expect(subject).to eq(something)`) but in this case we'd prefer explicit parents.

PR #1383, opened 2016-08-10
Update progressbar to latest.

jmcarp:
Incorporate several fixes to progressbar since version 0.9, including
a fix for terminals that are too narrow or resized: see
https://github.com/jfelchner/ruby-progressbar/pull/42. This should
resolve the issue reported in
https://cloudfoundry.slack.com/archives/bosh/p1458233389000292, which we
have encountered as well.


dpb587-pivotal:
Closing - ruby cli is frozen, so I'm closing. The new beta [gobosh](https://github.com/cloudfoundry/bosh-cli) does a better job at rendering progress bars and I haven't seen any issues with it. Feel free to reopen something in that repo if you start seeing similar issues.

Issue #1387, opened 2016-08-10
Quote all proxy environment variables to account for spaces


ljfranklin:
We recently fixed a bug in the CPIs where the proxy environments variables (`http_proxy`, `https_proxy`, and `no_proxy`) were not quoted in our bash scripts. This caused unexpected failures when a user entered spaces in their `no_proxy` list. After a little digging, looks like we copied those buggy bash lines from bosh-core!

Here's a quick list of places that look to be missing quotes (might be more):
- [director_ctl](https://github.com/cloudfoundry/bosh/blob/afe7a27e51430f275a7bb1489efc8c5ffca11a9b/release/jobs/director/templates/director_ctl.erb#L32-L43)
- [health_monitor_ctl](https://github.com/cloudfoundry/bosh/blob/afe7a27e51430f275a7bb1489efc8c5ffca11a9b/release/jobs/health_monitor/templates/health_monitor_ctl.erb#L14-L25)
- [worker_ctl](https://github.com/cloudfoundry/bosh/blob/afe7a27e51430f275a7bb1489efc8c5ffca11a9b/release/jobs/director/templates/worker_ctl.erb#L35-L46)
- [scheduler_ctl](https://github.com/cloudfoundry/bosh/blob/afe7a27e51430f275a7bb1489efc8c5ffca11a9b/release/jobs/director/templates/scheduler_ctl.erb#L19-L30)

CPI fix is [here](https://github.com/cloudfoundry-incubator/bosh-vsphere-cpi-release/commit/e07b5ebe433587fe79c1850dcb58fc8712a3bd84) for reference.

To see the bug in action:
1. Deploy a director with `jobs.bosh.properties.env.http_proxy` + `https_proxy` set to a fake endpoint (e.g. `127.0.0.1:24`), `no_proxy` set to `127.0.0.1, DIRECTOR_IP` (note the space between items), and `blobstore.address` set to `DIRECTOR_IP`
2. Attempt to upload a release and verify that you receive an error similar to:

```
Connection refused - connect(2) for "127.0.0.1" port 24 (127.0.0.1:24)
```

ljfranklin:
Story in here: https://www.pivotaltracker.com/story/show/128111241

cppforlife:
Merged.

PR #1388, opened 2016-08-10
Quote all proxy environment variables to account for spaces

ljfranklin:
- Prior to this fix, if a user entered spaces under `no_proxy`, only the first address would
  be set due to missing quotes around bash values

[#128111241](https://www.pivotaltracker.com/story/show/128111241)

Signed-off-by: Lyle Franklin <lfranklin@pivotal.io>


ljfranklin:
After this is merged, you can close out https://github.com/cloudfoundry/bosh/issues/1387

PR #1390, opened 2016-08-11
stemcell_builder: force-remove SSH host keys

evandbrown:
If no host keys exist in /etc/ssh, the `rm` command  will return 1 and the
remaining commands will not be executed in an exit-on-error shell. This will
result in no host keys being configured, thus disabling SSH.


cunnie:
Thanks @mingxiao — that will really help my GCE deployments.

PR #1402, opened 2016-08-17
Allow bosh_cli to use IAM profiles

combor:
### What
Currently it is not possible to use IAM roles for `bosh create release --final` command to allow it to upload blobs using temporary security credentials. 

If no AWS keys are specified, blobstore type is S3 and `credentials_source` is set to
`env_or_profile` bosh cli should try to use IAM profile.

### How to test
Use this `private.yml`:

```
---
blobstore:
  s3:
    credentials_source: env_or_profile
```

And execute `bosh create release --final` from a VM that has IAM role assigned and check if blobs are uploaded to your S3 bucket.


cppforlife:
Updated new bosh CLI (https://github.com/cloudfoundry/bosh-cli/issues/24) to latest s3cli which supports everything that director/bosh-agent supports when talking to S3 (https://github.com/pivotal-golang/s3cli#usage).

Issue #936, opened 2015-08-20
Custom Tags in manifest


squeedee:
The AWS console can be quite painful with multiple depoyments. I can use the [Resource Groups](https://docs.aws.amazon.com/awsconsolehelpdocs/latest/gsg/managing-resource-groups.html) to seperate deployments which helps a lot, however, the tagging mechanism for resource groups is 'and' not 'or'. This means that if I use cloudformation or manually deploy non-bosh, but related, resources on AWS, I have to tag them with 'deployment' as well, which can be a little misleading. 

The easiest way to manage complex deployments on a single AWS account is to use tags, but without custom tag support, I am limited in the ways I can take advantage of what AWS provides.

cppforlife:
Would providing a CPI level configuration for tags work for you? e.g. all resources created by the CPI are tagged with some set of same tags.

dpb587-pivotal:
Someone else brought this up in [slack](https://cloudfoundry.slack.com/archives/bosh/p1454025113000259) a few weeks ago.

dpb587-pivotal:
Director [v260+](https://github.com/cloudfoundry/bosh/releases/tag/v260) now supports configuring custom tags via [runtime config](http://bosh.io/docs/runtime-config.html#tags) and [deployment manifests](http://bosh.io/docs/manifest-v2.html#tags).

PR #1405, opened 2016-08-21
Add arbitrary metadata tags to deploy manifest.

jmcarp:
This is a first pass at the interface described by @cppforlife in https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/issues/33#issuecomment-223487970. I'm not very familiar with the bosh internals, so let me know if there's a better way to do this, and I'll be happy to revise.

[Resolves https://github.com/cloudfoundry/bosh/issues/936]
[Resolves https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/issues/33]

### DIFF
--- a/bosh-director/lib/bosh/director/vm_metadata_updater.rb
+++ b/bosh-director/lib/bosh/director/vm_metadata_updater.rb
@@ -13,6 +13,12 @@ module Bosh::Director
     def update(instance, metadata)
       if @cloud.respond_to?(:set_vm_metadata)
         metadata = metadata.merge(@director_metadata)
+
+        manifest = Manifest.load_from_model(instance.deployment)
+        if manifest.raw_manifest_hash.is_a?(Hash)
+          metadata = metadata.merge(manifest.to_hash.fetch('tags', {}))
+        end
+
         metadata[:deployment] = instance.deployment.name
 
         metadata[:id] = instance.uuid
--- a/bosh-director/spec/unit/vm_metadata_updater_spec.rb
+++ b/bosh-director/spec/unit/vm_metadata_updater_spec.rb
@@ -1,6 +1,7 @@
 require 'spec_helper'
 require 'logger'
 require 'timecop'
+require 'yaml'
 
 describe Bosh::Director::VmMetadataUpdater do
   describe '.build' do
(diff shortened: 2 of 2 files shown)


cppforlife:
@jmcarp any reason why we need to save tags into db? we already save entire manifest so might as well read it off from there? wdyt?

jmcarp:
Makes sense to me--do you mean something like:

```ruby
manifest = YAML.load(instance.deployment.manifest || '{}')
metadata = metadata.merge(manifest.fetch('tags', {}))
```

or

```ruby
manifest = Manifest.load_from_model(instance.deployment)
metadata = metadata.merge(manifest.to_hash.fetch('tags', {}))
```

If we read the tags from the manifest in `VmMetadataUpdater`, then can we also drop the changed to `DeploymentSpecParser`? I added those based on discussion in https://github.com/cloudfoundry-incubator/bosh-aws-cpi-release/issues/33#issuecomment-223487970, and my impression was that the purpose of parsing tags in `DeploymentSpecParser` would be to add them as a new field to the `Deployment` model.

jmcarp:
@cppforlife: does the snippet above make sense? If so, I'll revise the patch so that tags aren't stored in the database or parsed in `DeploymentSpecParser`.

cppforlife:
@jmcarp looks about right. have to guard against empty string or null i think.

jmcarp:
@cppforlife: OK, updated to read tags from the manifest instead of saving to the db.

jmcarp:
Ping @cppforlife. Is this closer to what you had in mind?

cppforlife:
@jmcarp we are pulling this in and rearranging some tests/implementations.

dpb587-pivotal:
Related story: https://www.pivotaltracker.com/story/show/129890801

jmcarp:
Interesting, looks like you all switched back to the implementation I was working with originally, when tags were parsed and stored in their own model: https://github.com/cloudfoundry/bosh/commit/06eb96835d3928cdc13bedece1c7a9902d3ff604.

jmcarp:
For reference, old work at https://github.com/jmcarp/bosh/commit/d0987072ed9544f68234c67096e549602f17dc28.

cppforlife:
@jmcarp misunderstanding on their part, we are doing what you did in memory. code coming shortly with additional tests.

cppforlife:
I think this was delivered in v258 release.

cppforlife:
Thanks for getting the ball rolling on this.

PR #1406, opened 2016-08-22
Support parallel execution of integration tests

friegger:
Hi,

we have added a new task `fly:integration_parallel` that supports parallel execution of fly exec tasks similar to what happens in the concourse pipeline. It runs all one-offs on the `fly-integration` worker. Console output is interlaced, but to have a look what is actually happening one can look on concourse. Using this task execution time is reduced to ca. 30 min.

Signed-off-by: Mauro Morales <mamorales@suse.com>

### DIFF
--- a/bosh-dev/lib/bosh/dev/tasks/fly.rake
+++ b/bosh-dev/lib/bosh/dev/tasks/fly.rake
@@ -5,14 +5,24 @@ namespace :fly do
   # bundle exec rake fly:unit
   desc 'Fly unit specs'
   task :unit do
-    execute('test-unit', '-p')
+    execute('test-unit', '-p', {
+        DB: (ENV['DB'] || 'postgresql')
+    })
   end
 
   # bundle exec rake fly:integration
   desc 'Fly integration specs'
   task :integration do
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
Great idea! But... I don't think it's working as expected. We tried this out locally and it seems like all three builds are running the last group of tests. Does this happen in your environment, or do you have suggestions on what other changes might be required?

    build1: bundle exec parallel_test 'spec/integration' -n 24 --only-group 3,6,9,12,15,18,21,24 --group-by filesize --type rspec -o '--format documentation'
    build2: bundle exec parallel_test 'spec/integration' -n 24 --only-group 3,6,9,12,15,18,21,24 --group-by filesize --type rspec -o '--format documentation'
    build3: bundle exec parallel_test 'spec/integration' -n 24 --only-group 3,6,9,12,15,18,21,24 --group-by filesize --type rspec -o '--format documentation'

Kiemes:
Good catch! We have made the env vars an explicit argument. The `unit` task was missing the DB env var, we fixed that, too.

PR #1407, opened 2016-08-23
show explicit -t <target> in 'bosh task <num>' error command

drnic:
Previously, error messages from a failed bosh command said to run `bosh task <num> --debug`. Now it includes the explicit target: `bosh -t <target> task <num> --debug`


dpb587-pivotal:
Closing - ruby CLI is feature frozen in favor of gocli, so I'm going to close this. Since `bosh` commands require the environment/deployment options, this shouldn't be an issue.

Issue #1360, opened 2016-07-20
Job release upgrade will trigger recreate vm after bosh upgrade


gu-bin:
After upgrading the bosh from version 1.3153.0 to 1.3262.3.0, we try to upgrade cf release version for the deployment. This triggers recreate (delete and create) vm which is not expected.

In the bosh task debug log, it detected networks changed so recreate vm was triggered.

```
D, [2016-07-20 03:53:48 #26642] [canary_update(ipsec-orchestrator/0 (38b37b1d-a028-4aad-9e54-fc09e9837a92))] DEBUG -- DirectorJobRunner: networks_changed? network settings changed FROM: {"default"=>{"type"=>"dynamic", "cloud_properties"=>{"security_groups"=>["default", "cf"]}, "dns"=>["159.8.180.206", "10.0.80.11", "10.0.80.12"], "default"=>["dns", "gateway"], "dns_record_name"=>"0.ipsec-orchestrator.default.gubin-preyf.microbosh", "ip"=>"169.50.71.93", "netmask"=>"255.255.255.224", "gateway"=>"169.50.71.65"}} TO: {"default"=>{"type"=>"dynamic", "cloud_properties"=>{"security_groups"=>["default", "cf"]}, "dns"=>["159.8.180.206", "10.0.80.11", "10.0.80.12"], "default"=>["dns", "gateway"], "ip"=>"169.50.71.93", "netmask"=>"255.255.255.224", "gateway"=>"169.50.71.65"}} on instance ipsec-orchestrator/0 (38b37b1d-a028-4aad-9e54-fc09e9837a92)
D, [2016-07-20 03:53:48 #26642] [task:207] DEBUG -- DirectorJobRunner: Waiting for tasks to complete
D, [2016-07-20 03:53:48 #26642] [canary_update(ipsec-orchestrator/0 (38b37b1d-a028-4aad-9e54-fc09e9837a92))] DEBUG -- DirectorJobRunner: Networks have changed. Recreating VM
```

So `"dns_record_name"=>"0.ipsec-orchestrator.default.gubin-preyf.microbosh"` is removed in the new bosh.

Is there any way to avoid recreate vm in this case? Thanks.

cppforlife:
@gu-bin yeah we removed 'dns_record_name' since it's not supposed to be sent to the agent. not sure how to avoid it. i think we would be open to the pr to ignore that key in networks_changed? method.

maximilien:
[...]

PR #1411, opened 2016-08-24
Ignore dns_record_name when validating changes of network_settings

jianqiu:
This PR aims to fix the issue [1360] (https://github.com/cloudfoundry/bosh/issues/1360)

### DIFF
--- a/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
@@ -1,3 +1,5 @@
+require 'common/deep_copy'
+
 module Bosh
   module Director
     module DeploymentPlan
@@ -274,10 +276,13 @@ module Bosh
 
         def remove_dns_record_name_from_network_settings(network_settings)
           return network_settings if network_settings.nil?
-          network_settings.each do |name, network_setting|
+
+          modified_network_settings = Bosh::Common::DeepCopy.copy(network_settings)
+
+          modified_network_settings.each do |name, network_setting|
             network_setting.delete_if{|key, value| key == "dns_record_name"}
           end
-          network_settings
+          modified_network_settings
         end
 
         def env_changed?
--- a/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
@@ -269,7 +269,14 @@ module Bosh
 
         def network_settings_changed?(old_network_settings, new_network_settings)
           return false if old_network_settings == {}
-          old_network_settings != new_network_settings
+          remove_dns_record_name_from_network_settings(old_network_settings) != new_network_settings
+        end
+
+        def remove_dns_record_name_from_network_settings(network_settings)
+          network_settings.each do |name, network_setting|
+            network_setting.delete_if{|key, value| key == "dns_record_name"}
+          end
+          network_settings
         end
 
         def env_changed?
--- a/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
@@ -273,6 +273,7 @@ module Bosh
         end
 
         def remove_dns_record_name_from_network_settings(network_settings)
+          return network_settings if network_settings == {}
           network_settings.each do |name, network_setting|
             network_setting.delete_if{|key, value| key == "dns_record_name"}
           end
--- a/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
+++ b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
@@ -273,7 +273,7 @@ module Bosh
         end
 
         def remove_dns_record_name_from_network_settings(network_settings)
-          return network_settings if network_settings == {}
+          return network_settings if network_settings.nil?
           network_settings.each do |name, network_setting|
             network_setting.delete_if{|key, value| key == "dns_record_name"}
           end
--- a/bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
+++ b/bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
@@ -119,25 +119,31 @@ module Bosh::Director::DeploymentPlan
           instance_plan.networks_changed?
         end
 
-        it 'should ignore dns_record_name change of network settings' do
-          old_network_settings = {
-              'existing-network' =>{
-                  'type' => 'dynamic',
-                  'cloud_properties' =>{},
-                  'dns_record_name' => '0.job-1.my-network.deployment.bosh',
-                  'dns' => '10.0.0.1',
-              },
-          }
+        context 'when dns_record_name exists in network_settings' do
+          let(:network_settings) do
+            {
+                'existing-network' =>{
+                    'type' => 'dynamic',
+                    'cloud_properties' =>{},
+                    'dns_record_name' => '0.job-1.my-network.deployment.bosh',
+                    'dns' => '10.0.0.1',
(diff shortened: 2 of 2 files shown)


jianqiu:
@cppforlife 

Hi Dmitriy, 

Could you please help us look at the PR? We are looking forward for your comments, thanks!

cc: @maximilien @mattcui

bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb
```diff
@@ -269,7 +269,15 @@ def persist_current_spec
 
         def network_settings_changed?(old_network_settings, new_network_settings)
           return false if old_network_settings == {}
-          old_network_settings != new_network_settings
+          remove_dns_record_name_from_network_settings(old_network_settings) != new_network_settings
+        end
+
+        def remove_dns_record_name_from_network_settings(network_settings)
```
cppforlife: let's use Bosh::Common::DeepCopy to make sure old_network_settings arent modified.

jianqiu: @cppforlife ,
Thanks for your comments, I had already updated my PR, could you please review it again? Thanks!

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
```diff
@@ -119,6 +119,36 @@ module Bosh::Director::DeploymentPlan
...
+                'existing-network' =>{
+                    'type' => 'dynamic',
+                    'cloud_properties' =>{},
+                    'dns' => '10.0.0.1',
+                }
+            }
+
+            allow(logger).to receive(:debug)
+            expect(logger).to_not receive(:debug).with(
+                "networks_changed? network settings changed FROM: #{network_settings} TO: #{new_network_settings} on instance #{instance_plan.existing_instance}"
```
mingxiao: This tests passes even if network_settings_changed? returns true. The content of the 'new_network_settings' variable does not accurately reflect what is being compared against 'network_settings'.

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
```diff
@@ -119,6 +119,36 @@ module Bosh::Director::DeploymentPlan
...
+            )
+
+            instance_plan.networks_changed?
```
mingxiao: Is this expected to be `true` or `false`?

jianqiu:
@mingxiao @cppforlife @maximilien 
I saw you merged this PR and then reverted it, and also changed my testcase? How could I handle this PR? Thanks!

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
jianqiu: The result is definitely  true, since the PR aims to ignore the difference "dns_record_name" when new_network_settings does not have this property.

mingxiao:
@jianqiu The issues are already noted. 
- The test on debug output is a false positive
- Please make an assertion on the result of `networks_changed`.

We also saw that `networks_changed?` is returning true b/c the test case is configured so that `desired_network_plans.any?` is true. https://github.com/jianqiu/bosh/blob/533d03117c1a48c1bdad56bd197b0c7bde1fb15b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb#L105
Is that the intended behavior of that test?

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
jianqiu: @mingxiao 

The content of `new_network_settings` does not contain key `dns_record_name` while `old_network_settings`  had it

```
    let(:network_settings) do
         {
                   'existing-network' =>{
                   'type' => 'dynamic',
                   'cloud_properties' =>{},
                   'dns_record_name' => '0.job-1.my-network.deployment.bosh',
                   'dns' => '10.0.0.1',
               }
             }
```
In the issue [1360](https://github.com/cloudfoundry/bosh/issues/1360), the difference is caused by different version of bosh-agent,  it will trigger unnecessary recreate_vm.

jianqiu:
@mingxiao, 
https://github.com/jianqiu/bosh/blob/533d03117c1a48c1bdad56bd197b0c7bde1fb15b/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb#L105 
The above line is not included in my PR. 
Could you please take a look at the issue [1360](https://github.com/cloudfoundry/bosh/issues/1360) which is the context of this PR.

- The test on debug output is a false positive : what it the meaning? 
- Please make an assertion on the result of networks_changed.
I just took the following test case as example:
```
        it 'should log the changes' do
          new_network_settings = {
            'existing-network' =>{
              'type' => 'dynamic',
              'cloud_properties' =>{},
              'dns' => '10.0.0.1',
            },
            'a' =>{
              'ip' => '192.168.1.3',
              'netmask' => '255.255.255.0',
              'cloud_properties' =>{},
              'default' => ['dns', 'gateway'],
              'dns' =>['192.168.1.1', '192.168.1.2'],
              'gateway' => '192.168.1.1',
            }
          }

          allow(logger).to receive(:debug)
          expect(logger).to receive(:debug).with(
              "networks_changed? network settings changed FROM: #{network_settings} TO: #{new_network_settings} on instance #{instance_plan.existing_instance}"
            )

          instance_plan.networks_changed?
        end
```

mingxiao:
@jianqiu I think I have enough context now to make the changes.

When I say false positive, I mean it is passing when it should not.
The issue with your current test is that `new_network_settings`, https://github.com/cloudfoundry/bosh/pull/1411/files#diff-c0ae009b7ea6019dc2b271e8c7ef98d0R135,  does not match what is being processed in `networks_changed?`

This is your assert:
`networks_changed? network settings changed FROM: {"existing-network"=>{"type"=>"dynamic", "cloud_properties"=>{}, "dns_record_name"=>"0.job-1.my-network.deployment.bosh", "dns"=>"10.0.0.1"}} TO: {"existing-network"=>{"type"=>"dynamic", "cloud_properties"=>{}, "dns"=>"10.0.0.1"}} on instance job-1/fake-uuid-1 (1)`

And this is what being logged:
`networks_changed? network settings changed FROM: {"existing-network"=>{"type"=>"dynamic", "cloud_properties"=>{}, "dns_record_name"=>"0.job-1.my-network.deployment.bosh", "dns"=>"10.0.0.1"}} TO: {"existing-network"=>{"type"=>"dynamic", "cloud_properties"=>{}, "dns"=>"10.0.0.1"}, "a"=>{"ip"=>"192.168.1.3", "netmask"=>"255.255.255.0", "cloud_properties"=>{}, "default"=>["dns", "gateway"], "dns"=>["192.168.1.1", "192.168.1.2"], "gateway"=>"192.168.1.1"}} on instance job-1/fake-uuid-1 (1)
`

Notice that your test is missing the `new_network_settings['a']` key

bosh-director/spec/unit/deployment_plan/instance_plan_spec.rb
mingxiao: consulted with @dpb587 , and we think that the result should be `false`. Since the network did not change and we are now ignoring `dns_record_name`, `networks_changed?` should return false

dpb587-pivotal: Since it was already merged at one point and our existing tests were a bit
weird to begin with, maybe lets just re-merge and we clean up whatever we
need to. Seems easier and more helpful than back-and-forth on these details.

mingxiao:
Fix and pushed onto `develop` branch, with fix https://github.com/cloudfoundry/bosh/commit/791a8211f8d4d9f6d6b0dfc227381f20fafc83e1

PR #1419, opened 2016-08-29
Local DNS should have canonicalized names

barthy1:
[#127506377](https://www.pivotaltracker.com/story/show/127506377)


dpb587-pivotal:
Merged - thanks!

PR #1420, opened 2016-08-29
bosh cleanup also delete orphaned disks

mrdavidlaing:
@cppforlife says that `bosh cleanup`

> it includes orphaned disks as part of cleaning up.

This PR updates the command line help to make that clear

### DIFF
--- a/bosh_cli/lib/cli/commands/maintenance.rb
+++ b/bosh_cli/lib/cli/commands/maintenance.rb
@@ -6,8 +6,8 @@ module Bosh::Cli::Command
 
     # bosh cleanup
     usage 'cleanup'
-    desc 'Cleanup releases and stemcells'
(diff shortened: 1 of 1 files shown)


dpb587-pivotal:
FWIW, I see the command description in the new cli [says](https://github.com/cloudfoundry/bosh-init/blob/59df0ac3b75eef3f938bcc9cc2acfc0e08531ff4/cmd/opts.go#L54)...

    Clean up releases, stemcells, disks, etc.

Although for `--all` it doesn't mention disks and still [has](https://github.com/cloudfoundry/bosh-init/blob/59df0ac3b75eef3f938bcc9cc2acfc0e08531ff4/cmd/opts.go#L227)...

    Remove all unused releases, stemcells, etc.; otherwise most recent resources will be kept

We should probably keep in mind to update that, too.

PR #1424, opened 2016-08-31
Look up instances IPs in director DB

Kiemes:
Previously they were retrieved via the bosh agent

[#115972549](https://www.pivotaltracker.com/story/show/115972549)

### DIFF
--- a/bosh-director/lib/bosh/director/jobs/vm_state.rb
+++ b/bosh-director/lib/bosh/director/jobs/vm_state.rb
@@ -46,7 +46,7 @@ module Bosh::Director
         {
           :vm_cid => instance.vm_cid,
           :disk_cid => instance.managed_persistent_disk_cid,
-          :ips => ips,
+          :ips => instance.ip_addresses.map {|ip| NetAddr::CIDR.create(ip.address).ip },
           :dns => dns_records,
           :agent_id => instance.agent_id,
           :job_name => instance.job,
--- a/bosh-director/lib/bosh/director/jobs/vm_state.rb
+++ b/bosh-director/lib/bosh/director/jobs/vm_state.rb
@@ -34,7 +34,7 @@ module Bosh::Director
       def process_instance(instance)
         dns_records = []
 
-        job_state, job_vitals, processes, ips = vm_details(instance)
+        job_state, job_vitals, processes, _ = vm_details(instance)
 
         if dns_manager.dns_enabled?
           dns_records = dns_manager.find_dns_record_names_by_instance(instance)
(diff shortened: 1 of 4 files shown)


cppforlife:
@Kiemes let's merge (uniq) the agent reported ips with whatever database has. database will only contain ips if deployment is switched to global networking.
cc @voelzmo

Kiemes:
@cppforlife thanks for the hint. We actually tried both ways, doing a deployment which uses cloud-config and one without cloud-config. In both cases we saw the IP addresses in the `instances` table in `spec_json` column. This is even the case regardless of manual or dynamic networks.

```
{
	"deployment": "dummy-legacy",
	"networks": {
		"dummy-dynamic": {
			"type": "dynamic",
			"cloud_properties": {...},
			"default": ["dns", "gateway"],
			"ip": "10.0.2.102",
			"netmask": "255.255.255.0",
			"gateway": "10.0.2.1"
		},
		"public": {
			"type": "vip",
			"ip": "192.168.147.159",
			"cloud_properties": {}
		}
	},
	"links": {},
	"address": null,
	..
}
```

```
{
	"deployment": "dummy-with-cloud-config",
	"networks": {
		"dummy-manual": {
			"ip": "10.0.1.12",
			"netmask": "255.255.255.0",
			"cloud_properties": {...},
			"default": ["dns", "gateway"],
			"dns": ["172.16.0.1"],
			"gateway": "10.0.1.1"
		},
		"public": {
			"type": "vip",
			"ip": "192.168.147.160",
			"cloud_properties": {}
		}
	},
	"links": {},
	"address": "10.0.1.12",
}
```

Our implementation takes every ip_address from the networks array. So even if the `address` is set to null in legacy deployments, we would still include every IP. Is this correct or are we missing something?

BTW the [director schema](https://github.com/cloudfoundry/bosh/blob/master/docs/director_schema/bosh-director-schema.png) in bosh docs is outdated.

Kiemes:
@cppforlife you were right, the association 'ip_addresses' is empty in some cases. Therefor we change the behavior to use 'spec' as a fallback. There we showed already that IP addresses are listed in all cases.
In contrast to using the agent data as fallback, our current proposal will work for unresponsive agents as well.

Units and integration tests are green. Feel free to merge. Unless you have additional feedback.

tylerschultz:
LGTM, @cppforlife would you please make/prioritize a story to merge it in?

tylerschultz:
merged! Thank you.

dpb587-pivotal:
Commits were rebased as https://github.com/cloudfoundry/bosh/commit/6ba9c7004a336d491201d60bf79699d018de03cf, https://github.com/cloudfoundry/bosh/commit/9fe4cd8e76e689b5c2360bef16c7345e73d67780.

voelzmo:
thanks!

PR #1428, opened 2016-09-06
Bump bosh_openstack_cpi and switch to fog-openstack

Kiemes:
- consume fixes in excon 0.49, especially freezing connections
  in the registry
  - fixes https://github.com/cloudfoundry/bosh/issues/1234
- get rid of full fog (we only need fog-aws and fog-openstack)
- some other gems were taken in newer version by
  `bundle update bosh_openstack_cpi`

[#119363951](https://www.pivotaltracker.com/story/show/119363951)


Kiemes:
@cppforlife That is basically a cherry-pick of https://github.com/cloudfoundry/bosh/commit/78cc74c5f72601c08361c6c777dd94f368456283 onto 257.x in order to get it released faster.

PR #1433, opened 2016-09-12
Display job name and id.

knm3000:
[#126983019](https://www.pivotaltracker.com/story/show/126983019)

### DIFF
--- a/bosh-director/lib/bosh/director/jobs/ssh.rb
+++ b/bosh-director/lib/bosh/director/jobs/ssh.rb
@@ -35,11 +35,8 @@ module Bosh::Director
 
             logger.info("ssh #{@command} '#{instance.job}/#{instance.uuid}'")
             result = agent.ssh(@command, @params)
-            if target.ids_provided?
-              result["index"] = result["id"] = instance.uuid
-            else
-              result["index"] = result["id"] = instance.index
-            end
+            result["id"] = instance.uuid
+            result["index"] = instance.index
             result["job"] = instance.job
 
             if Config.default_ssh_options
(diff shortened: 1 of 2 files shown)


bosh-director/lib/bosh/director/jobs/ssh.rb
```diff
@@ -38,8 +38,9 @@ def perform
             if target.ids_provided?
               result["id"] = instance.uuid
             else
-              result["index"] = instance.index
+              result["index"] = result["id"] = instance.index
```
dpb587-pivotal: Should we always send back `id` and `index`? Including both values seems valid for an SSH results response. Then the UI/cli becomes responsible for showing the user whichever is appropriate. Or maybe that would impact the ruby CLI in ways we want to avoid? Dunno; just a thought.

cppforlife:
@knm3000 agree with @dpb587-pivotal. let's just return them both no matter what.

bosh-director/lib/bosh/director/jobs/ssh.rb
```diff
@@ -36,10 +36,11 @@ def perform
             logger.info("ssh #{@command} '#{instance.job}/#{instance.uuid}'")
             result = agent.ssh(@command, @params)
             if target.ids_provided?
-              result["id"] = instance.uuid
+              result["index"] = result["id"] = instance.uuid
```
dpb587-pivotal: Hmm, this would end up changing the data format of `index` which I don't think is intentional (maybe some clients require integers). Rather than having a conditional for `target.ids_provided?`, let's just do...

```
result["id"] = instance.id
result["index"] = instance.index
```

Old CLI will look at `index` and have the same info as normal and new CLI will correctly show the uuid.

PR #1446, opened 2016-09-22
Update system guest package for Google stemcell

evandbrown:
This change uses the latest [`compute-image-packages`] for Google guest VMs.
The new package allows more granular configuration of daemons and improves
compatibility with other services on boot.

[`compute-image-packages`]: https://github.com/GoogleCloudPlatform/compute-image-packages/releases/tag/20160803

### DIFF
--- a/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
+++ b/bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
@@ -142,6 +142,7 @@ module Bosh::Stemcell
 
     def google_stages
       [
+        :rsyslog_config,
         :system_network,
         :system_google_modules,
         :system_google_packages,
--- a/stemcell_builder/stages/rsyslog_config/assets/rsyslog_upstart.conf
+++ b/stemcell_builder/stages/rsyslog_config/assets/rsyslog_upstart.conf
@@ -11,12 +11,6 @@ stop on runlevel [06]
 expect fork
 respawn
 
-pre-start script
-	until mountpoint -q /var/log; do
-	  sleep .1
-	done
-end script
-
 script
     . /etc/default/rsyslog
     exec /usr/sbin/rsyslogd $RSYSLOGD_OPTIONS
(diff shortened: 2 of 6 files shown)


evandbrown:
**An important note:**

The stemcell fails for me when applied to the `develop`

bosh-stemcell/lib/bosh/stemcell/stage_collection.rb
```diff
@@ -142,6 +142,7 @@ def aws_stages
 
     def google_stages
       [
+        :rsyslog_config,
```
evandbrown: This stage must be run for the Google stemcell to provide a correct `/etc/init/rsyslog.conf`

stemcell_builder/stages/rsyslog_config/assets/rsyslog_upstart.conf
```diff
@@ -11,12 +11,6 @@ stop on runlevel [06]
 expect fork
 respawn
 
-pre-start script
```
evandbrown: This declaration is removed as `/var/log` never appears on the Google stemcell. There was discussion that `bosh-agent` may be expected to mount this location to ephemeral disk, which does not exist on GCE.

stemcell_builder/stages/system_google_packages/assets/instance_configs.cfg.template
```diff
@@ -1,3 +1,10 @@
...
-    'set_host_keys': 'false',
-},
+[Daemons]
+accounts_daemon = true
+clock_skew_daemon = true
+ip_forwarding_daemon = true
+[InstanceSetup]
+network_enabled = true
+optimize_local_ssd = true
+set_host_keys = false
```
evandbrown: Prevent the Google daemon from setting SSH host keys, which races with BOSH's own configuration.

dpb587-pivotal:
This PR looks like it was superseded by https://github.com/cloudfoundry/bosh/pull/1454, although I don't see any reference to `:rsyslog_config` there. If this PR is still relevant after those changes, can you rebase/fix for whatever changes are still needed? Otherwise I think this can be closed.

PR #1447, opened 2016-09-23
go binaries built with <1.7 broken on MacOS Sierra

loewenstein:
We have seen the problems with both bosh-init and bosh go-cli.
We locally built bosh-init with go 1.7.1 and deployment was fine.

We did _not_ figure out what might be broken with 1.7.1 though.

### DIFF
--- a/ci/docker/main-ruby-go/install-go.sh
+++ b/ci/docker/main-ruby-go/install-go.sh
@@ -3,8 +3,8 @@
 set -eux
 
 GOPATH=/home/vagrant/go
-GO_ARCHIVE_URL=https://storage.googleapis.com/golang/go1.6.1.linux-amd64.tar.gz
-GO_ARCHIVE_SHA256=6d894da8b4ad3f7f6c295db0d73ccc3646bce630e1c43e662a0120681d47e988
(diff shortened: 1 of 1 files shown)


ljfranklin:
Some food for thought going forward, Concourse recently switched to producing dynamically linked OSX binaries for `fly`. Apparently this fixes some DNS related issues and OSX Keychain integration. I also remember a blog post recommending to always statically compile Golang Linux binaries, but recommended dynamically linking OSX binaries for similar reason. Not sure if the BOSH team would want to do the same long-term.

tylerschultz:
We've updated the boshcli and bosh-init concourse docker images to include 1.7.1. The latest versions of the boshcli should now be built using 1.7.1. bosh-init will require a new version is published, blocked on PM.

dpb587-pivotal:
Published [bosh-init/0.0.97](https://github.com/cloudfoundry/bosh-init/releases/tag/v0.0.97) and [bosh-cli/0.0.67](https://github.com/cloudfoundry/bosh-cli/releases/tag/v0.0.67) with go/1.7.1.

cppforlife:
I think this can be closed now.

PR #1461, opened 2016-09-29
Force director to use proxy settings for remote download

geofffranks:
Previously, the Net::HTTP.start() call in the download helper
for the bosh director was ignoring the http_proxy/https_proxy
environment variables. This remedies that.

### DIFF
--- a/bosh-director/lib/bosh/director/download_helper.rb
+++ b/bosh-director/lib/bosh/director/download_helper.rb
@@ -13,7 +13,7 @@ module Bosh::Director
     def download_remote_file(resource, remote_file, local_file, num_redirects = 0)
       @logger.info("Downloading remote #{resource} from #{remote_file}") if @logger
       uri = URI.parse(remote_file)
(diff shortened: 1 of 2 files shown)


geofffranks:
@dpb587 or @cppforlife is the pivotal tracker integration for this broken? Shouldn't I see a link to a ticket for this PR?

dpb587-pivotal:
Story: https://www.pivotaltracker.com/story/show/131738547

This repo doesn't currently use a pr/issue/tracker integration.

geofffranks:
Thanks!

Some background for testing - BOSH was behaving with the proxy settings for most everything (talking to the CPI for instance). However when doing `bosh upload release` or `bosh upload stemcell` using remote URLs, the content download was not using the proxy (though the upload to the blobstore after content was downloaded is using the proxy).

cppforlife:
moved story to backlog.

dpb587-pivotal:
Merged onto `develop` via a206491fa1becca018288ff02359af7c7a0297c2.

Issue #1462, opened 2016-09-30
S3 Blobstore in read only mode ignores advanced configs


johnsonj:
# Background
When no credentials are provided and no endpoint is set, the `S3BlobstoreClient` will create a `SimpleBlobstoreClient` based off of the bucket name and the default endpoint (`S3BlobstoreClient::ENDPOINT`). This ignores any options that would be used when credentials are set (`host, use_ssl, port`).

# Repro
Given `config/fnal.yaml`:
```yaml
---
final_name: example-app
blobstore:
  provider: s3
  options:
    bucket_name: my-gcp-bucket
    host: storage.googleapis.com
    use_ssl: false
```
## Expected
Blobstore attempts to fetch from: `http://storage.googleapis.com/my-gcp-bucket/`
## Actual
Blobstore attempts to fetch from: `https://s3.amazonaws.com/my-gcp-bucket/`

cppforlife:
Should be covered by https://github.com/cloudfoundry/bosh-cli/issues/24 in new bosh cli.

PR #1463, opened 2016-09-30
S3 Blobstore in read only mode ignores advanced configs

johnsonj:
Defect:
When no credentials are provided and no endpoint is set, the
S3BlobstoreClient will create a SimpleBlobstore based off of the bucket
name and the default endpoint (S3BlobstoreClient::ENDPOINT).
This ignores any options that would be used when credentials are set.

Fix:
If no endpoint is set, use the resolved endpoint from the @aws_options.
If no endpoint is resolved there, default back to S3BlobstoreClient::ENDPOINT

Fixes issue #1462

### DIFF
--- a/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
+++ b/blobstore_client/lib/blobstore_client/s3_blobstore_client.rb
@@ -49,7 +49,7 @@ module Bosh
           end
 
           @options[:bucket] ||= @options[:bucket_name]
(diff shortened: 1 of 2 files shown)


cppforlife:
@johnsonj we are trying to switch over to cloudfoundry/bosh-cli soon which means that configuration covered by https://github.com/pivotal-golang/s3cli will work. im not sure it makes sense for us to pull in this change.

cppforlife:
Closed in favor of fixes made in https://github.com/cloudfoundry/bosh-cli/issues/24.

PR #1467, opened 2016-10-07
Check if instance has vm for ssh job

barthy1:
[#126983269](https://www.pivotaltracker.com/story/show/126983269)

### DIFF
--- a/bosh-director/lib/bosh/director/api/instance_lookup.rb
+++ b/bosh-director/lib/bosh/director/api/instance_lookup.rb
@@ -32,8 +32,8 @@ module Bosh::Director
         instance
       end
 
-      def by_filter(filter)
-        instances = Models::Instance.filter(filter).all
+      def by_filter(filter, exclude = {})
+        instances = Models::Instance.filter(filter).exclude(exclude).all
         if instances.empty?
           raise InstanceNotFound, "No instances matched #{filter.inspect}"
         end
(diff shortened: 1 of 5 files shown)


dpb587-pivotal:
Merge story: https://www.pivotaltracker.com/story/show/131926683

voelzmo:
@dpb587 @barthy1 I'm not sure if checking for `vm_cid == nil` is what you want to do. That means you now got different behavior depending on whether an errand is running when you call `bosh ssh` or not. What about checking the `expects_vm` flag on an instances as suggested in https://github.com/cloudfoundry/bosh-cli/issues/18#issue-180256086?

cppforlife:
definitely want to ssh into errand vms, otherwise it will be hard to debug if you are usin keep alive.

voelzmo:
@cppforlife I'm not saying you shouldn't be able to ssh into errand VMs. I'm saying the *automatic selection* of an instance, if you didn't provide an instance to the command should not be different based on if an errand is running. That sounds weird to me.

cppforlife:
@voelzmo not sure i agree. if i see the vm in bosh instances and run `bosh ssh -c 'blah'` command i would expect it will cover all machines. i dont think it would be obvious why some machines wouldnt be affected.

voelzmo:
@cppforlife I'm not talking about running a command on all VMs, I'm talking about the automatic selection of an instance for interactive ssh in the case that the user didn't provide the instance.
This selection should not be different if by coincidence an errand is running? It probably shouldn't exist at all, if there is more than one instance.

dpb587-pivotal:
imo, `bosh ssh`, without a specific instance, puts you at the mercy of director to connect to any particular instance which happens to be available. I think you should always have to give it an instance if you want interactive; and the only use for not specifying a slug is when the deployment contains one instance. /opinion

PR #1471, opened 2016-10-11
Support for multiple CPIs (mvp)

MatthiasWinzeler:
This PR provides basic support for multiple CPIs in a single BOSH director. 

The high level concept is outlined here: https://github.com/cloudfoundry/bosh-notes/blob/master/multi-cpi.md
@cppforlife supported me with implementation details and reviews of some code parts.

Some notable points:
- The heart of the PR is the newly introduced `CpiConfig` which stores the multiple CPI endpoint configurations. [CloudFactory](https://github.com/swisscom/bosh/blob/multi-cpi-mvp/bosh-director/lib/bosh/director/cloud_factory.rb) then looks up the cpi config for a given AZ and returns an instantiated `Cloud`. To achieve this lookup, the cloud factory uses the latest cpi config and the deployment's cloud config (which contains the AZ -> CPI relation).
- Every cloud interaction thus needs an up-to-date `deployment.cloud_config`. In some cases, i.e. while updating the deployment, `deployment.cloud_config` did not contain the most recent `CloudConfig` but an older one, because this association was only set and persisted _after_ updating the deployment. I had to do some [ugly](https://github.com/cloudfoundry/bosh/commit/6539a5f6f8b1ef49a558e071bbdbb5b78bd4bd57) [hacks](https://github.com/swisscom/bosh/commit/a6f6808bb48f0ac01b47bdcf29d391bff5b50b5c#diff-54a465ad1bd54f155ee42ef5a5a15991) to set `deployment.cloud_config` before every cloud interaction. I guess this needs some improvements.
- I added only very basic integration tests. The branch is built [in BOSH's concourse](https://main.bosh-ci.cf-app.com/teams/main/pipelines/bosh:multi-cpi) and is successful. I never ran BATS tests though. Do we need some more integration tests or BATS tests?
- I was having a hard time completely understanding the code parts of stemcell compilation. I’m not quite sure I changed it the right way, so I’d appreciate if you could look at this more carefully.
- For stemcells, we extended the unique constraint to the new `cpi` attribute. But since postgresql seems to completely ignore the uniqueness constraint if one of the fields is `NULL` (which is the case for the `cpi`attribute when not using cpi config), I added a [second validation](https://github.com/swisscom/bosh/commit/515a7dbcb6c1e95513b06b3ad1802d2ef3ec5acf). I assume this validation could fail if there is a stemcell with name=x, version=y, cpi=z in DB and we add one with name=x, version=y, cpi=nil. Is there a better way to do this? A custom validator? Ignore this case?
- To be able to support the cpi config, each CPI needs to be modified to also be able to read the configuration from the `context` hash. We implemented this small change for the [openstack cpi](https://github.com/swisscom/bosh-openstack-cpi-release/tree/multi-cpi). 
  We tested this in our lab and were successfully able to stretch deployments over multiple openstacks. I will file the PR there as soon as this PR is merged and a new version of the `bosh_cpi` gem is ready for downstream use. (cc @voelzmo)


cppforlife:
> since postgresql seems to completely ignore the uniqueness constraint if one of the fields is NULL

@MatthiasWinzeler should be we use "" instead of NULL then. i would like us to have db enforced constraint is possible.

ljfranklin:
@cppforlife @MatthiasWinzeler question about the context hash handling in the CPIs, e.g. `cloud_properties['openstack'].merge!(context['openstack'])`. I like that the global CPI configuration is now dynamic, but we seem to have this weird mix of dynamic values vs static values in the [cpi.json file](https://github.com/swisscom/bosh-openstack-cpi-release/blob/master/jobs/openstack_cpi/templates/cpi.json.erb).

Since the expectation is for all `cloud.properties.openstack` values to be pulled at runtime from the `context`, I'd like to push back against the `merge` call and go back to overriding: `cloud_properties['openstack'] = context['openstack']`. Having to add properties to spec files and ERB templates seems like a waste when its only use is to provide default values. **I think it would be easier as a developer to have default values in the code instead and make everything dynamic.** I'd imagine that users don't look at the CPI spec file for defaults, but would use the bosh.io CPI docs instead.

Big thanks for the PR btw 👍

bosh-director/lib/bosh/director/cpi_config/cpi.rb
```diff
@@ -0,0 +1,36 @@
...
+        @name = name
+        @type = type
+        @exec_path = exec_path
+        @properties = properties
+        validate
+      end
+
+      def self.parse(cpi_hash)
+        name = safe_property(cpi_hash, 'name', :class => String)
+        version = safe_property(cpi_hash, 'type', :class => String)
```
tjvman: Change `version` to `type` for consistency

cppforlife:
not everything, blobstore and nats still stay in the cpi for the moment. will pull them out of cpi completely at some point.

ljfranklin:
@cppforlife agreed that `cloud.properties.{registry,agent,blobstore}` should come from cpi.json for now. But I would like for `cloud.properties.openstack` to come exclusively from dynamic values by removing the merge call.

voelzmo:
@cppforlife @ljfranklin We decided for `merge` instead of a simple replacement because of properties like [`use_dhcp`](http://bosh.io/jobs/openstack_cpi?source=github.com/cloudfoundry-incubator/bosh-openstack-cpi-release&version=27#p=openstack.use_dhcp). Users mostly don't really care about those and don't set them in their manifest, because "it just works".

I'd like to keep it this way without compromising the idea that defaults in a boshrelase are exactly in one place: the `spec` file. The current way isn't ideal, I know, because now CPI configuration is coming from two places instead of just one (if you're using this new feature). Seemed like the lesser evil to me, though.

ljfranklin:
@voelzmo I agree that all defaults should ideally go in a single place. But a CPI can have default properties for `vm_type.cloud_properties` or `disk_pool.cloud_properties` that cannot live in the spec file. So you now have some defaults in your code (e.g. AWS `disk_pools.cloud_properties.type`) and some in the spec file (e.g. `use_dhcp`). So the "all defaults in one place" rule for CPIs has to be defaults defined in the code. Defaults in the code are also much easier to unit test as compared to ERB rendering.

tjvman:
@MatthiasWinzeler In the `cpi_config` you used for manually testing the feature, were your `properties` for each CPI constructed like this?
```
- name: blah
  type: aws
  properties:
    openstack:
      some-prop: foo
      some-other-prop: bar
...
```

If you could provide an example `cpi_config` based on the one you used, that would be great too.

MatthiasWinzeler:
@cppforlife Good idea, I will change the default to "" then.

@tjvman My cpi config looked like this (omitted the third cpi for better readability):
```
cpis:
- name: openstack1
  type: openstack
  properties:
    openstack:
      auth_url: https://api.openstack1:5000/v2.0/tokens
      tenant: openstack1
      username: openstack1-user
      api_key: openstack1-password
      connection_options:
        ssl_verify_peer: true
        read_timeout: 300
        write_timeout: 300
      config_drive: disk
      use_dhcp: false
      default_key_name: cloudfoundry
      private_key: /home/ubuntu/bosh-workspace/ssh/cloudfoundry
      default_security_groups: []
      human_readable_vm_names: true
      endpoint_type: publicURL
      state_timeout: 300
      boot_from_volume: false
      stemcell_public_visibility: false
      wait_resource_poll_interval: 5
      ignore_server_availability_zone: false
- name: openstack2
  type: openstack
  properties:
    openstack:
      auth_url: https://api.openstack2:5000/v2.0/tokens
      tenant: openstack2
      username: openstack2-user
      api_key: openstack2-password
      connection_options:
        ssl_verify_peer: true
        read_timeout: 300
        write_timeout: 300
      config_drive: disk
      use_dhcp: false
      default_key_name: cloudfoundry
      private_key: /home/ubuntu/bosh-workspace/ssh/cloudfoundry
      default_security_groups: []
      human_readable_vm_names: true
      endpoint_type: publicURL
      state_timeout: 300
      boot_from_volume: false
      stemcell_public_visibility: false
      wait_resource_poll_interval: 5
      ignore_server_availability_zone: false
```

Some of the properties are just defaults of the openstack cpi. They are now picked up by default (see @voelzmo 's comment above) and could most likely be omitted.

tjvman:
@MatthiasWinzeler @voelzmo Thanks for this PR! We've been looking over the code and doing local testing and it's looking good so far. There are a couple issues that we'd like to bring to your attention before we finalize the review, though.

1. When using `cpi_config`, if you run `bosh task x --debug` then the logs will contain any credentials that were passed in the `request['context']` to the CPI; redact creds in the `context`  before sending them to debug logs. See `bosh_cpi/lib/cloud/external_cpi.rb:84`.

2. Always using the newest `cloud_config` when performing CPI calls breaks `bosh recreate` functionality (may break other things). We would like to preserve the versioning functionality for the CloudConfig, but from your initial PR comments it seemed like always using the latest CloudConfig was necessary. Can you give us some more context on why the newest CloudConfig has to be used so we can work together and figure out a solution that allows both the new PR and desired pre-existing functionality to work?

ljfranklin:
@cppforlife @MatthiasWinzeler I realize this thought is coming a little late, but do we actually need a separate `cpi-config` file and command or could we just add those properties to the `cloud-config`?

The distinction is more or less an implementation detail as the cloud config is already full of properties that will be passed to the CPI. I'm worried users not familiar with the internals of BOSH won't understand why some IaaS properties need to go in cloud-config and others in cpi-config. I'd imagine this also simplifies the multi-cpi implementation and solves the issue mentioned by @tjvman where multi-cpi could break `bosh recreate`. Let me know if I'm missing something.

cppforlife:
thinking behind separation was that cpi config isnt versioned just like cloud config. its likely that you will have to update/shift configuration over time for your cpi (cred change, etc) and i dont think we want users to have to redeploy for that.

voelzmo:
@ljfranklin @cppforlife Furthermore, cpi-config contains credentials and should potentially handled differently by admins and also the director than the rest of the cloud-config.

bosh_cpi/lib/cloud/external_cpi.rb
```diff
@@ -66,12 +67,15 @@ def ping(*arguments); invoke_cpi_method(__method__.to_s, *arguments); end
     private
 
     def invoke_cpi_method(method_name, *arguments)
+      context = {
```
MatthiasWinzeler: Makes sense. I removed the check for `context` in the openstack cpi: https://github.com/swisscom/bosh-openstack-cpi-release/commit/77970370f428447586925001414c4a84cb892d6d

bosh-director/spec/unit/cpi_config/parsed_cpi_config_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+                          'properties' => {
+                              'somekey' => 'someproperty'
+                          }
+                      }),
+        ]
+
+      }
+
+      subject(:parsed_cpi_config) { described_class.new(cpis) }
+      describe '#find_cpi_by_name' do
```
MatthiasWinzeler: No, there's not. I changed it.

bosh-director/spec/unit/deployment_plan/stemcell_spec.rb
```diff
@@ -9,8 +9,8 @@ def make_deployment(name)
     BD::Models::Deployment.make(:name => name)
   end
 
-  def make_stemcell(name, version, os = 'os1')
-    BD::Models::Stemcell.make(:name => name, :operating_system=>os, :version => version)
+  def make_stemcell(name, version, os = 'os1', params = {})
```
MatthiasWinzeler: It's not only `cpi`, but also `cid` (see further below in the spec). Should I add both as separate arguments? I prefer the `params` way since it's more readable.

bosh-director/lib/bosh/director/deployment_plan/stemcell.rb
```diff
@@ -79,6 +90,32 @@ def spec
         }
       end
 
+      def cid_for_az(az)
+        raise 'please bind model first' if @models.nil?
```
MatthiasWinzeler: I kept is a generic runtime exception on purpose - it should always happen during develop time and bubble up to the developer (it means he didn't bind the model before using `cid_for_az`). 

What's the preferred way in such situations? Should we still introduce a more specific error? Does it need to be put in `errors.rb` (even though it should never leave the system)?

MatthiasWinzeler:
Thanks for reviewing the changes! 
I pushed most of the requested changes and will complete the remaining open points in the coming days.

MatthiasWinzeler:
> Always using the newest cloud_config when performing CPI calls breaks bosh recreate functionality (may break other things). We would like to preserve the versioning functionality for the CloudConfig, but from your initial PR comments it seemed like always using the latest CloudConfig was necessary. Can you give us some more context on why the newest CloudConfig has to be used so we can work together and figure out a solution that allows both the new PR and desired pre-existing functionality to work?

It's best explained when looking at this [integration test](https://github.com/cloudfoundry/bosh/pull/1471/files#diff-593ff6d078d878162d64a9df4e2d08dcR255) which tests the switch from non-cpi-config to cpi-config:

- Deploy without cpi-config
- Create cpi-config
- Update cloud-config (add CPIs to AZs)
- Deploy

My assumption was that this works out of the box. Without my [ugly fix](https://github.com/cloudfoundry/bosh/pull/1471/files#diff-54a465ad1bd54f155ee42ef5a5a15991R122), this did not work - BOSH used the older cloud config while deploying, so it did not know which AZ uses which CPI and thus deployed to the director cpi.

You can easily reproduce this by removing [my hack](https://github.com/cloudfoundry/bosh/pull/1471/files#diff-54a465ad1bd54f155ee42ef5a5a15991R122) and run the integration test above.

Maybe this is some other bug?

MatthiasWinzeler:
Pushed some last improvements - it's ready now for another review - and merged current develop. [integration-1.9-postgres](https://main.bosh-ci.cf-app.com/teams/main/pipelines/bosh:multi-cpi/jobs/integration-1.9-postgres) has some broken specs, but it's the same on develop.

cppforlife:
@MatthiasWinzeler we are planning to take a look at the changes next week.

bosh-director/lib/bosh/director/deployment_plan/stemcell.rb
tjvman: That's a fair point, this will be fine for now.

bosh-director/spec/unit/deployment_plan/stemcell_spec.rb
tjvman: Ah, didn't catch the `cid`. I don't have strong feelings about params vs separate args in this case, so this seems fine as is.

tjvman:
@MatthiasWinzeler Thanks for making those changes. We looked into how the change to when the deployment model gets updated, and we believe we've identified the source of the bug as well as a way to fix it. 

With that in mind, we intend to merge this PR as it currently stands and make our change to how `deployment.cloud_config` works on top of that.

mingxiao:
@MatthiasWinzeler the `cpi-config` and `update-cpi-config` has also been ported over to the go bosh cli

youngm:
Nice work @MatthiasWinzeler !  Thanks tons for contributing this!

Issue #1468, opened 2016-10-10
bosh health manager reports metrics with incomplete names


holgero:
Bosh health manager intermittently reports via its graphite plugin metrics with names where the deployment name is missing. These metrics contain usually only one value. This means we have e.g. a metric with the name `CF..cell_z1.22b7cd86-cb63-4599-99b6-9da3b4faf529.0d0e45d8-87ca-420a-a086-d7613f1b6fd1.system_cpu_sys` that contains just one value and another metric with the correct name `CF.cf-diego.cell_z1.22b7cd86-cb63-4599-99b6-9da3b4faf529.0d0e45d8-87ca-420a-a086-d7613f1b6fd1.system_cpu_sys` which contains the expected values (one per minute).

This seems to occur often, but only for a short period of time (during startup of a new VM or at shutdown?).

The effect for us is that we have many useless metric names in the database (which is annoying in the Grafana dashboard editor, it displays a dropdown menu of the metric names). Would it be a good idea to filter out metrics without deployment names set in the health manager graphite plugin? If we created a pull request for that change, would it have a chance to get accepted?

holgero:
I created a pull request anyway: #1473

dpb587-pivotal:
Closing - as of ~v260.1, this should be working properly.

PR #1473, opened 2016-10-12
Skip sending incomplete events

holgero:
If a heartbeat does not contain all of a node_id, deployment
and job, skip it. This is for issue #1468.

### DIFF
--- a/bosh-monitor/lib/bosh/monitor/plugins/graphite.rb
+++ b/bosh-monitor/lib/bosh/monitor/plugins/graphite.rb
@@ -18,7 +18,7 @@ module Bosh::Monitor
 
       def process(event)
         if !valid?(event)
-          logger.warn("Skipping invalid heartbeat: #{event..to_s}")
+          logger.warn("Skipping invalid heartbeat: #{event.to_s}")
           return
         end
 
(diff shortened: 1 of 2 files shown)


cppforlife:
@holgero would be nice to log dropped events.

holgero:
I added a log of level 'warn', but I have no test for it. Is that OK?

holgero:
Forget my last comment, I added a test for the log message.

cppforlife:
Thank you. Story: https://www.pivotaltracker.com/story/show/132265151

cppforlife:
@holgero We have made improvements in the latest release; however, still observed lots of events with missing info. We've ended up pulling changes in and should make them available in the upcoming release version.

holgero:
@cppforlife Thank you for the update. I am not sure I got it correctly what you meant when you wrote: "We've ended up pulling changes in". Does this mean you pulled my change? Or did you say that you did some other changes that improve only some of the events but still lots of incomplete events are sent?

dpb587-pivotal:
Closing - starting ~v260.1, incomplete heartbeats should no longer be sent (via [76a6d7343](https://github.com/cloudfoundry/bosh/commit/76a6d7343b69dacf088df7a15ddbe800a0755c61?w=1) in [#132265151](https://www.pivotaltracker.com/story/show/132265151)).

PR #1483, opened 2016-10-20
Cleanup task output directory before start

barthy1:
[#114582319](https://www.pivotaltracker.com/story/show/114582319)


cppforlife:
pretty sure rmdir will fail if there are any files inside the dir which there will be some in the worst case. lets do rm_rf without a conditional.

barthy1:
@cppforlife Sorry, my bad. Fixed it.

PR #1487, opened 2016-10-21
Logs collecting from multiple instances

barthy1:
Logs can be collected from 1 instance, instances of 1 job and all instances in the deployment.

[#126754759](https://www.pivotaltracker.com/story/show/126754759)

Signed-off-by: Konstantin Maksimov kmaksimov@ru.ibm.com


bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
```diff
@@ -129,8 +129,7 @@ def initialize(config)
       # GET /deployments/foo/jobs/dea/2/logs
       get '/:deployment/jobs/:job/:index_or_id/logs' do
         job = params[:job]
-        index_or_id = params[:index_or_id]
-
+        index_or_id = params[:index_or_id] == "*" ? nil : params[:index_or_id]
```
cppforlife: does this work when `:job` is empty (needed for entire deployment)?

PR #1497, opened 2016-11-04
Also consider STEMCELL_BUILD_NUMBER when determining the build number

aduffeck:
40bf0b41 changed the way how the build number is determined (https://github.com/cloudfoundry/bosh/commit/40bf0b413b2608a7a439596d1762ff3f020db838#diff-ebd1e62819e1a31cfb010ad1aa609962R177).

The new method `Bosh::Dev::Build.build_number` does not consider `STEMCELL_BUILD_NUMBER`, though, which makes builds with local os images fail if `STEMCELL_BUILD_NUMBER` is set to something different than `0000` due to a version mismatch in `/etc/stemcell_version`.

### DIFF
--- a/bosh-dev/lib/bosh/dev/build.rb
+++ b/bosh-dev/lib/bosh/dev/build.rb
@@ -42,7 +42,7 @@ module Bosh::Dev
     end
 
     def self.build_number
-      ENV.fetch('CANDIDATE_BUILD_NUMBER', '0000')
+      ENV['CANDIDATE_BUILD_NUMBER'] || ENV['STEMCELL_BUILD_NUMBER'] || '0000'
     end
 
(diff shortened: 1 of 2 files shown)


dpb587-pivotal:
Closing - thanks for looking into this. We've recently moved stemcell functionality into a new repo ([cloudfoundry/bosh-linux-stemcell-builder](https://github.com/cloudfoundry/bosh-linux-stemcell-builder)) and we [no longer](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/bfd968105eba841cd7d6569fd9659b942030dd4b) use `STEMCELL_BUILD_NUMBER` (in favor of `CANDIDATE_BUILD_NUMBER`).

PR #1498, opened 2016-11-07
Exclude option for addons

barthy1:
[#133018773](https://www.pivotaltracker.com/story/show/133018773)

### DIFF
--- a/bosh-director/lib/bosh/director/errors.rb
+++ b/bosh-director/lib/bosh/director/errors.rb
@@ -290,8 +290,8 @@ module Bosh::Director
   RuntimeInvalidReleaseVersion = err(530001)
   RuntimeReleaseNotListedInReleases = err(530002)
   RuntimeInvalidDeploymentRelease = err(530003)
-  RuntimeIncompleteIncludeExcludeJobSection = err(530004)
-  RuntimeIncompleteIncludeExcludeStemcellSection = err(530005)
+  RuntimeIncompleteFilterJobSection = err(530004)
+  RuntimeIncompleteFilterStemcellSection = err(530005)
 
   # Config server errors
   ConfigServerMissingNames = err(540000)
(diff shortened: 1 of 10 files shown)


bosh-director/spec/unit/runtime_config/addon_spec.rb
```diff
@@ -218,6 +233,32 @@ module RuntimeConfig
...
+          let(:exclude_spec) { {'deployments' => [deployment_name]} }
+          let(:deployment_instance_group) {
+            instance_group_parser = DeploymentPlan::InstanceGroupSpecParser.new(deployment, Config.event_log, logger)
+            instance_group_parser.parse(Bosh::Spec::Deployments.dummy_job)
+          }
+
+          it 'applies' do
+            expect(addon.applies?(deployment_name, nil)).to eq(false)
+          end
+        end
```
dpb587-pivotal: I don't see a test for when both include and exclude are configured, but they are not a complete union. Perhaps we can add one?

bosh-director/lib/bosh/director/errors.rb
```diff
@@ -290,8 +290,8 @@ def self.err(error_code, response_code = BAD_REQUEST)
   RuntimeInvalidReleaseVersion = err(530001)
   RuntimeReleaseNotListedInReleases = err(530002)
   RuntimeInvalidDeploymentRelease = err(530003)
-  RuntimeIncompleteIncludeJobSection = err(530004)
-  RuntimeIncompleteIncludeStemcellSection = err(530005)
+  RuntimeIncompleteIncludeExcludeJobSection = err(530004)
+  RuntimeIncompleteIncludeExcludeStemcellSection = err(530005)
```
dpb587-pivotal: I see `AddonInclude` was renamed to `AddonFilter` - I think we should do the same thing here instead of `IncludeExclude`. e.g. `RuntimeIncompleteFilterJobSection`.

barthy1:
@dpb587-pivotal Will update the code asap

PR #1507, opened 2016-11-11
Avoid installing EFI grub

manno:
If the machine which is running the docker container is booting from EFI, then the
grub2-install inside the chroot, inside docker, will see
/sys/firmware/efi and decide to install an EFI bootloader. This will fail. The stemcell
should not be using EFI at all.

### DIFF
--- a/bosh-stemcell/spec/support/stemcell_shared_examples.rb
+++ b/bosh-stemcell/spec/support/stemcell_shared_examples.rb
@@ -20,6 +20,12 @@ shared_examples_for 'All Stemcells' do
     end
   end
 
+  context 'grub does not install an EFI bootloader' do
+    describe file('/boot/efi') do
+      it { should_not be_directory }
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


cppforlife:
@manno any stemcell test we can add to verify that efi bootloader isnt installed?

manno:
@cppforlife for us it will just fail on EFI systems since the grub2 installation is missing the EFI modules.

cppforlife:
@manno right, but we would like to be aware if we break it ourselves and a unit test will help with that.

manno:
Oh, you mean if there is a grub installation which is able to install EFI? Right, a test for that would be useful, since it probably won't boot.

I would check if `/boot/efi` was created and bail out. I'll try to add it to this PR later today.

cppforlife:
@manno yup that's exactly what i mean.

cppforlife:
Pulled this. Thanks.

PR #1520, opened 2016-11-24
use the more general class 'TrueClass' so make it compatible with mssql

andyliuliming:


andyliuliming:
made my Azure organization public now. so CLA not need to be signed here : )

andyliuliming:
here's one pull request I sent to sequel library, and Jeremy suggests us to use the TrueClass because it's more genernal
https://github.com/jeremyevans/sequel/pull/1264

PR #1531, opened 2016-12-09
Add context_id to tasks for deployments and errands

Samze:
CF BOSH Tracker story [#134702861](https://www.pivotaltracker.com/n/projects/956238/stories/134702861)

## Why

As a bosh user
I can assign a context to bosh deployments and errands
So that I can query a group of bosh tasks

This could be extended to other bosh tasks in future.

## What

This PR extends the BOSH Director API.
No changes to the bosh CLI.

Accept `X-Bosh-Context-Id` header in the following requests:
- `POST /deployments` - create & update deployment
- `DELETE /deployments/:deployment` - delete deployment
- `POST /deployments/:deployment/errands/:errand_name/runs` - run errand

Accept `context_id` query parameter for:
- `GET /tasks` - list tasks

Add `context_id` column to tasks table
Add `context_id` to JSON response for tasks

Sam and @jamesjoshuahill 

cc: @avade

### DIFF
--- a/src/bosh-director/db/migrations/director/20161209104649_add_context_id_to_tasks.rb
+++ b/src/bosh-director/db/migrations/director/20161209104649_add_context_id_to_tasks.rb
@@ -1,7 +1,8 @@
 Sequel.migration do
   change do
     alter_table :tasks do
-      add_column :context_id, String, default: '', null: false
+      add_column :context_id, String, size: 64, default: '', null: false
+      add_index :context_id
     end
   end
 end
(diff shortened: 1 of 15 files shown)


Samze:
Some questions:

- Is `varchar(255)` suitable for the new `context_id` column? with non-null, default `''`? We think a UUID of some kind would be sufficient, so it could be shorter.
- Should we add an index for the new `context_id` column?
- Is any user input santisation necessary for the `X-Bosh-Context-Id` header?
- Should the request be rejected if the header is larger than the database field?

cppforlife:
> Is varchar(255)... We think a UUID of some kind would be sufficient, so it could be shorter.

let's make it a bit shorter for now: `varchar(64)`.

> Should we add an index for the new context_id column? 

let's do that. see other migrations for adding indices (later ones for good examples).

> Is any user input santisation necessary for the X-Bosh-Context-Id header?

i think it's fine as it is. it's opaque to the Director at this point.

> Should the request be rejected if the header is larger than the database field?

sure.

Samze:
Thanks @cppforlife, see latest commit with improvements.

PR #1534, opened 2016-12-13
Move event and result logs to database

barthy1:
event_output and result_output columns in tasks table.
New information is appended to existing one
Support for byte range.

[#133019409](https://www.pivotaltracker.com/story/show/133019409)


src/bosh-director/spec/unit/db/migrations/director/20161128181900_add_logs_to_tasks_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        id: 1,
+        state: 'finished',
+        type: 'something',
+        deployment_name: 'test-deployment',
+        timestamp: '2016-04-14 11:53:42',
+        description: 'description',
+      }
+
+      DBSpecHelper.migrate(migration_file)
+
```
tylerschultz: This test should check the pre-existing task has an empty string in the new columns, per the test name. It should also check that it's able to add large strings to the new columns.

barthy1: @tylerschultz Mysql is not allowed to set default value for text type https://dev.mysql.com/doc/refman/5.7/en/blob.html. So I updated the test name and added extra checks, you mentioned. If you have an example, how it's possible to set default value for text, could you please share it with me?

zankich:
@barthy1 can you please resolve these merge conflicts, thanks!

src/bosh-director/spec/unit/db/migrations/director/20161128181900_add_logs_to_tasks_spec.rb
zankich: @barthy1 You can create a default value using the `before_create` [method](https://github.com/jeremyevans/sequel/blob/master/doc/model_hooks.rdoc#basic-usage).

Here's an example of `before_create` in the cloud config model
https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/models/cloud_config.rb#L5-L7

barthy1: @zankich approach with `before_create` in model does not work for migration, when there are already some records. Anyway I've added it for new records.

src/bosh-director/lib/bosh/director/task_db_writer.rb
```diff
@@ -0,0 +1,15 @@
...
+  class TaskDBWriter
+    def initialize(column, task)
+      @column_name = column
+      @task = task
+      @transactor = Transactor.new
+    end
+
+    def write(text)
+      @transactor.retryable_transaction(Bosh::Director::Config.db) do
+        @task.update({@column_name => "#{@task[@column_name]}#{text}"})
```
DennisDenuto: let's avoid pulling down existing value of the column when updating it. Instead we should be able to use sequel [concat](http://stackoverflow.com/questions/14263069/how-to-concat-a-string-with-a-table-field-with-sequel))

src/bosh-director/spec/unit/api/controllers/tasks_controller_spec.rb
```diff
@@ -399,8 +399,26 @@ module Api
...
+              # Range test
+              get "/#{task1.id}/output?type=event", {}, {'HTTP_RANGE' => 'bytes=5-'}
+              expect(last_response.status).to eq(206)
+              expect(last_response.body).to eq('output')
+              expect(last_response.headers['Content-Length']).to eq('6')
+              expect(last_response.headers['Content-Range']).to eq('bytes 5-10/11')
+            end
+
             it 'supports returning different types of output (debug, cpi, event)' do
-              %w(debug event cpi).each do |log_type|
```
DennisDenuto: Should the test name also mention `result`? Also the new `result output` behavior of persisting to the db is not being tested.

src/bosh-director/lib/bosh/director/task_db_writer.rb
barthy1: @DennisDenuto Thank you for the review. Could you please provide any sample of the sequel concat for model objects? So far I found that `Sequel.x` works only for dataset actions. And in case if it is the dataset action, to update the model object I need to use `task.refresh`, so no reduction of interactions with db.

src/bosh-director/spec/unit/api/controllers/tasks_controller_spec.rb
barthy1: I will extend the unit tests

src/bosh-director/lib/bosh/director/task_db_writer.rb
manno: @barthy1 we successfully used something like `Task.where(:id => task.id).update(description: Sequel.join([@column_name, "asdfasdf"]))`.
Refreshing the task model should not be necessary, we don't expect any code to read from the new logging columns while the task is running.

DennisDenuto:
LGTM, @barthy1 can you please resolve these merge conflicts so that I can Merge them in, thanks!

PR #1536, opened 2016-12-14
fix monit restart director_nginx did not work

xiujiao:
When I ran  `monit restart director_nginx`, it failed and `monit summary` showed that `Process director_nginx does not exist`. 

The reason is that the PID file does not match the PID of the running director_nginx process.

The fix in this PR directly borrows how the stop function is written for the `blobstore_nginx` job.

### DIFF
--- a/release/jobs/director/templates/nginx_ctl
+++ b/release/jobs/director/templates/nginx_ctl
@@ -9,6 +9,10 @@ SSL_DIR=/var/vcap/jobs/director/config/ssl
 TMP_DIR=/var/vcap/data/director/tmp
 RUNAS=vcap
 
+function pid_exists() {
+  ps -p $1 &> /dev/null
+}
+
 case $1 in
 
   start)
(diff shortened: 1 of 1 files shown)


cppforlife:
@xiujiao im not sure that this change fixes situation you are describing. at the worst prev code will call `kill ` (with an empty string) which will error but because script is not running with set -e it will continue on executing.

xiujiao:
@cppforlife Hi to be honest, the fix does not make 100% sense to me either. I used it because it fixed the problem in my deployment  and it is used in blobstore_nginx script in the repo.
I suspect  the PID management has some problem. My original fix actually is very different , please see my new commit https://github.com/cloudfoundry/bosh/pull/1536/commits/318ef0baca0adb04ca2c87752364dc310976603f

kitsirota:
+1 I just ran into this with 260.6

jaresty:
merge story [#152973456](https://www.pivotaltracker.com/story/show/152973456)

cppforlife:
closing in favor of completing bpm-izing of the bosh release: https://www.pivotaltracker.com/story/show/154148761

PR #1539, opened 2016-12-15
Update director monit

paolostivanin:
Increase timeout of `director` and `director_nginx` as we were facing problems with the default 30s timeout.
We tried many values and 75s is the best we found (we were facing problems on Openstack but not on AWS).


paolostivanin:
Even though the better approach would be to have it configurable...

cppforlife:
@paolostivanin i think you just though of what i will ask. let's make it configurable so that current (default) behaviour isnt affected.

paolostivanin:
Ok! I'll do it :) I close this PR and will open a new one!

PR #902, opened 2015-07-31
Add syslog forwarder plugin to HM

jsievers:
Add a healthmanager plugin that writes alerts to the syslog
using programname 'bosh.hm'.
If enabled, rsyslogd will be configured to
forward 'bosh.hm' messages to a remote syslog drain
(e.g. a syslog ingestor instance of an ELK stack)

The following new manifest configuration parameters
can be used to configure the plugin:

```
  hm.syslog_event_forwarder_enabled:
    description: Enable syslog forwarder plugin
    default: false
  hm.syslog_event_forwarder.address:
    description: IP address of the syslog drain
  hm.syslog_event_forwarder.port:
    description: Port of the syslog drain
    default: 5514
  hm.syslog_event_forwarder.transport:
    description: One of ["udp", "tcp", "relp"]
    default: tcp
```

Signed-off-by: Jan Sievers <jan.sievers@sap.com>
Also-by: Marco Voelz <marco.voelz@sap.com>
Also-by: Stephan Weber <stephan.weber@sap.com>

(edit: formatting)

PR #1540, opened 2016-12-15
Remove custom syslog forwarding for health_monitor

voelzmo:
Use the syslog-release instead. Using this custom forwarding
would break BOSH syslog configuration for auditing.

Effectively, this reverts d5d61f1550aebe6d9585aaab96b3a8ffc74f5f0c
and PR https://github.com/cloudfoundry/bosh/pull/902


voelzmo:
cc @jsievers

cppforlife:
@voelzmo nice. i wonder if we should raise an error in ERB if hm.syslog_event_forwarder_enabled specified so that it's easier to discover that this functionality is gone and ppl should use syslog-release.

voelzmo:
@cppforlife added deprecation error, should I just merge it?

cppforlife:
yup.

PR #1541, opened 2016-12-16
Add configurable timeout for monit director and director_nginx jobs

paolostivanin:
We saw that the default 30s timeout is enough for our deployment. So with this PR I introduced the possibility to configure the monit timeout for the director and director_nginx jobs.


paolostivanin:
bosh v260.1 didn't solve the problem for us. The only way to have bosh deployable is by setting a higher timeout...

voelzmo:
@paolostivanin is it still the case that v260.1 didn't fix it after our debugging last week?

paolostivanin:
@voelzmo no, everything works fine after our debugging session!

PR #1547, opened 2016-12-20
Add config server to info endpoint

Samze:
## Why

As a bosh API consumer
I want to know whether a config server is enabled
So that I can discover the config server

## What

Add config server details to `GET /info` endpoint

Sam & @jamesjoshuahill

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/info_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/info_controller.rb
@@ -26,9 +26,16 @@ module Bosh::Director
             },
             'snapshots' => {
               'status' => Config.enable_snapshots
+            },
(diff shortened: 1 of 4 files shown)


cppforlife:
@Samze please change `url` to `urls` and make it an array of item one.

Samze:
Hey @cppforlife, thanks for looking at this so quickly.

That was the original idea, however after looking at the [directors job spec for config_server](https://github.com/cloudfoundry/bosh/blob/master/jobs/director/spec#L144) we noticed that you can only specify a single url. Are there other reasons to have an array of one?

cppforlife:
@Samze upcoming compatibility. we already changed uaa to have `urls`.

Samze:
@cppforlife Changed to `urls`

PR #1560, opened 2017-01-12
Add Json health_monitor plugin

robertjsullivan:
* sends heartbeats to stdin of any process located in
`/var/vcap/jobs/*/bin/bosh-monitor/<any executable>`
* Change the system.healthy metric from int to string for conformity
* Translate the metric value to string when hashing the metrics object.

Signed-off-by: Robert Sullivan <rsullivan@pivotal.io>


src/bosh-monitor/README.md
```diff
@@ -46,6 +46,7 @@ Among the included plugins are:
 - Emailer - Sends configurable Emails on events reciept
 - Consul Event Forwarder - Sends heartbeats as events and TTL checks to a consul cluster
 - EventLogger - Stores events in Director DB
+- Json Emitter - Sends metrics in json format to any binaries in /var/vcap/jobs/*/bin/boshmonitor
```
cppforlife: glob is actually `'/var/vcap/jobs/*/bin/bosh-monitor/*'`

jobs/health_monitor/templates/health_monitor.yml.erb
```diff
@@ -101,6 +101,16 @@ if p('hm.email_notifications')
   params['plugins'] << smtp_plugin
 end
 
+if p('hm.json_enabled')
+ params['plugins'] << {
+    'name' => 'json',
+    'events' => [
+      'heartbeat'
```
cppforlife: let's also send `alert`

src/bosh-monitor/lib/bosh/monitor/metric.rb
```diff
@@ -12,5 +12,14 @@ def initialize(name, value, timestamp, tags)
       @timestamp = timestamp
       @tags = tags
     end
+
+    def to_hash
+      {
+          :name => @name,
+          :value => @value.to_s,
```
cppforlife: seems a bit weird to cast it to string given that some of these values are ints: https://github.com/cloudfoundry/bosh/blob/c6c419f9eff844c35c8c7307e977427f05a23a51/src/bosh-monitor/lib/bosh/monitor/events/heartbeat.rb#L107

src/bosh-monitor/lib/bosh/monitor/plugins/json.rb
```diff
@@ -0,0 +1,31 @@
...
+        return false
+      end
+
+      @processes = Dir[bin_glob].map do |bin|
+        EventMachine::DeferrableChildProcess.open(bin)
+      end
+    end
+
+    def process(event)
+      return false if event.is_a? Bosh::Monitor::Events::Alert
```
cppforlife: allow alerts as well.

src/bosh-monitor/lib/bosh/monitor/metric.rb
jtarchie: Only one of the values are ints. It was nice to keep it consistent, especially when parsing the values for a golang struct.

cppforlife: i see. seems reasonable then.

AmberAlston:
@cppforlife do you have an estimate for when this might be merged in? Had trouble locating the backlog story to answer the question myself. Thank you!

cppforlife:
Merged into develop. Changed it to be enabled by default.

cppforlife:
cc @frodenas

PR #1577, opened 2017-02-06
Add endpoint to get one event

barthy1:
[#136637059](https://www.pivotaltracker.com/story/show/136637059)

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
@@ -9,12 +9,10 @@ module Bosh::Director
       get '/:id' do
         content_type(:json)
 
-        event_id = params[:id]
+        event_id = params[:id].to_i
         event = Models::Event[event_id]
         if event.nil?
-          status(400)
-          body("Event #{event_id} not found")
-          return
+          not_found
         end
         json_encode(@event_manager.event_to_hash(event))
       end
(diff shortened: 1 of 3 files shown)


src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
```diff
@@ -6,6 +6,19 @@ module Api::Controllers
     class EventsController < BaseController
       EVENT_LIMIT = 200
 
+      get '/:id' do
+        content_type(:json)
+
+        event_id = params[:id]
+        event = Models::Event[event_id]
```
cppforlife: not sure if we need to turn it back into integer: https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/api/event_manager.rb#L12 ?

src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
```diff
@@ -6,6 +6,19 @@ module Api::Controllers
...
+        if event.nil?
+          status(400)
```
cppforlife: 404?

src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
barthy1: @cppforlife it works as is. Do I need to update the code to support event ids as integer?

cppforlife: if it works i guess not. im somewhat surprised.

barthy1: I had a deeper look at it, and also don't understand the reason, why it works. So I've updated the code to feel more comfortable

PR #1583, opened 2017-02-13
Add stemcell_formats property for stemcell manifest and check if cpi …

barthy1:
…supports the given stemcell format

[#135954883](https://www.pivotaltracker.com/story/show/135954883)


src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -141,6 +151,20 @@ def update_steps
...
+
+      def is_supported?(cloud)
+        info = cloud[:cpi].info
+        unless @stemcell_formats.nil? || info["stemcell_formats"].nil?
+          return (info["stemcell_formats"] & @stemcell_formats).empty? ? false : true
+        else
+          logger.info("There is no enough information to check if stemcell format is supported")
+          return true
+        end
+      rescue
```
cppforlife: let's only rescue particular error, i think it's NotImplemented

barthy1: by default with old versions of bosh_cpi, cpis return another error - InvalidCall, converted to UnknownError with varied text messages. Is it particular enough to check?

src/bosh-director/lib/cloud/dummy.rb
```diff
@@ -363,6 +368,12 @@ def spawn_agent_process(agent_id, legacy_agent_path = nil)
 
       private
 
+      def method_missing(method, *args, &block)
+        @outer.send(method, *args, &block)
```
tylerschultz: `@outer` is not defined anywhere, nor is it assigned from anywhere that I can find in this PR. Is there something I don't understand? 
Why is method missing overridden?

src/spec/gocli/integration/cli_stemcell_spec.rb
```diff
@@ -32,13 +32,18 @@
...
   context 'if cpi config is used' do
-    it 'creates a stemcell for each configured cpi' do
+    it 'creates a stemcell for each configured cpi if it matches stemcell format' do
       stemcell_filename = spec_asset('valid_stemcell.tgz')
 
       cpi_path = current_sandbox.sandbox_path(Bosh::Dev::Sandbox::Main::EXTERNAL_CPI)
       cpi_config_manifest = yaml_file('cpi_manifest', Bosh::Spec::Deployments.simple_cpi_config(cpi_path))
       bosh_runner.run("update-cpi-config #{cpi_config_manifest.path}")
 
+      stemcell_filename_not_dummy = spec_asset('valid_stemcell_not_dummy_stemcell_format.tgz')
```
zankich: These new assertions should be in a different `it` block. This test currently mixes valid stemcells and invalid stemcells, so the expected behavior is testing two different things. There should be a new test which tests the unsupported stemcells explicitly.

So there should be one test which successfully uploads a stemcell and we can see the stemcells in the stemcell table, and a new test which uploads a stemcell, with the wrong format, and the stemcells table should not have this new stemcell.

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
```diff
@@ -324,6 +326,77 @@
       end
     end
 
+    context 'when information about stemcell formats is not enough' do
```
zankich: The descriptions of the`it` blocks in this `context` should be updated to convey the assertion in the test. Currently they read like nested `context`, and give no clear indication of what the test is trying to prove.

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -141,6 +151,20 @@ def update_steps
         steps += 1 if @stemcell_sha1 # also verify remote stemcell
         steps + cloud_factory.all_configured_clouds.count * 3 # check, upload and save for each cloud
       end
+
+      def is_supported?(cloud)
+        info = cloud[:cpi].info
+        unless @stemcell_formats.nil? || info["stemcell_formats"].nil?
```
zankich: For readability, can you please switch this to use an `if` instead of an `unless`.

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -141,6 +151,20 @@ def update_steps
...
+          return (info["stemcell_formats"] & @stemcell_formats).empty? ? false : true
```
zankich: The ternary operator is unnecessary here, you can simple negate the value of `empty?` and achieve the same behavior.

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -81,41 +82,50 @@ def perform
         stemcell = nil
         cloud_factory.all_configured_clouds.each do |cloud|
           cpi_suffix = " (cpi: #{cloud[:name]})" unless cloud[:name].blank?
-
+          supported = true
```
zankich: It looks like you could move the `is_supported?` check to this location instead of nesting the `supported` logic throughout this method. If you did an `if !is_supported?` at this location, you could fast fail out of this method. There you could log the cpi format error, raise `StemcellNotSupported` and no other modifications to this method are necessary.

src/Gemfile.lock
```diff
@@ -75,7 +75,7 @@ PATH
   specs:
     bosh-registry (0.0.0)
       aws-sdk
-      bosh_cpi
+      bosh_cpi (~> 2.0.6)
```
zankich: Is locking this gem version necessary?

src/bosh-registry/bosh-registry.gemspec
```diff
@@ -24,5 +24,5 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'yajl-ruby', '~>1.2.0'
   spec.add_dependency 'fog-openstack'
   spec.add_dependency 'aws-sdk'
-  spec.add_dependency 'bosh_cpi'
+  spec.add_dependency 'bosh_cpi',  '~>2.0.6'
```
zankich: Is locking this gem version necessary?

zankich:
@barthy1 can you also resolve the merge conflicts please, thanks!

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
barthy1: I just didn't want to do extra checks (+ cpi call) in case if it is not necessary and stemcell is already available. But I see your point and will update the code

src/Gemfile.lock
barthy1: I updated it, because there was "connected" error in unit test. But now I see that I had to fix unit test instead. Thank you for the good catch

beyhan:
We introduced a new method for the CPIs called`set_disk_metadata`. We found out that different CPIs  behave different in case of a notimplemented/missing method. With this commit:  https://github.com/cloudfoundry/bosh/commit/3d2b2340b241e204e29096396ccd608ece2ab69a we introduced a workaround until its handled properly in the CPIs. We saw that in this pull request you also handle some of the cases. May be you can have a look into the commit.

barthy1:
@beyhan thank you for the info. I found your code very useful. I will update my PR according to it

src/bosh-director/lib/cloud/dummy.rb
```diff
@@ -372,11 +372,6 @@ def spawn_agent_process(agent_id, legacy_agent_path = nil)
       end
 
       private
-
-      def method_missing(method, *args, &block)
```
dpb587-pivotal: Nice cleanup

dpb587-pivotal:
Oops; was curious and only looking at a single commit. Ignore me.

PR #1590, opened 2017-02-16
User can pause picking of new delayed jobs.

knm3000:
[#133603609](https://www.pivotaltracker.com/story/show/133603609)

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/tasks_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/tasks_controller.rb
@@ -1,5 +1,4 @@
 require 'bosh/director/api/controllers/base_controller'
-require 'yaml'
 
 module Bosh::Director
   module Api::Controllers
@@ -110,10 +109,6 @@ module Bosh::Director
           task.save
         end
 
-        if Models::DirectorAttribute.get_attribute('tasks_paused')
-          task.state = :paused
-        end
-
         content_type(:json)
         json_encode(@task_manager.task_to_hash(task))
       end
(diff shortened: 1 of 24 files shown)


src/bosh-director/lib/bosh/director/api/controllers/tasks_controller.rb
```diff
@@ -109,6 +110,10 @@ def filter_task_by_deployment_and_teams(dataset, deployment, token_scopes, limit
           task.save
         end
 
+        if Models::DirectorAttribute.get_attribute('tasks_paused')
```
cppforlife: i dont think we should do this since task may be already in progress.

src/bosh-director/lib/bosh/director/api/controllers/tasks_controller.rb
```diff
@@ -138,6 +143,27 @@ def filter_task_by_deployment_and_teams(dataset, deployment, token_scopes, limit
         end
       end
 
+      put '/', consumes: :json do
```
cppforlife: @knm3000 i was just thinking more about this, should we have a new controller: /tasks_configs which provides GET and PUT with a single value `{"paused": true|false}`. that way endpoints are a bit more symmetric.

cc @tylerschultz

src/spec/gocli/integration/cli_tasks_spec.rb
```diff
@@ -13,4 +13,31 @@
...
+    deploy_from_scratch
+    out = bosh_runner.run('pause-tasks')
+    expect(out).to include("Succeeded")
+    out = bosh_runner.run('env')
+    expect(out).to include("pause_tasks: enabled")
+
+    sleep 3
+
+    output = bosh_runner.run('delete-deployment', deployment_name: 'simple')
+    expect(output).to match(/Task \d+\. Paused/)
```
cppforlife: "paused" shouldnt be a state, but rather inactivity of a queue.

src/bosh-director/lib/bosh/director/api/controllers/tasks_configs_controller.rb
```diff
@@ -0,0 +1,28 @@
+require 'bosh/director/api/controllers/base_controller'
+
+module Bosh::Director
+  module Api::Controllers
+    class TasksConfigsController < BaseController
+      put '/', consumes: :json do
```
cppforlife: let's also add 'get /task_configs' and make it run similarly to runtime_configs.

cppforlife:
@knm3000 after thinking a bit more and looking at the code, i think we should change the direction and have it be more consistent with other "configs" (cpi, runtime, cloud). instead of having bosh pause-tasks command let's have `bosh update-task-config` and `bosh task-config`. that yaml like other command will allow to specify `paused: true/false` (let's default to false). this way this becomes a first class feature and we dont have to keep on expanding surface area of /info endpoint. this will also be helpful when we expand pause criteria to be able to pause particular team's tasks, etc.

cppforlife:
@knm3000 @barthy1 closing this one since generic configs made it to master.

PR #1603, opened 2017-02-27
Event for run errand

barthy1:
[#139265045](https://www.pivotaltracker.com/story/show/139265045)


src/bosh-director/spec/unit/errand/runner_spec.rb
```diff
@@ -152,22 +152,6 @@ module Bosh::Director
             end
           end
```
zankich: It seems like it would be better to change the behavior of this test, rather than removing it completely, so that we're testing that we now return the errand_result object rather than the description.

src/bosh-director-core/spec/bosh/director/core/templates/rendered_job_instance_spec.rb
```diff
@@ -3,6 +3,7 @@
 require 'bosh/director/core/templates/rendered_job_template'
 require 'bosh/director/core/templates/rendered_file_template'
 require 'bosh/director/agent_client'
+require 'securerandom'
```
zankich: Is this require necessary?

barthy1: It seems that this is not necessary. We had `NameError: uninitialized constant Bosh::Director::Core::Templates::SecureRandom` error on the machine without this require. Just checked on another machine and everything worked well. So, problem with local config.

PR #1622, opened 2017-03-14
remove unneded echo of pid as nginx has same path for pid

Lafunamor:
This commit removes the unnecessary creation of the nginx pid file in the nginx_ctl script as nginx will create the pid itself.
The removed line causes an issue if the start script is called when nginx is already running. The pid will be overwritten and causes monit to fail. Further monit is no longer able to stop the process.

### DIFF
--- a/jobs/blobstore/spec
+++ b/jobs/blobstore/spec
@@ -8,6 +8,7 @@ templates:
   mime.types:         config/mime.types
   read_users.erb:     config/read_users
   write_users.erb:    config/write_users
+  utils.sh:           helpers/utils.sh
 
 packages:
 - nginx
--- a/jobs/blobstore/templates/nginx_ctl
+++ b/jobs/blobstore/templates/nginx_ctl
@@ -12,10 +12,16 @@ function pid_exists() {
   ps -p $1 &> /dev/null
 }
 
+source /var/vcap/packages/common/utils.sh
+
 case $1 in
 
   start)
     mkdir -p $RUN_DIR $LOG_DIR $STORE_DIR $TMP_DIR
+
+    pid_guard $PIDFILE "nginx"
+
+    echo $$ > $PIDFILE
 
     chown -R $RUNAS:$RUNAS $RUN_DIR $LOG_DIR $STORE_DIR $TMP_DIR
 
--- a/jobs/blobstore/templates/nginx_ctl
+++ b/jobs/blobstore/templates/nginx_ctl
@@ -12,7 +12,7 @@ function pid_exists() {
   ps -p $1 &> /dev/null
 }
 
-source /var/vcap/packages/common/utils.sh
+source $JOB_DIR/helpers/utils.sh
 
 case $1 in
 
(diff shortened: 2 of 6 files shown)


cppforlife:
@tylerschultz @dpb587-pivotal looks good?

tylerschultz:
Seems lovely. When can I click merge?

If I understand correctly, this line is why the nginx_ctl line can be deleted:
https://github.com/cloudfoundry/bosh/blob/master/jobs/director/templates/nginx.conf.erb#L5

Also, it seems we need to give the same treatment to the blobstore nginx_ctl file:
https://github.com/cloudfoundry/bosh/blob/master/jobs/blobstore/templates/nginx_ctl#L20

Lafunamor:
@tylerschultz correct. Line 5 defines the PID location.
I've removed the line from blobstore nginx as well.

dpb587-pivotal:
Another option would be to leave the ctl scripts the way they are and remove/change the nginx.conf files to not write pids or write to another pid file (do you happen to know if `pid` is optional?). This would help keep the PID file management in one place, where it would be easier to understand and would be more consistent.

By writing the PID in the ctl script, we avoid any ambiguity between when nginx is attempting to start/bind/write the PID file (to avoid monit trying to repeatedly start it again).

Lafunamor:
@dpb587-pivotal I disagree. A start script should never write the PID file for an appllication unless the application cannot do itself.
As mentioned before writing the PID file in the start script causes the problem that wenn the start script is run while nginx is still running, the start script will overwrite the PID file. This will then result in monit not being able to monitor the process and fail after trying to start  nginx multiple times.
The system will not recover unless an operator reacts to the problem.

voelzmo:
I actually had the same thought as @dpb587 when I saw this change. Having it in the `ctl` scripts for every job makes it crystal clear when what happens. I guess is might be a matter of perspective: When you're coming from the BOSH world, it makes sense to do it in the script, when you're coming from the software you're installing as a release, you want to do it like you would do without using BOSH.

MatthiasWinzeler:
Could we fix this by using the `pid_guard` helper function that is used in a lot of bosh releases?
i.e. https://github.com/cloudfoundry/cf-release/blob/ef70b3f389a44bd5065745b7b30bfce45a8acccd/jobs/collector/templates/collector_ctl.erb#L16 / https://github.com/cloudfoundry/cf-release/blob/690e8155510acac067193f1076c74ce6312c0479/src/common/utils.sh#L7

This would ensure that the PID of the `ctl`is not writing its PID when nginx is already running (the exact error case Lafunamor describes).

Lafunamor:
I've added pidguard as @MatthiasWinzeler suggested. Please let me know if this is better.

jaresty:
merge story [#152973308](https://www.pivotaltracker.com/story/show/152973308)

jaresty:
@Lafunamor thanks for your contribution and sorry for the delay.
We fixed a [small issue](https://github.com/cloudfoundry/bosh/commit/cdb9ca69652c78fd7f1a8f2739b62175683f7cd4) and tested that everything works as expected.

PR #1628, opened 2017-03-17
Add deployment level addons

barthy1:
- All options are supported except deployment filter

[#139612921](https://www.pivotaltracker.com/story/show/139612921)

### DIFF
--- a/src/bosh-director/lib/bosh/director/addon/addon_filter.rb
+++ b/src/bosh-director/lib/bosh/director/addon/addon_filter.rb
@@ -1,5 +1,5 @@
 module Bosh::Director
-  module RuntimeConfig
+  module Addon
     class AddonFilter
 
       extend ValidationHelper
@@ -11,8 +11,11 @@ module Bosh::Director
         @filter_type = filter_type
       end
 
-      def self.parse(addon_filter_hash, filter_type)
+      def self.parse(addon_filter_hash, filter_type, deployment_level = false)
         applicable_deployment_names = safe_property(addon_filter_hash, 'deployments', :class => Array, :default => [])
+        if deployment_level && !applicable_deployment_names.empty?
+          raise DeploymentFilterNotAllowed, 'Deployment filter is not allowed for deployment level addons.'
+        end
         applicable_jobs = safe_property(addon_filter_hash, 'jobs', :class => Array, :default => [])
         applicable_stemcells = safe_property(addon_filter_hash, 'stemcell', :class => Array, :default => [])
 
@@ -48,7 +51,7 @@ module Bosh::Director
           name = safe_property(job, 'name', :class => String, :default => '')
           release = safe_property(job, 'release', :class => String, :default => '')
           if name.empty? || release.empty?
-            raise RuntimeIncompleteFilterJobSection.new("Job #{job} in runtime config's #{filter_type} section must " +
+            raise IncompleteFilterJobSection.new("Job #{job} in config's #{filter_type} section must " +
               'have both name and release.')
           end
         end
@@ -57,7 +60,7 @@ module Bosh::Director
       def self.verify_stemcells_section(applicable_stemcells, filter_type)
         applicable_stemcells.each do |stemcell|
           if safe_property(stemcell, 'os', :class => String, :default => '').empty?
-            raise RuntimeIncompleteFilterStemcellSection.new("Stemcell #{stemcell} in runtime config's #{filter_type} " +
+            raise IncompleteFilterStemcellSection.new("Stemcell #{stemcell} in config's #{filter_type} " +
               'section must have an os name.')
           end
         end
--- a/src/bosh-director/lib/bosh/director/addon/addon_parser.rb
+++ b/src/bosh-director/lib/bosh/director/addon/addon_parser.rb
@@ -1,0 +1,37 @@
+module Bosh::Director
+  module Addon
+    class AddonParser
+      include ValidationHelper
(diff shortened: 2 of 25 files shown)


src/bosh-director/lib/bosh/director/errors.rb
```diff
@@ -289,10 +289,11 @@ def self.err(error_code, response_code = BAD_REQUEST)
   # Addons
   RuntimeAmbiguousReleaseSpec = err(530000)
   RuntimeInvalidReleaseVersion = err(530001)
-  RuntimeReleaseNotListedInReleases = err(530002)
+  ReleaseNotListedInReleases = err(530002)
   RuntimeInvalidDeploymentRelease = err(530003)
-  RuntimeIncompleteFilterJobSection = err(530004)
-  RuntimeIncompleteFilterStemcellSection = err(530005)
+  IncompleteFilterJobSection = err(530004)
```
dpb587-pivotal: We're removing the `Runtime` namespace which makes sense, but perhaps we could keep an `Addon` prefix?

src/spec/integration/deploy_job_with_addons_spec.rb
```diff
@@ -38,8 +38,6 @@
     expect(File.exist?(foobar_vm.job_path('foobar'))).to eq(true)
   end
 
-
-
```
dpb587-pivotal: 👍

src/bosh-director/lib/bosh/director/addon/addon_filter.rb
```diff
@@ -1,5 +1,5 @@
 module Bosh::Director
-  module RuntimeConfig
+  module Addon
     class AddonFilter
```
dpb587-pivotal: Can we rename this class to `Filter` since it's now namespaced in `Addon` to avoid redundancy?

src/bosh-director/lib/bosh/director/addon/addon_parser.rb
```diff
@@ -0,0 +1,37 @@
+module Bosh::Director
+  module Addon
+    class AddonParser
```
dpb587-pivotal: Same thing as `Filter` - make sense to rename to `Bosh::Director::Addon::Parser`?

src/bosh-director/lib/bosh/director/addon/addon_filter.rb
```diff
@@ -11,8 +11,11 @@ def initialize(applicable_jobs, applicable_deployment_names, applicable_stemcell
         @filter_type = filter_type
       end
 
-      def self.parse(addon_filter_hash, filter_type)
+      def self.parse(addon_filter_hash, filter_type, deployment_level = false)
```
dpb587-pivotal: Rather than a boolean value here, could we consider a constant representing the scope? For example, `Bosh::Director::Addon::CONTEXT_RUNTIME` (or DIRECTOR?) and `...::CONTEXT_DEPLOYMENT`? Then we could also use it in the `Addon` class, too. It makes arguments a bit more explicit and we don't have to wonder what a boolean value means.

src/bosh-director/lib/bosh/director/addon/addon_filter.rb
```diff
@@ -48,7 +51,7 @@ def self.verify_jobs_section(applicable_jobs, filter_type)
           name = safe_property(job, 'name', :class => String, :default => '')
           release = safe_property(job, 'release', :class => String, :default => '')
           if name.empty? || release.empty?
-            raise RuntimeIncompleteFilterJobSection.new("Job #{job} in runtime config's #{filter_type} section must " +
+            raise IncompleteFilterJobSection.new("Job #{job} in config's #{filter_type} section must " +
```
dpb587-pivotal: Should we be explicit about where the bad addon was located to avoid the scenario where a runtime config has an error, but the user is getting an error about something they can't find in their deployment manifest.

src/spec/gocli/integration/deploy_job_with_addons_spec.rb
```diff
@@ -3,180 +3,267 @@
...
+      runtime_config = Bosh::Spec::Deployments.runtime_config_with_addon_includes
+      runtime_config['addons'][0]['include'] = {'jobs' => [
+        {'name' => 'foobar', 'release' => 'bosh-release'}
+      ]}
+      runtime_config['addons'][0]['exclude'] = {'deployments' => ['dep2']}
 
-    runtime_config_file = yaml_file('runtime_config.yml', runtime_config)
-    expect(bosh_runner.run("update-runtime-config #{runtime_config_file.path}")).to include('Succeeded')
+      runtime_config_file = yaml_file('runtime_config.yml', runtime_config)
+      expect(bosh_runner.run("update-runtime-config #{runtime_config_file.path}")).to include('Succeeded')
```
dpb587-pivotal: 👍  reformatting

barthy1:
@dpb587-pivotal Thank you for your opinion. I think your ideas are very reasonable. I've updated the code.

antonsoroko:
@barthy1 
Hi!
Thanks for this useful feature.

Is there any ETA for "deployments exclusions" feature?

Thanks in advance.

barthy1:
@antonsoroko, what feature do you mean? deployment filter is available for exclude section in addons and deployment level addons can contain exclude section

antonsoroko:
@barthy1 
Actually `exclude` sections works, my problem was in release version check.
Thanks.

PR #1648, opened 2017-04-06
Added orphan-disk command

knm3000:
[#136520793](https://www.pivotaltracker.com/story/show/136520793)

Signed-off-by: Yulia Gaponenko <yulia.gaponenko1@de.ibm.com>

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/disks_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/disks_controller.rb
@@ -20,7 +20,12 @@ module Bosh::Director
 
       delete '/:orphan_disk_cid' do
         job_queue = JobQueue.new
-        task = Bosh::Director::Jobs::DeleteOrphanDisks.enqueue(current_user, [params[:orphan_disk_cid]], job_queue)
+        orphan_param = params['orphan']
+        if orphan_param.nil? || orphan_param == 'false'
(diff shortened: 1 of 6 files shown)


src/bosh-director/lib/bosh/director/api/controllers/disks_controller.rb
```diff
@@ -20,7 +20,12 @@ class DisksController < BaseController
 
       delete '/:orphan_disk_cid' do
         job_queue = JobQueue.new
-        task = Bosh::Director::Jobs::DeleteOrphanDisks.enqueue(current_user, [params[:orphan_disk_cid]], job_queue)
+        orphan_param = params['orphan']
+        if orphan_param.nil? || orphan_param == 'false'
+          task = Bosh::Director::Jobs::DeleteOrphanDisks.enqueue(current_user, [params[:orphan_disk_cid]], job_queue)
+        else
+          task = Bosh::Director::Jobs::OrphanDiskJob.enqueue(current_user, params[:orphan_disk_cid], job_queue)
+        end
```
dpb587-pivotal: I disagree with using `DELETE` for both deletion, but also non-deletion behavior. Technically I think this endpoint should delete the CID, regardless of orphaning status (even if we error for active disks).

Can we consider a `post /:orphan_disk_cid/orphan` endpoint instead?

dpb587-pivotal: Chatted about this more. Ignore my comment.

Long-term, the committed code seems like the most practical approach (though I still think it's going to be a slightly confusing API outside of the CLI).

PR #1650, opened 2017-04-09
Fix date header in bosh-monitor email plugin to conform with rfc5322

bpicode:
The date header in the email plugin of bosh-monitor did not conform to rfc5322, causing problems in some mail clients.


dpb587-pivotal:
Thanks for the PR! I know it's a small change, but if you wouldn't mind signing the CLA we can merge this change in more easily.

bpicode:
Thanks for the reply. Actually I did sign the CLA, it seems @cfdreddbot didn't notice?

PR #1655, opened 2017-04-13
Specified credentials_source as a required option for backup destination

simonkey007:


dpb587-pivotal:
Thanks for the PR! This reminded me that the spec should be changed in a slightly different way, so f80997af21311c8a855f0fb5b6b3c1a3168cb868 includes this change. Thanks again for mentioning it.

Please feel free to still sign the CLA for your next PR... then dreddbot won't feel quite so concerned about you next time :)

PR #1677, opened 2017-05-08
PR improve meltdown handling

coreyti:
This PR improves the bosh-monitor handling of meltdown state. The key changes are:

- Adds additional deployment/instance details to logs and alerts, for debugging purposes.
- Increases the fidelity of severity in alerts, which (notably) means that many alerts that previously reported as "normal" now show actual status.
- Adds consideration of alert severity when determining meltdown state. Specifically, "warning" alerts no longer trigger "meltdown" state.

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/events/alert.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/events/alert.rb
@@ -58,13 +58,16 @@ module Bosh::Monitor
 
       def to_hash
         {
-          :kind       => "alert",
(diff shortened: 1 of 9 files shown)


coreyti:
Note that specs are currently failing at this point, due (in part) to the fact that the specs are passing a timestamp, instead of the Alert object that is now expected.

Sorry. We need to fix that before merging.

coreyti:
This PR is superseded by https://github.com/cloudfoundry/bosh/pull/1680

PR #1679, opened 2017-05-09
Change object_id to object_name in events query

barthy1:
(#144882283)[https://www.pivotaltracker.com/story/show/144882283]


barthy1:
This PR works together with https://github.com/cloudfoundry/bosh-cli/pull/217

src/bosh-director/lib/bosh/director/api/controllers/events_controller.rb
```diff
@@ -73,8 +73,8 @@ class EventsController < BaseController
           events = events.where(object_type: params['object_type'])
         end
 
-        if params['object_id']
-          events = events.where(object_name: params['object_id'])
+        if params['object_name']
+          events = events.where(object_name: params['object_name'])
```
cppforlife: @dpb587-pivotal i would say not for this case.

PR #1680, opened 2017-05-09
Improve HM's handling of "meltdown" state and "rogue" VMs (v2)

coreyti:
This PR improves the bosh-monitor handling of meltdown state and of "rogue" VMs. The key changes are:

- Adds additional deployment/instance details to logs and alerts.
- Increases the fidelity of severity in alerts, which (notably) means that many alerts that previously reported as "normal" now show actual status (warning/error).
- Adds consideration of alert *category* when determining meltdown state. Specifically, "VM health" alerts might trigger "meltdown" state, but others (such as SSH login/logout) will not do so.
- Adds a check to see that an does not have VM, in addition to it expecting one. This allows all instances to be recorded, there by resolving an issue in which errand jobs were considered "rogue".

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/deployment.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/deployment.rb
@@ -33,11 +33,6 @@ module Bosh::Monitor
         return false
       end
 
-      unless instance.expects_vm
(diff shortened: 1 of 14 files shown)


voelzmo:
nice work!

coreyti:
Thanks @voelzmo !

PR #1692, opened 2017-05-19
Add support for spec.release.name and spec.release.version to retriev…

barthy1:
…e release info in a job's erb template

Colocated jobs from different releases are also supported

[#119621427](https://www.pivotaltracker.com/story/show/119621427)


src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb
```diff
@@ -23,6 +24,13 @@ def render(spec)
         spec = remove_unused_properties(spec)
       end
 
+      if @release
```
dpb587-pivotal: Can this conditional be removed? As far as I can think, templates should always be affiliated with a particular release.

barthy1: @dpb587-pivotal, thank you for you comment. Actually because of it I found some non-standard behaviour in case of cloud-check related vm recreation. In that case there is no release information so far. Investigating, how I could fix it...

barthy1:
The code was updated. Please, review it, when you have some time

src/spec/gocli/integration/cross_release_collocation_spec.rb
```diff
@@ -22,6 +22,69 @@
...
+      bosh_runner.run("upload-stemcell #{spec_asset('valid_stemcell.tgz')}")
+
+      cloud_config_manifest = yaml_file('cloud_manifest', Bosh::Spec::Deployments.simple_cloud_config)
+      bosh_runner.run("update-cloud-config #{cloud_config_manifest.path}")
+      manifest_hash = Bosh::Spec::Deployments.simple_manifest.merge(manifest_for_properties)
+
+      # We manually change the deployment manifest release version, because of weird issue where
+      # the uploaded release version is `0+dev.1` and the release version in the deployment manifest
+      # is `0.1-dev`, similar to links_spec
+      manifest_hash['releases'][1]['version'] = '0+dev.1'
```
dpb587-pivotal: seems weird and distracting and it should be fixed sometime, but outside of scope of this change; thanks for the reminder

PR #1694, opened 2017-05-24
Add task event log for deleting vm

barthy1:
* show if vm deletion was skipped, because vm does not exist

[#145243901](https://www.pivotaltracker.com/story/show/145243901)


src/bosh-director/lib/bosh/director/jobs/delete_vm.rb
```diff
@@ -17,23 +17,31 @@ def initialize(vm_cid)
...
-        begin
-          instance = Bosh::Director::Api::InstanceLookup.new.by_vm_cid(@vm_cid).first
-          @deployment_name = instance.deployment.name
-          @instance_name = instance.name
-          parent_id = add_event
-          @vm_deleter.delete_for_instance(instance, false)
-        rescue InstanceNotFound
-          parent_id = add_event
-          @vm_deleter.delete_vm_by_cid(@vm_cid)
+        event_log_stage = Config.event_log.begin_stage("Delete vm", 1)
```
cppforlife: let's make it `VM` instead of `vm` in UI text

src/bosh-director/lib/bosh/director/jobs/delete_vm.rb
```diff
@@ -17,23 +17,31 @@ def initialize(vm_cid)
...
-          instance = Bosh::Director::Api::InstanceLookup.new.by_vm_cid(@vm_cid).first
-          @deployment_name = instance.deployment.name
-          @instance_name = instance.name
-          parent_id = add_event
-          @vm_deleter.delete_for_instance(instance, false)
-        rescue InstanceNotFound
-          parent_id = add_event
-          @vm_deleter.delete_vm_by_cid(@vm_cid)
+        event_log_stage = Config.event_log.begin_stage("Delete VM", 1)
+        event_log_stage.advance_and_track(@vm_cid, false) do
```
dpb587-pivotal: Hi! Sorry this PR has sat for so long. We were looking at this today, but had a question about why the `log_exception` change was necessary. It seems odd for a task to ignore an exception and not consider itself failed. If you recall more about this change, can you clarify for us?

The other changes make sense for better logging to UI about what happened or didn't happen. Just confused why task-level behavior needs to change, too.

Thanks!

barthy1: @dpb587-pivotal the main idea, as far as I remember, that in case of vm delete, there are 2 exceptions - VMNotFound and InstanceNotFound, and both of them shouldn't be shown in logs as task failure. However today I haven't been able to recreate this misleading behaviour, so let me update the pr, removing this logic

dpb587-pivotal: @barthy1 - wanted to bump this. Were you still looking at removing this logic. Really appreciate the PR. 👍

PR #1695, opened 2017-05-24
Create new endpoint for getting orphan disks

barthy1:
* new endpoint /orphan_disks
* leave existing logic with endpoint /disks to be compatible with older versions

[#145243179](https://www.pivotaltracker.com/story/show/145243179)


src/bosh-director/lib/bosh/director/api/controllers/orphan_disks_controller.rb
```diff
@@ -0,0 +1,13 @@
+require 'bosh/director/api/controllers/base_controller'
+
+module Bosh::Director
+  module Api::Controllers
+    class OrphanDisksController < BaseController
+      get '/' do
+        content_type(:json)
+        orphan_json = OrphanDiskManager.new(@logger).list_orphan_disks
```
dpb587-pivotal: Traditionally we try to create these managers once in the `initialize`, then in the function invoke the desired method ([example](https://github.com/barthy1/bosh/blob/c85667afea2e1e12b06240542243a3d49f68004a/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb#L43)). Not critical, but helps us keep our `new`ers in a consistent place until we can do better at DI.

src/bosh-director/lib/bosh/director/api/controllers/orphan_disks_controller.rb
```diff
@@ -0,0 +1,13 @@
+require 'bosh/director/api/controllers/base_controller'
+
+module Bosh::Director
+  module Api::Controllers
+    class OrphanDisksController < BaseController
```
dpb587-pivotal: For consistency with the route, can we rename this class/file to be "orphaned".

cppforlife:
i think we ve named all are things "orphan" so route should be "/orphan_disks".

dpb587-pivotal:
I liked how the story title said `GET /orphan_disks` and the description said `GET /orphaned_disks` - I couldn't decide which was correct :)

So I guess it's a route rename, then; not a class/file rename.

PR #1696, opened 2017-05-26
WIP add IPv6 manual networking support

cppforlife:
- switches IpAddress model to store ip addresses as strings
- renamed column to address_str to avoid confusion that address returns int string

Tests running: https://main.bosh-ci.cf-app.com/teams/main/pipelines/ipv6


cppforlife:
closing this one in favor of https://github.com/cloudfoundry/bosh/pull/1769

PR #1705, opened 2017-06-12
Improve HM Heartbeat Reporting

coreyti:
The BOSH Agent [sends heartbeats](https://github.com/cloudfoundry/bosh-agent/blob/master/jobsupervisor/monit_job_supervisor.go#L249-L275) with one of the following states:

- "stopped"
- "starting"
- "running"
- "failing"
- "unknown"

However, the BOSH Health Monitor reports "running" state as `system.healthy = 1` and all other states as `system.healthy = 0`.

This change addresses that as follows:

- In addition to "running", "starting" and "stopped" are now considered as healthy. **NOTE:** it could be argued that "stopped" should be considered as unhealthy. However, it's generally going to be the case that a process on the instance is stopped because a `bosh deploy` is underway, or some other similar and intentional action. For processes that are actually in a bad state, and not down intentionally, those are very likely to be reported as "failing" or "unknown". Thus...
- A `system.unhealthy` metric is added in order to make it easier to monitor for unhealthy states. So, "failing" and "unknown" states will emit `system.unhealthy = 1`, and all others will emit `system.unhealthy = 0`.
- A distinct metric is added for each particular state, making it easier to add detailed monitoring:
  - `system.health.stopped = 1`
  - `system.health.starting = 1`
  - `system.health.running = 1`
  - `system.health.failing = 1`
  - `system.health.unknown = 1`

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/events/heartbeat.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/events/heartbeat.rb
@@ -1,7 +1,7 @@
 module Bosh::Monitor
   module Events
     class Heartbeat < Base
-      HEALTHY_STATES = ['stopped', 'starting', 'running']
+      HEALTHY_STATES = ['running']
 
       attr_reader :agent_id, :deployment, :job, :index, :metrics, :instance_id, :teams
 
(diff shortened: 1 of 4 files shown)


coreyti:
@cppforlife @dpb587 @tylerschultz, updated this PR to restore the existing `system.healthy` definition, and simply add the new `system.health.#{state}` metrics.

jaresty:
merge story [#152973272](https://www.pivotaltracker.com/story/show/152973272)

jfmyers9:
It seems that this change got lost in the backlog madness. I'm going to close and re-open this PR to generate a story.

PR #1717, opened 2017-06-27
Simplify and reduce main-bosh-docker image size

dpb587:
Switch main-bosh-docker to be an independent image (and simplify it a bit).

Avoids many unnecessary layers and packages, reducing to 0.6 GB (from 2.2 GB). If downstream users were relying on packages, they'll want to build their own images using this as the base image.

This significantly reduces the time and disk I/O required by Concourse for spinning up a new container. Unofficial times... on my cheap workers, it reduces start time from ~7 minutes to ~1 minute, and on regular workers, reduces start time from ~1 minute to <15 seconds.

You might care that it uses the [`brightbox/ruby-ng PPA`](https://launchpad.net/~brightbox/+archive/ubuntu/ruby-ng) to install ruby on the system. It seems like a well-known repository and avoids the chruby/rbenv mess since multiple versions aren't needed.

I built and pushed `dpb587/bosh-main-bosh-docker:light` with this code for testing.

Might want to change the CI pipeline.


ci/docker/main-bosh-docker/install.sh
```diff
@@ -0,0 +1,33 @@
...
+  openssh-client \
+  software-properties-common
+
+curl -o /tmp/download https://storage.googleapis.com/golang/go1.8.3.linux-amd64.tar.gz
+echo "1862f4c3d3907e59b04a757cfda0ea7aa9ef39274af99a784f5be843c80c6772 /tmp/download" | sha256sum -c
+
+tar -xzf /tmp/download -C /usr/local --strip-components=1
+rm /tmp/download
+
+apt-add-repository ppa:brightbox/ruby-ng
```
cppforlife: why not `apt-get install ruby`? http://bosh.io/docs/cli-env-deps.html

dpb587-pivotal: ruby apt only had 1.9.3 and i thought it'd be better to use the same as director

dpb587:
Switched to `ubuntu:xenial` and removed `ppa:brightbox/ruby-ng ` and tested a couple things with it and it all seems to work well.

```
# ruby --version
ruby 2.3.1p112 (2016-04-26) [x86_64-linux-gnu]
```

PR #1719, opened 2017-06-27
Gocli bats docs

manno:
After switching our pipeline to gocli, we updated the documentation on running them and moved it to the gocli-bats branch of https://github.com/cloudfoundry/bosh-acceptance-tests/tree/gocli-bats

We could do something similar for master.


dpb587-pivotal:
Thanks for the great docs cleanup!

PR #1730, opened 2017-07-07
Update DNS after updating persistent disks

loewenstein:
- there are CPIs that have to re-create a VM for `attach_disk`
- re-creation changes the IP, hence syncing DNS before leads
  to wrong DNS enties in /etc/hosts

### DIFF
--- a/src/bosh-director/lib/bosh/director/instance_updater.rb
+++ b/src/bosh-director/lib/bosh/director/instance_updater.rb
@@ -6,7 +6,7 @@ module Bosh::Director
 
     def self.new_instance_updater(ip_provider, template_blob_cache, dns_encoder)
       logger = Config.logger
-      disk_manager = DiskManager.new(logger)
+      disk_manager = DiskManager.new(logger, template_blob_cache, dns_encoder)
       agent_broadcaster = AgentBroadcaster.new
       dns_state_updater = DirectorDnsStateUpdater.new(dns_encoder)
       vm_deleter = VmDeleter.new(logger, false, Config.enable_virtual_delete_vms)
@@ -104,8 +104,8 @@ module Bosh::Director
 
         instance_plan.release_obsolete_network_plans(@ip_provider)
 
-        update_dns(instance_plan)
         @disk_manager.update_persistent_disk(instance_plan)
+        update_dns(instance_plan)
 
         unless recreated
           instance.update_instance_settings
--- a/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb
+++ b/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb
@@ -197,7 +197,7 @@ module Bosh::Director
     end
 
     def vm_creator(template_cache, dns_encoder)
-      disk_manager = DiskManager.new(@logger)
+      disk_manager = DiskManager.new(@logger, template_cache, dns_encoder)
       agent_broadcaster = AgentBroadcaster.new
       @vm_creator ||= VmCreator.new(@logger, vm_deleter, disk_manager, template_cache, dns_encoder, agent_broadcaster)
     end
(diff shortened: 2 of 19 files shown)


cppforlife:
this one deserves a test since its easy to regress.

loewenstein:
Agreed. How would that be done? 
Unit testing the timing between two separate ruby objects (updater and disk_manager)... Is there support in rspec mocks?

src/bosh-director/lib/bosh/director/instance_updater.rb
```diff
@@ -104,8 +104,8 @@ def update(instance_plan, options = {})
 
         release_obsolete_ips(instance_plan)
 
-        update_dns(instance_plan)
         @disk_manager.update_persistent_disk(instance_plan)
+        update_dns(instance_plan)
```
dpb587-pivotal: Possible to test a scenario where persistent disk updates fail, and then ensure the "latest" DNS version is the version prior to the update?

loewenstein:
@cppforlife @dpb587 This is not yet ready to be merged.

I reworked DiskManager to get the DB up to date prior to sending out DNS blobs.
The change does solve the problem for the k8s CPI.
However, there are a couple of open points:
- Write a unit test for the new behaviour (all units are green, just no assertion on the new things happening)
- Fix the failing integration tests
```
rspec ./spec/gocli/integration/cli_ignore_spec.rb:387 # ignore/unignore-instance when there are ignored instances and a deploy happens when the existing instances is larger than the desired ones when the ignored instances is equal to desired ones deletes all non-ignored vms and leaves the ignored alone without updating them
rspec ./spec/gocli/integration/cli_ignore_spec.rb:440 # ignore/unignore-instance when there are ignored instances and a deploy happens when the existing instances is larger than the desired ones when the ignored instances are fewer than the desired ones should keep the ignored instances untouched and adjust the number of remaining functional instances
rspec ./spec/gocli/integration/cli_ignore_spec.rb:957 # ignore/unignore-instance when ignoring all of the instances in a zone when using static IPs balances the vms correctly, and it errors if removing an az containing ignored vms, and it errors if removing static IP assigned to an ignored VM
```
- more places that have to update DB & broadcast DNS

loewenstein:
@dpb587-pivotal @cppforlife 
I would evolve this PR to
- concentrate all disk operations inside DiskManager, i.e. nowhere should `cpi.attach_disk` / `cpi.detach_disk` be called directly
- conditionally handle the "changing IP" behaviour inside DiskManager by
  - wait for agent to start
  - apply_initial_vm_spec -> get_state -> update DB spec with 'networks' from agent state, render templates (1)
  - broadcast dns (2)

I propose to have explicit director configuration for when/if this actions should be triggered. For certain CPIs (for now I only know of one) this will always be required, for others it will never be required, e.g. `update_dns_on_disk_operation`.

This will involve creating a deployment plan in places where this was not yet necessary and might lead to parts of these operations called multiple times (e.g. instance updater calls disk_manager.update_persistent_disks, which will broadcast dns, and itself broadcasts dns).

WDYT?

1)
```
def apply_initial_vm_state(instance_plan)
       instance_plan.instance.apply_initial_vm_state(instance_plan.spec)
 
       unless instance_plan.instance.compilation?
         # re-render job templates with updated dynamic network settings
         @logger.debug("Re-rendering templates with updated dynamic networks: #{instance_plan.spec.as_template_spec['networks']}")
         JobRenderer.render_job_instances_with_cache([instance_plan], @template_blob_cache, @dns_encoder, @logger)
       end
     end
```

2)
```
@dns_state_updater.update_dns_for_instance(instance_plan.instance.model, instance_plan.network_settings.dns_record_info)
```

jaresty:
merge story [#152973240](https://www.pivotaltracker.com/story/show/152973240)

jfmyers9:
Sorry that we never got to merging or reviewing this PR. At this point it seems that it is quite out of date.

I'm going to close this PR. Please feel free to re-submit this PR if you would still like us to consider the change.

Best,

@jfmyers9, CF BOSH

PR #1737, opened 2017-07-19
Extend integration test to verify that after delete-vm command resurr…

barthy1:
…ector recreates vm

[#149000461](https://www.pivotaltracker.com/story/show/149000461)

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/plugins/resurrector_helper.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/plugins/resurrector_helper.rb
@@ -57,7 +57,9 @@ module Bosh::Monitor::Plugins
       end
 
       def state_for(deployment)
-        agents = @instance_manager.get_agents_for_deployment(deployment)
+        #do not forget about instances with deleted vm, which expect to have vm
+        agents = @instance_manager.get_agents_for_deployment(deployment).values +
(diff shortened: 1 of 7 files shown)


cppforlife:
@barthy1 so we are saying that it works? ive repro-ed this on 262.x version so i am a bit surprised that it wouldnt be a problem on master.

barthy1:
@cppforlife found use case to recreate the error - there should not be any "healthy" instances in the deployment and it prevented resurrector to send request to scan_and_fix. I added commit to fix it

src/bosh-monitor/lib/bosh/monitor/plugins/resurrector_helper.rb
```diff
@@ -111,7 +113,7 @@ def normal?
       end
 
       def summary
-        "deployment: '#{@eployment}'; #{@unhealthy_count} of #{@agent_count} agents are unhealthy (#{(unhealthy_percent * 100).round(1)}%)"
+        "deployment: '#{@deployment}'; #{@unhealthy_count} of #{@agent_count} agents are unhealthy (#{(unhealthy_percent * 100).round(1)}%)"
```
dpb587-pivotal: good catch; odd that tests were incorrectly written

PR #1738, opened 2017-07-19
Fix documentation for running test

Kiemes:
As the Gemfile is located in src directory, the old commands
did not work anymore.


dpb587-pivotal:
Thanks!

PR #1739, opened 2017-07-20
Fix inconsistencies in disk and vm metadata

Kiemes:
Use Manifest v2 Schema naming for instance group.
Keep v1 schema tags in VM for backwards compatibility.

[#149144325](https://www.pivotaltracker.com/story/show/149144325)

### DIFF
--- a/src/bosh-director/lib/bosh/director/metadata_updater.rb
+++ b/src/bosh-director/lib/bosh/director/metadata_updater.rb
@@ -19,6 +19,7 @@ module Bosh::Director
         metadata['deployment'] = instance.deployment.name
         metadata['id'] = instance.uuid
         metadata['job'] = instance.job
+        metadata['instance_group'] = instance.job
         metadata['index'] = instance.index.to_s
         metadata['name'] = "#{instance.job}/#{instance.uuid}"
         metadata['created_at'] = Time.new.getutc.strftime('%Y-%m-%dT%H:%M:%SZ')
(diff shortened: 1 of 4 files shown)


src/bosh-director/lib/bosh/director/metadata_updater.rb
```diff
@@ -19,6 +19,7 @@ def update_vm_metadata(instance, metadata, factory = CloudFactory.create_with_la
         metadata['deployment'] = instance.deployment.name
         metadata['id'] = instance.uuid
```
dpb587-pivotal: should we be adding `instance_id` in this round of consistency fixing, too?

Kiemes: @dpb587-pivotal from the line you are commenting on I guess you ask for `instance_id` in `update_VM_metadata`? We explicitly left that tag out as its value is in the metadata tag `id`. As we are within the context of the VM/instance, I think we should not "namespace" `id` with `instance`. We are using `instance_id` in DISK metadata as we are in the context of the disk and we need to namespace the id of the instance.

Now you could argue why we used `instance_group` in the context of VM metadata and not just `group`. We actually were not sure there. `instance_group` is what people know from the manifest. I guess they would wonder about what `group` would be.

If the outcome of the discussion will be we need to add `instance` namespace here, then this would also mean `instance_name` and `instance_index`. But I hope we do not need the namespace.

Kiemes:
@dpb587-pivotal @cppforlife any further comments?

src/bosh-director/lib/bosh/director/metadata_updater.rb
dpb587-pivotal: Well reasoned. Personally I consider tags to be global and shared across resources if they have the same meaning/context, so I generally lean towards consistency between the two resources. Optimistically, I think these tags would be namespaced with `bosh.` or something as well, but I'm not holding my breath for that idea :)

But you're right, these tags are definitely within the context of an instance, so they could theoretically be just `group`. Given current conventions,`group` might be more appropriate. I defer to @cppforlife.

PR #1762, opened 2017-08-15
show locking task id in the acquire lock error

barthy1:
[#148999523](https://www.pivotaltracker.com/story/show/148999523)

### DIFF
--- a/src/bosh-director/lib/bosh/director/lock.rb
+++ b/src/bosh-director/lib/bosh/director/lock.rb
@@ -129,7 +129,11 @@ module Bosh::Director
         end
 
         unless acquired
-          raise TimeoutError, "Failed to acquire lock for #{@name} uid: #{@uid}" if Time.now - started > @timeout
+          if Time.now - started > @timeout
+            lock = current_lock
+            task_id = lock ? lock.task_id : "expired"
+            raise TimeoutError, "Failed to acquire lock for #{@name} uid: #{@uid}. Locking task id is #{task_id}"
+          end
           sleep(0.5)
           lock_expiration = Time.now.to_f + @expiration + 1
         end
@@ -150,7 +154,6 @@ module Bosh::Director
       )
     end
 
-
     def delete
       if Models::Lock.where(name: @name, uid: @uid).delete > 0
         @logger.debug("Deleted lock: #{@name} uid: #{@uid}")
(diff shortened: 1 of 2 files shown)


src/bosh-director/lib/bosh/director/lock.rb
```diff
@@ -129,7 +129,11 @@ def acquire
         end
 
         unless acquired
-          raise TimeoutError, "Failed to acquire lock for #{@name} uid: #{@uid}" if Time.now - started > @timeout
+          if Time.now - started > @timeout
+            lock = current_lock
+            task_id = lock ? lock.task_id : "expired"
```
dpb587-pivotal: Perhaps we can say `(gone)` instead of `expired`. Parens to specifically call out there's something special with the value, and "gone" instead of expired since expired typically means a lock was lost (rather than intentionally dropped). Wondering if people will read it and come to us saying "my lock expired but the deployment still failed". Just a though; whatever you think.

src/bosh-director/spec/unit/lock_spec.rb
```diff
@@ -47,26 +47,38 @@ module Bosh::Director
...
     end
 
+    it 'should return info about expired task' do
+      lock_a = Lock.new('foo', )
+      lock_b = Lock.new('foo', timeout: 0.1)
+
+      lock_a.lock do
+        allow(lock_b).to receive(:current_lock).and_return(nil)
+        expect do
+          lock_b.lock {  something }
```
dpb587-pivotal: Extra space :)

src/bosh-director/spec/unit/lock_spec.rb
```diff
@@ -47,26 +47,38 @@ module Bosh::Director
...
         ran_once = true
-        expect { lock_b.lock {} }.to raise_exception(Lock::TimeoutError)
+        expect { lock_b.lock {} }.to raise_exception(Lock::TimeoutError, /Failed to acquire lock for foo uid: .* Locking task id is 1/)
       end
 
       expect(ran_once).to be(true)
     end
 
+    it 'should return info about expired task' do
+      lock_a = Lock.new('foo', )
```
dpb587-pivotal: Unused second argument

barthy1:
@dpb587-pivotal thank you for catching that

PR #1763, opened 2017-08-21
allow filtering of addons by team

barthy1:
if addon is specified on deployment level - no actions

[#150158345](https://www.pivotaltracker.com/story/show/150158345)

### DIFF
--- a/src/bosh-director/lib/bosh/director/addon/filter.rb
+++ b/src/bosh-director/lib/bosh/director/addon/filter.rb
@@ -4,11 +4,12 @@ module Bosh::Director
 
       extend ValidationHelper
 
-      def initialize(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, filter_type)
+      def initialize(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, applicable_teams, filter_type)
         @applicable_jobs = applicable_jobs
         @applicable_deployment_names = applicable_deployment_names
         @applicable_stemcells = applicable_stemcells
         @applicable_networks = applicable_networks
+        @applicable_teams = applicable_teams
         @filter_type = filter_type
       end
 
@@ -20,16 +21,24 @@ module Bosh::Director
         applicable_jobs = safe_property(addon_filter_hash, 'jobs', :class => Array, :default => [])
         applicable_stemcells = safe_property(addon_filter_hash, 'stemcell', :class => Array, :default => [])
         applicable_networks = safe_property(addon_filter_hash, 'networks', :class => Array, :default => [])
+        applicable_teams = safe_property(addon_filter_hash, 'teams', :class => Array, :default => [])
+        applicable_teams = [] if addon_level == DEPLOYMENT_LEVEL
 
         #TODO throw an exception with all wrong jobs
         verify_jobs_section(applicable_jobs, filter_type, addon_level)
 
         verify_stemcells_section(applicable_stemcells, filter_type, addon_level)
 
-        new(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, filter_type)
+        new(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, applicable_teams, filter_type)
       end
 
-      def applies?(deployment_name, deployment_instance_group)
+      def applies?(deployment, deployment_instance_group)
+        deployment_name = deployment ? deployment.name : nil
+        deployment_teams = deployment ? deployment.model.teams.map(&:name) : nil
+
+        if has_teams? && !has_applicable_team?(deployment_teams)
+          return false
+        end
         if has_stemcells? && !has_applicable_stemcell?(deployment_instance_group)
           return false
         end
@@ -45,9 +54,10 @@ module Bosh::Director
             return @applicable_deployment_names.include?(deployment_name) && has_applicable_job?(deployment_instance_group)
           else
             return true if @filter_type == :include
-            # cases with `has_stemcells? && !has_applicable_stemcell?` and `has_networks? && !has_applicable_network?` are checked before.
-            # all other cases are covered by simple check `has_stemcells? || has_networks?`.
-            return @filter_type == :exclude && (has_stemcells? || has_networks?)
+            # cases with `has_stemcells? && !has_applicable_stemcell?`, `has_networks? && !has_applicable_network?` and
+            # `has_team? && !has_applicable_team?` are checked before.
+            # all other cases are covered by simple check `has_stemcells? || has_networks? || has_teams?`.
+            return @filter_type == :exclude && (has_stemcells? || has_networks? || has_teams?)
         end
       end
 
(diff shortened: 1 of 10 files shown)


src/bosh-director/lib/bosh/director/addon/filter.rb
```diff
@@ -20,16 +21,24 @@ def self.parse(addon_filter_hash, filter_type, addon_level =  RUNTIME_LEVEL)
...
         verify_jobs_section(applicable_jobs, filter_type, addon_level)
 
         verify_stemcells_section(applicable_stemcells, filter_type, addon_level)
 
-        new(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, filter_type)
+        new(applicable_jobs, applicable_deployment_names, applicable_stemcells, applicable_networks, applicable_teams, filter_type)
       end
 
-      def applies?(deployment_name, deployment_instance_group)
+      def applies?(deployment, deployment_instance_group)
```
dpb587-pivotal: Hand the whole `deployment` class down to `applies?` feels a bit heavy. Could we potentially just pass the teams from the model along with the `deployment_name`?

src/bosh-director/lib/bosh/director/addon/filter.rb
```diff
@@ -20,16 +21,24 @@ def self.parse(addon_filter_hash, filter_type, addon_level =  RUNTIME_LEVEL)
...
+        deployment_name = deployment ? deployment.name : nil
+        deployment_teams = deployment ? deployment.model.teams.map(&:name) : nil
```
dpb587-pivotal: If we were to pass an array of team names to `applies?` it would be great if deployment had a method that would return said array of team names so that the caller does not have to understand or reach into the model and make a map itself.

src/bosh-director/lib/bosh/director/addon/filter.rb
```diff
@@ -106,6 +116,15 @@ def has_applicable_network?(deployment_instance_group)
           deployment_instance_group.has_network?(network_name)
         end
       end
+
+      def has_teams?
+        !@applicable_teams.nil? && !@applicable_teams.empty?
+      end
+
+      def has_applicable_team?(deployment_teams)
```
dpb587-pivotal: the `has_teams?`and `has_applicable_team?` methods look to be mirroring the stemcell implementation which introduces a bit of necessary complexity. Could we remove the extra calls to `has_teams?` that appear in `has_applicable_team?`.

src/bosh-director/spec/unit/addon/filter_spec.rb
```diff
@@ -66,32 +70,51 @@ module Addon
...
-              expect(addon_include.applies?('deployment_blarg', nil)).to be(false)
+              expect(addon_include.applies?(deployment, nil)).to be(false)
             end
           end
         end
 
+        describe 'when the filter spec has only a teams section' do
+          let(:filter_hash) { {'teams' => ['team_1', 'team_2']} }
+
+          describe 'when the one of the teams matches one from the include spec' do
```
dpb587-pivotal: grammar should be "when one of the teams matches one from the included spec".

src/bosh-director/spec/unit/addon/filter_spec.rb
```diff
@@ -66,32 +70,51 @@ module Addon
...
+            it 'applies' do
+              expect(addon_include.applies?(deployment, nil)).to be(true)
+            end
+          end
+
+          describe 'when the none of the teams does not match one from the filter spec' do
```
dpb587-pivotal: grammar correction here - "when none of the teams match the filter spec"

src/bosh-director/spec/unit/addon/filter_spec.rb
```diff
@@ -213,7 +238,16 @@ module Addon
 
             it 'applies' do
               expect(addon_include.applies?(nil, nil)).to be(true)
-              expect(addon_include.applies?('anything', instance_group)).to be(true)
+              expect(addon_include.applies?(deployment, instance_group)).to be(true)
+            end
+          end
+
+          describe 'when addon is in deployment manifest and has team filter' do
```
dpb587-pivotal: Could we add a context here for the `when clause`. Code right now is a bad example but we are going to start to break these whens out.

src/bosh-director/spec/unit/addon/filter_spec.rb
```diff
@@ -213,7 +238,16 @@ module Addon
...
+            let(:addon_level) { DEPLOYMENT_LEVEL }
+            let(:filter_hash) { {'teams' => ['team_5']} }
+
+            it 'does not do anything' do
```
dpb587-pivotal: Is this actually correct? The spec says it does not apply but our expectation says that it should apply. One of these two is wrong.

src/bosh-director/spec/unit/runtime_config/runtime_filter_spec.rb
```diff
@@ -5,7 +5,14 @@ module Bosh::Director
 
     subject(:addon_filter) { Addon::Filter.parse(filter_spec, filter_type) }
     let(:deployment_name) { 'dep1' }
-    let(:deployment_model) { Models::Deployment.make(name: deployment_name) }
+    let(:deployment_model) do
+      deployment_model =Models::Deployment.make(name: deployment_name)
```
dpb587-pivotal: spacing of `=`

src/spec/gocli/integration/deploy_job_with_addons_spec.rb
```diff
@@ -97,6 +97,51 @@
       expect(File.exist?(no_addon_instance.job_path('dummy'))).to eq(false)
     end
 
+    context 'when team is specified' do
+      with_reset_sandbox_before_each(user_authentication: 'uaa')
+
+      it 'allows addons to be added for specific team' do
```
dpb587-pivotal: `for the specificed team`

src/spec/gocli/integration/deploy_job_with_addons_spec.rb
```diff
@@ -97,6 +97,51 @@
       expect(File.exist?(no_addon_instance.job_path('dummy'))).to eq(false)
     end
 
+    context 'when team is specified' do
```
dpb587-pivotal: `when a team is specified`

src/bosh-director/lib/bosh/director/addon/filter.rb
barthy1: I believe each method should check the input it used to avoid problems with nil array. has_applicable_team? method uses @applicable_teams array, so I think it's better to have the check

src/spec/gocli/integration/deploy_job_with_addons_spec.rb
barthy1: I think it's `for the specified team`

barthy1:
@dpb587-pivotal thank you for your comments. Could you please review the pr again?

GarfieldIsAPhilosopher:
Fixed the has_applicable_team? logic, and merged it to master

dpb587-pivotal:
I don't think this was merged correctly, but it looks like these changes made it to master. Will follow up to figure out what may have happened in the local merge.

Thanks!

PR #1766, opened 2017-08-23
Json plugin manages processes

wfernandes:
If a json plugin compatible process dies, the json plugin will restart it. Also, jobs appearing in the manifest after the health_monitor will be detected and started.

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/plugins/json.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/plugins/json.rb
@@ -1,29 +1,106 @@
+require 'thread'
+
 module Bosh::Monitor::Plugins
+
   class Json < Base
-    attr_reader :processes
+    attr_reader :process_manager
+
+    def initialize(options = {})
+      super(options)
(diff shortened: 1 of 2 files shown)


joerodriguez:
@cppforlife This is the work that you had suggested during our conversation at CF Summit.

wfernandes:
@cppforlife Just wanted an update on this PR. We'd like to get this in as soon as possible as it is vital for the smooth operation of the bosh-system-metrics-server. Thanks.

zachgersh:
@wfernandes would it be possible to write an integration test for this? Would be really nice if we could go build a plugin that would be started / receive and maybe echo the events?

Just a quick glance first idea.

jtarchie:
@zachgersh, we'd love to. We've had some troubles with testing multiple loops with EventMachine (ie we don't know how to). Is anyone available to pair on it next week?

cppforlife:
@jtarchie 

> Is anyone available to pair on it next week?
yup, reach out to us next week and we'll pair

tjvman:
Confirmed that new behavior works with manual testing; integration and unit tests are passing locally.

PR #1768, opened 2017-08-24
user can bosh ssh by ip

barthy1:
- search in Models::IpAddress, if nothing was found - in the Models::Instance.spec_json

[#149386633](https://www.pivotaltracker.com/story/show/149386633)

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/instance_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/instance_manager.rb
@@ -30,7 +30,8 @@ module Bosh::Director
       # @return [Array] List of instances that matched the filter
       # @raise [InstanceNotFound]
       def filter_by(deployment, filter)
-        InstanceLookup.new.by_filter(filter.merge(deployment_id: deployment.id))
+        return InstanceLookup.new.by_filter(filter.merge(deployment_id: deployment.id)) if filter.is_a?(Hash)
+        return InstanceLookup.new.by_filter(filter << { deployment_id: deployment.id }) if filter.is_a?(Array)
       end
 
       # @param [Models::Instance] instance Instance
--- a/src/bosh-director/lib/bosh/director/api/instance_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/instance_manager.rb
@@ -30,8 +30,10 @@ module Bosh::Director
       # @return [Array] List of instances that matched the filter
       # @raise [InstanceNotFound]
       def filter_by(deployment, filter)
-        return InstanceLookup.new.by_filter(filter.merge(deployment_id: deployment.id)) if filter.is_a?(Hash)
-        return InstanceLookup.new.by_filter(filter << { deployment_id: deployment.id }) if filter.is_a?(Array)
+        if filter.is_a?(Hash)
+          filter = filter.empty? ? [] : [filter]
+        end
+        InstanceLookup.new.by_filter(filter << { deployment_id: deployment.id })
       end
 
       # @param [Models::Instance] instance Instance
--- a/src/bosh-director/lib/bosh/director/api/instance_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/instance_manager.rb
@@ -30,10 +30,7 @@ module Bosh::Director
       # @return [Array] List of instances that matched the filter
       # @raise [InstanceNotFound]
       def filter_by(deployment, filter)
-        if filter.is_a?(Hash)
-          filter = filter.empty? ? [] : [filter]
-        end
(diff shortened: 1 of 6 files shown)


src/bosh-director/lib/bosh/director/api/instance_manager.rb
```diff
@@ -30,7 +30,8 @@ def find_instances_by_deployment(deployment)
       # @return [Array] List of instances that matched the filter
       # @raise [InstanceNotFound]
       def filter_by(deployment, filter)
-        InstanceLookup.new.by_filter(filter.merge(deployment_id: deployment.id))
+        return InstanceLookup.new.by_filter(filter.merge(deployment_id: deployment.id)) if filter.is_a?(Hash)
```
dpb587-pivotal: I'm not sure of Sequel syntax, but instead of this dual-call thing which looks a little confusing and fragile, do you know if it's possible to convert filter hashes to arrays with something like the following? Mainly trying to think how to only have a single `InstanceLookup.new.by_filter` and `deployment_id` filter.

    filter = [filter] if filter.is_a?(Hash)

src/bosh-director/lib/bosh/director/jobs/ssh.rb
```diff
@@ -21,17 +22,32 @@ def initialize(deployment_id, options = {})
 
       def perform
         target = Target.new(@target_payload)
-
         filter = {}
-        filter[:job] = target.job if target.job
-        filter.merge!(target.id_filter)
+        if target.job && ip_address?(target.job)
```
dpb587-pivotal: Theoretically, I suppose there could be ambiguity on whether it's a job or an IP address. Is it worth the complexity of only falling back to IP address if job name did not match? Probably not, but I wondered if you had considered the possibility.

barthy1: @dpb587-pivotal at this moment you cannot specify job name as something like '1.1.1.1', because of error `Invalid DNS canonical name '...', must begin with a letter`. At least that's what I had. But you are correct, it's better to cover this possibility for the future possible updates.

src/bosh-director/lib/bosh/director/jobs/ssh.rb
```diff
@@ -21,17 +22,32 @@ def initialize(deployment_id, options = {})
...
-        filter[:job] = target.job if target.job
-        filter.merge!(target.id_filter)
+        if target.job && ip_address?(target.job) && instance_count(target.job) == 0
+          ip_address = target.job
+          net_address = ip_to_netaddr(ip_address)
+          ip_error_message = "{:ip=>#{ip_address}}"
+          ipaddress_model = Models::IpAddress.where(address: net_address.to_i).first
+          if ipaddress_model.nil?
+            filter = []
+            filter << Sequel.like(:spec_json, "%\\\"ip\\\":\\\"#{ip_address}\\\"%")
```
jraqula: Hi Barthy1

We wouldn't expect to be aware of Sequel interface at this point -- can we move this parameter into a hash component in the call to `filter_by`, and rely on the InstanceLookup to implement the `spec_json` search in DB? This seems like a high level of exposure of the DB layout/contract.

what do you think?
eve

jraqula: Additionally, it looks like L#28-30 can be pushed inside the instance lookup also?

src/bosh-director/lib/bosh/director/jobs/ssh.rb
```diff
@@ -21,17 +22,32 @@ def initialize(deployment_id, options = {})
...
+          filter[:job] = target.job if target.job
+          filter.merge!(target.id_filter)
+        end
 
         deployment = Models::Deployment[@deployment_id]
-
-        instances = @instance_manager.filter_by(deployment, filter).reject { |i| i.active_vm.nil? }
+        begin
+          instances = @instance_manager.filter_by(deployment, filter).reject { |i| i.active_vm.nil? }
+        rescue InstanceNotFound
```
jraqula: Following above comment: if we isolate the abstraction as described there, wouldn't we be able to have this error be swallowed neatly, so that we don't need to start handling this case.

barthy1:
@jraqula thank you for your ideas, I think they are very reasonable. I moved the logiс into InstanceLookup, and now, I believe, code looks clearer

jaresty:
merge story [#152973098](https://www.pivotaltracker.com/story/show/152973098)

jfmyers9:
This PR seems to have gotten lost in our many backlogs. I'm going to close and reopen this PR to generate a tracker story.

PR #1784, opened 2017-09-18
Log error messages reported in body of responses from CredHub

dannysullivan:
The code was reporting the value of the "message" attribute of the HTTP response from CredHub, but the detailed error message is contained in the body, under the key "error."

[#150933164] Submit PR to fix the lack of descriptive error in BOSH generation failure message

Signed-off-by: Edie Beer <ebeer@pivotal.io>


dpb587-pivotal:
This is a Pivotal PR; mergers feel free to ignore @cfdreddbot in this case.

@dannysullivan, might want to publish one of your orgs to keep @cfdreddbot happy in the future.

dannysullivan:
@dpb587-pivotal thanks for the heads up - I'll do that now.

Issue #1525, opened 2016-11-30
Support source code diffs between deployments


petergtz:
This issue is a proposal for discussion rather than a concrete feature request:

Sometimes it is incredibly helpful to troubleshoot an issue by knowing what has changed since the last deployment. While it is kind of possible to track down such changes with the information provided by BOSH, it's difficult and error prone.

I can imagine different scenarios what a real use case could look like.

**Proposal 1:**
```
$ bosh diff deployment 2345 2346
Deployment changes
----------------------------
Releases
etcd
  version type changed: String -> Fixnum
    - 86
    + 86

Properties
cc
  bits_service
    + enabled: true
    + password: blahblahblahbblah
    + private_endpoint: http://bits-service.service.cf.internal
    + public_endpoint: http://bits-service.cf1.bits.cf-app.com
    + username: admin


Release changes
------------------

bits-service 0.0.234 -> 0.0.252:
https://github.com/cloudfoundry-incubator/bits-service-release/compare/1a59a892f42f946d7181a738c84e69af786c9f5f...b88a79d26c9df7070ed6e305a6b06fb056eafef3

```

**Proposal 2:**

```
$ bosh diff deployment 2345 2346
Deployment changes
----------------------------
///// snip. same as above

Release changes
--------------------

bits-service 0.0.234 -> 0.0.252:
jobs/bits-service/spec
@@ -157,15 +157,27 @@ properties:
bits-service.nginx.access_log_destination:
description: "Nginx access log destination. Can be used to route access logs to a file, syslog, or a memory buffer."
- default: "/var/vcap/sys/log/nginx_bits/nginx.access.log"
+ default: "/var/vcap/sys/log/nginx_bits/access.log"
+ bits-service.nginx.public_access_log_destination:
+ description: "Nginx access log destination. Can be used to route access logs to a file, syslog, or a memory buffer."
+ default: "/var/vcap/sys/log/nginx_bits/public_access.log"
+ bits-service.nginx.private_access_log_destination:
+ description: "Nginx access log destination. Can be used to route access logs to a file, syslog, or a memory buffer."
+ default: "/var/vcap/sys/log/nginx_bits/private_access.log"
```

**Thoughts on implementation**

To make this possible, a few things would be necessary in BOSH:
1. Don't just capture a commit sha in a BOSH release, but also the source it was generated from, e.g. the URL to a github repo. This could simply be an optional parameter to `bosh create release` or it could be deduced automatically from git remote.
[...]

PR #1788, opened 2017-09-19
Add new feature: deployment revisions

petergtz:
Addresses https://github.com/cloudfoundry/bosh/issues/1525 (which contains much more context and discussions with @cppforlife). Here's what this feature allows (taken partially from the issue):

## Some examples of its usage

**Get the history of a deployment including deployment manifests at any point in time:**
```bash
curl -k 'https://test:test@localhost:61004/deployments/simple/history' | jq
```
```yaml
[
  {
    "deployment_name": "simple",
    "revision_number": 4,
    "user": "test",
    "task": "7",
    "started_at": "2017-07-17 23:18:59 +0200",
    "completed_at": "2017-07-17 21:19:22 UTC",
    "error": null,
    "manifest_text": "---\nname: simple\ndirector_uuid: deadbeef\nreleases:\n- name: bosh-release\n  version: latest\nupdate:\n  canaries: 2\n  canary_watch_time: 4000\n  max_in_flight: 1\n  update_watch_time: 20\njobs:\n- name: foobar\n  templates:\n  - name: foobar\n  resource_pool: a\n  instances: 3\n  networks:\n  - name: a\n  properties: {}\n  persistent_disk_pool: disk_a\n"
  },
  {
    "deployment_name": "simple",
    "revision_number": 3,
    "user": "test",
    "task": "5",
    "started_at": "2017-07-17 23:18:44 +0200",
    "completed_at": "2017-07-17 21:18:53 UTC",
    "error": null,
    "manifest_text": "---\nname: simple\ndirector_uuid: deadbeef\nreleases:\n- name: bosh-release\n  version: 0.1-dev\nupdate:\n  canaries: 2\n  canary_watch_time: 4000\n  max_in_flight: 1\n  update_watch_time: 20\njobs:\n- name: foobar\n  templates:\n  - name: foobar\n  resource_pool: a\n  instances: 3\n  networks:\n  - name: a\n  properties: {}\n  persistent_disk_pool: disk_a\n"
  },
  {
    "deployment_name": "simple",
    "revision_number": 2,
    "user": "test",
    "task": "4",
    "started_at": "2017-07-17 23:18:32 +0200",
    "completed_at": "2017-07-17 21:18:41 UTC",
    "error": null,
    "manifest_text": "---\nname: simple\ndirector_uuid: deadbeef\nreleases:\n- name: bosh-release\n  version: 0.1-dev\nupdate:\n  canaries: 2\n  canary_watch_time: 4000\n  max_in_flight: 1\n  update_watch_time: 20\njobs:\n- name: foobar\n  templates:\n  - name: foobar\n  resource_pool: a\n  instances: 2\n  networks:\n  - name: a\n  properties: {}\n  persistent_disk_pool: disk_a\n"
  },
  {
    "deployment_name": "simple",
    "revision_number": 1,
    "user": "test",
    "task": "3",
    "started_at": "2017-07-17 23:18:18 +0200",
    "completed_at": "2017-07-17 21:18:28 UTC",
    "error": null,
    "manifest_text": "---\nname: simple\ndirector_uuid: deadbeef\nreleases:\n- name: bosh-release\n  version: 0.1-dev\nupdate:\n  canaries: 2\n  canary_watch_time: 4000\n  max_in_flight: 1\n  update_watch_time: 20\njobs:\n- name: foobar\n  templates:\n  - name: foobar\n  resource_pool: a\n  instances: 1\n  networks:\n  - name: a\n  properties: {}\n  persistent_disk_pool: disk_a\n"
  }
]
```

**Get the differences of manifests and releases for one deployment between two points in time:**

```bash
curl -k 'https://test:test@localhost:61004/deployments/simple/diff_revisions?revision1=3&revision2=4' | jq
```

```yaml
{
  "manifest": [
    [
      "releases:",
      null
    ],
    [
      "- name: bosh-release",
      null
    ],
    [
      "  version: 0.1-dev",
      "removed"
    ],
    [
      "  version: latest",
      "added"
    ]
  ],
  "releases": {
    "added": [
      "bosh-release/0+dev.2"
    ],
    "removed": [
      "bosh-release/0.1-dev"
    ]
  }
}

```
## A few notes on the implementation
* I debated long with myself if I should call the new entity `deployment_revision` or `audit_log_entry` (and similarly `history` versus `audit_log`). The latter would emphasize usefulness in contexts where precise audit_logs of production systems are required. OTOH, it's really different revisions of the same deployment. Suggestions welcome.
* I also debated with myself, if I should invent a completely new event type, or if I should simply formalize the existing deployment event. In the end, I decided for a new event type, as I feel like this is a first class concept, which I do not necessarily want to mix with the existing event structure.
* Using `object_name` as revision number is a bit hacky and I'm not super happy with it.
* The code comes only with a partially complete suite of tests just yet, but I can of course add them, should we decide to go forward with this.
* I have also not added CLI commands. Also waiting for feedback first.

### DIFF
--- a/src/bosh-director/lib/bosh/director/api.rb
+++ b/src/bosh-director/lib/bosh/director/api.rb
@@ -8,6 +8,7 @@ require 'bosh/director/api/http_constants'
 require 'bosh/director/api/api_helper'
 
 require 'bosh/director/api/deployment_manager'
+require 'bosh/director/api/revision_manager'
 require 'bosh/director/api/instance_manager'
 require 'bosh/director/api/problem_manager'
 require 'bosh/director/api/property_manager'
--- a/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
@@ -48,6 +48,7 @@ module Bosh::Director
         @instance_manager = Api::InstanceManager.new
         @deployments_repo = DeploymentPlan::DeploymentRepo.new
         @instance_ignore_manager = Api::InstanceIgnoreManager.new
+        @revision_manager = Api::RevisionManager.new
       end
 
       get '/:deployment/jobs/:job/:index_or_id' do
@@ -143,6 +144,21 @@ module Bosh::Director
 
       get '/:deployment/snapshots' do
         json_encode(@snapshot_manager.snapshots(deployment))
+      end
+
+      get '/:deployment/history' do
+        json_encode(@revision_manager.revisions(deployment.name,
+          include_manifest: params['include_manifest'],
+          include_cloud_config: params['include_cloud_config'],
+          include_runtime_configs: params['include_runtime_configs']))
+      end
+
+      get '/:deployment/history/:revision' do
+        json_encode(@revision_manager.revision(deployment.name, params['revision']))
+      end
+
+      get '/:deployment/diff_revisions' do
+        json_encode(@revision_manager.diff(deployment, params['revision1'], params['revision2'], should_redact: !params.fetch(:no_redact, false)))
       end
 
       get '/:deployment/jobs/:job/:index/snapshots' do
--- a/src/bosh-director/lib/bosh/director/api/revision_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/revision_manager.rb
@@ -1,0 +1,116 @@
+module Bosh::Director
+  module Api
+    class RevisionManager
+      include ApiHelper
+
+      def initialize
+        @event_manager = EventManager.new(true)
+      end
+
+      def create_revision(deployment_name:, user:, task:, started_at:, manifest_text:, cloud_config_id:, runtime_config_ids:, releases:, stemcells:, error: nil)
+        Models::Event.create(
+          timestamp:   Time.now,
+          user:        user,
+          task:        task,
+          action:      "create",
+          object_type: "deployment_revision",
+          object_name: (last_revision_number(deployment_name)+1).to_s,
+          deployment:  deployment_name,
+          context: {
+            manifest_text: manifest_text,
+            started_at: started_at,
+            cloud_config_id: cloud_config_id,
+            runtime_config_ids: runtime_config_ids,
+            releases: releases,
+            stemcells: stemcells,
+          },
+          error: error,
+          )
+      end
+
+      def last_revision_number(deployment_name)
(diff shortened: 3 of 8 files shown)


cppforlife:
@petergtz sorry it's taken some time to respond. ive been thinking a bit more about this problem recently and have thought about an interesting feature combination. @voelzmo and team is adding support for generic configs (https://github.com/cloudfoundry/bosh-notes/blob/master/configs-api.md) which should allow to store cloud, runtime, cpi, etc configs in a consistent manner. after some thought i think we can add new type: deployment, to represent deployment manifest updates. eventually i think we should replace manifest command to just use generic config api endpoint. given that generic configs are versioned and store some additional metadata (created at), we will get history effectively for free with recorded manifests. deployment events could record config ids (for deployment, runtime, cloud types etc) in their events. what do you think about this approach?

cc @dpb587-pivotal

petergtz:
@cppforlife It's an excellent initiative and makes perfect sense. I already noticed while implementing my change that part of it is very similar to cloud and runtime config. So I think generalizing these things avoids a lot duplication and definitely makes sense.

One thing I'm wondering how we would do it though in your proposal above: in my PR, I'm taking special care of the bosh releases (and stemcells), to make sure a deployment revision knows the **exact** versions of the bosh releases it deploys. I'm doing this because just having the versioned manifests is not enough, because it might only refer to release versions as `latest`. To get precise diffs we need those exact release versions. So I'm wondering how we could support this. Suggestions?

Also, the [notes](https://github.com/cloudfoundry/bosh-notes/blob/master/configs-api.md) state what is _implemented_. Does that mean it's already on `master`? Or what is the exact state. Maybe @voelzmo can also shed some light on the current state and the intended timeline. Thanks.

cppforlife:
> Also, the notes
<https://github.com/cloudfoundry/bosh-notes/blob/master/configs-api.md> state
what is *implemented*.

implemented there was referring to presence of that type of config in the
director, not it being consolidated. a bit confusing i would agree.

> I'm doing this because just having the versioned manifests is not enough,
because it might only refer to release versions as latest. To get precise
diffs we need those exact release versions. So I'm wondering how we could
support this. Suggestions?

hmm, that is an interesting question, i wonder if we need to add another
type of config, like deployment-processed that would have all that info in
there resolved.

jaresty:
Merge story [#152972966](https://www.pivotaltracker.com/story/show/152972966)

jfmyers9:
Sorry that we never got around to reviewing this PR. At this point it feels like this PR has gotten out of date as there have been a good number of changes in this space. Please feel free to resubmit this PR on the master branch if you would still like us to consider this change.

Best,

@jfmyers9, CF BOSH

PR #1795, opened 2017-09-26
Registry should die after 10 sec

achawki:
- currently we wait forever until the
  registry process dies
- we have seen registry process ignoring
  kill and hanging forever in production
- wait for max. 10 seconds, then force kill it

(same as for director, scheduler and workers
, see commit 156ef3ec9 )

Signed-off-by: Jan Sievers <jan.sievers@sap.com>

### DIFF
--- a/jobs/registry/templates/registry_ctl.erb
+++ b/jobs/registry/templates/registry_ctl.erb
@@ -55,7 +55,15 @@ case $1 in
     if [ ! -z $PID ] && pid_exists $PID; then
       kill $PID
     fi
-    while [ -e /proc/$PID ]; do sleep 0.1; done
+    TRIES=0
+    while [ -e /proc/$PID ]
+    do
+      TRIES=$(( $TRIES + 1 ))
+      if [ $TRIES -gt 100 ]; then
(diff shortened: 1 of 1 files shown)


voelzmo:
@cppforlife @dpb587-pivotal any objections merging this? If not, we can just merge this in.

jsievers:
@voelzmo we keep seeing hung registry processes in production...

jaresty:
@jsievers @voelzmo Does this PR solve the problem it intends to?  In reading the conversation from afar, it isn't clear whether the PR should be merged quickly because it solves your issue, or whether it does not go far enough to solve the problem (and therefore should be ignored in favor of waiting for a full fix).  Can you please clarify?

jsievers:
I think the root cause should be handled separately. I'm afraid I'm not enough into the details of bosh registry to help here. Making sure the kill works even if the normal kill SIGNAL is ignored can still improve the situation as otherwise there must be manual interaction to get bosh up and running again.

cppforlife:
going to switch to bpm: https://www.pivotaltracker.com/story/show/153189574

PR #1796, opened 2017-09-26
Nginx should die after 10 sec

achawki:
- currently we wait forever until the
  nginx  process dies
- wait for max. 10 seconds, then force kill it
- when nginx gets forced killed,
  all child processes gets killed as well

Signed-off-by: Amin Chawki <amin.chawki@sap.com>

### DIFF
--- a/jobs/director/templates/nginx_ctl
+++ b/jobs/director/templates/nginx_ctl
@@ -8,6 +8,10 @@ SSL_DIR=/var/vcap/jobs/director/config/ssl
 TMP_DIR=/var/vcap/data/director/tmp
 RUNAS=vcap
 
+function pid_exists() {
+  ps -p $1 &> /dev/null
+}
+
 case $1 in
 
   start)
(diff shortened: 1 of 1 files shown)


voelzmo:
@cppforlife @dpb587-pivotal any objections merging this? If not, we can just merge this in.

dpb587-pivotal:
Any ideas what causes nginx to hang? Is it mainly long uploads that we don't want to wait for?

voelzmo:
@dpb587-pivotal not entirely sure what the root cause was in the cases where we've seen a lockup of the process. It might be connections not correctly closed by a client or the fact that some clients kept issuing requests in the order of hundreds per second at the server. Uploads were not involved, afaik.

In any case, I guess killing the server is better than having the deploy fail because of this.

jaresty:
Merging story: [#152972674](https://www.pivotaltracker.com/story/show/152972674)

jobs/director/templates/nginx_ctl
```diff
@@ -26,8 +30,22 @@ case $1 in
...
-    while [ -e /proc/$PID ]; do sleep 0.1; done
+    if [ ! -z $PID ] && pid_exists $PID; then
+      kill $PID
+    fi
+    TRIES=0
+    while [ -e /proc/$PID ]
+    do
+      TRIES=$(( $TRIES + 1 ))
+      if [ $TRIES -gt 100 ]; then
+        CHILD_PIDS=$(ps --forest -o pid=  -g $(ps -o sid= $PID) | sed 's/ //g' | grep -v "^${PID}$")
```
tjvman: The `--forest` isn't needed here, and this command relies on nginx always creating its children in such a way that they have the same session id as the master process. 

Since the stemcells install the [`psmisc`](http://psmisc.sourceforge.net/) package, we have access to the pstree utility. we could instead use something like `pstree -p 679 | grep -E "\([0-9]+\)" --only-matching | tr --delete "()"`, which would ensure that we cover both the nested children case and the differing sid/gid case.

tjvman: @cppforlife @dpb587 wdyt

cppforlife:
going to switch to bpm: https://www.pivotaltracker.com/story/show/153189574

PR #1803, opened 2017-10-06
Use calculate_vm_cloud_properties during deployment

ansd:
Allows to use `vm_requirements` block (instead of referencing `vm_type`) in manifest.

[#150256388](https://www.pivotaltracker.com/story/show/150256388)

Signed-off-by: Kai Hofstetter <kai.hofstetter@sap.com>


dpb587-pivotal:
@ansd @voelzmo ive been chatting with tom and danny about naming for this features. i think we've come up with something that you all may like: `vm_resources`.

```
instance_groups:
- name: some-ig
  vm_resources:
    cpu: 16
    ram: 128
  vm_extensions:
  - asdf
  persistent_disk_resources: <--- future, for consistency
    size: 512
    iops: 5000
```

thoughts?

ansd:
What I like about `vm_requirements` is that it underlines that those specified parameters / values are really required in the sense that they represent the **minimum** resources. Also, in the above example, I guess that `persistent_disk` is the resource whereas `size` and `iops` are rather requirements ("That disk must have a size of at least 512GB and at least 5000 iops") than resources?

However, `vm_resources` also makes perfectly sense since cpu, ram and ephemeral disk are indeed resources. So, I'm happy with changing it to `vm_resources`.

cppforlife:
> What I like about vm_requirements is that it underlines that those specified parameters / values are really required in the sense that they represent the minimum resources.

these resources arent necessarily minimum, they are an approximation. though right now most cpis implement calculate_vm_cloud_properties as min.

Kiemes:
@cppforlife as a user, I would hope/assume that these values are minimums. It would be rather confusing if CPIs would behave differently. You say most CPIs are implementing this as min. Do you have a list and which CPI, if it is already implementing the experimental method, is not doing min. I would be interested if there is a good/technical reason for it.

PR #1805, opened 2017-10-09
as an operator, i expect runtime config tags are applied to compilati…

barthy1:
…on vms

[#150831026](https://www.pivotaltracker.com/story/show/150831026)


zachgersh:
This was verified on a bosh-lite deployment by using

```
bosh task X --debug
```

and looking for the `set_vm_metadata` call.

Ran the integration/unit tests as well to verify that all was passing.

PR #1809, opened 2017-10-12
as an operator, i expect to be able to retrieve exact same manifest t…

barthy1:
…hat was submitted

Manifest text, which is sent to director, is stored at database and retrieved from it without modifications.

Manifest.load_from_hash was renamed to Manifest.load_from_text according to input parameter type.
New :manifest_file option was added to deploy method of IntegrationExampleGroup to support ability to pass manifest files from fs directly.
Usage of hash which is based on manifest text in the actual deployment process is out of the scope

(#150869116)[https://www.pivotaltracker.com/story/show/150869116]

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
@@ -83,15 +83,15 @@ module Bosh::Director
         options['dry_run'] = true if params['dry_run'] == 'true'
 
         if (request.content_length.nil?  || request.content_length.to_i == 0) && (params['state'])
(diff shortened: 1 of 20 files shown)


GarfieldIsAPhilosopher:
Hi @barthy1 , I revert the pr because it has some un-intended consequence. To be more specific, if we use latest version in manifest and upload a newer version of the release after the initial deployment, when we recreate vm, it will automatically bump all the instances to use the latest release version. One potential way to solve this problem is to still use the old hash in planner_factory.rb, and record the manifest returned to user in a different location, so that we can separate the internal representation of manifest from manifest presented to the user.

PR #1825, opened 2017-11-03
Support DB connection options requiring paths to files [Fixes #1771]

### LINKED ISSUE #1771: Support TLS database connections
As a platform operator
I want to be able to configure secure (TLS) database connections for the director & registry
In order to be compliant to corporate security requirements.  

Currently the director & registry jobs do not expose properties that would allow to configure secure database connectivity. 

Sequel does support TLS secured connectivity for instance through the `mysql2` database driver (https://github.com/brianmario/mysql2#ssl-options). 

Please extend the directory & registry job spec to allow secure connectivity to (external) backend databases.

peterellisjones:
The BOSH director config allows arbitrary parameters to be passed to Sequel using the `connection_options` object. In order to connect to a MySQL or Postgres database over SSL you need to specify `sslkey`, `sslcert`, `sslca` (mysql) and `sslrootcert` (postgres)<sup>1</sup>. However, these options all refer to _paths to files_ not the contents.

Since there's no easy way to get arbitrary files onto the BOSH director, this means you cannot make secure connections to external databases that use self-signed certs such as Google Cloud SQL<sup>2</sup>.

This commits adds a step in the database configuration code that looks for these fields and saves the contents to `/var/vcap/store/director/db_config` and `/var/vcap/store/director/dns_db_config` before passing the paths to Sequel, fixing #1771.

Some questions:

- Are `/var/vcap/store/director/db_config` and `/var/vcap/store/director/dns_db_config` the right place to store these files?
- As well as the unit tests I've tested manually this by deploying a bosh director that uses the changes to connect to an external MySQL instance and verified that it connects securely. Does it need integration tests as well?
- Should this step happen in `config.rb` or in the bash/erb before being passed to the director process?

cheers,

Pete

---

1) https://github.com/jeremyevans/sequel/blob/master/doc/opening_databases.rdoc
2) https://cloud.google.com/sql/docs/mysql/configure-ssl-instance

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -236,7 +236,7 @@ properties:
     description: Name of the director database
     default: bosh
   director.db.connection_options:
-    description: Additional options for the database
+    description: Additional options for the database. Note that 'sslcert', 'sslca', 'sslkey' and 'sslrootcert' can be passed as strings and will be converted to files before being passed to the database client
     default:
       max_connections: 32  #Maximum size of the connection pool
       pool_timeout: 10     #Number of seconds to wait if a connection cannot be acquired before  raising an error
@@ -265,7 +265,7 @@ properties:
     description: Name of the powerdns database
     default: bosh
   dns.db.connection_options:
-    description: Additional options for the powerdns database
+    description: Additional options for the powerdns database. Note that 'sslcert', 'sslca', 'sslkey' and 'sslrootcert' will be converted to files before being passed to the database client
     default:
       max_connections: 32  #Maximum size of the connection pool
       pool_timeout: 10     #Number of seconds to wait if a connection cannot be acquired before  raising an error
--- a/src/bosh-director/bin/bosh-director-migrate
+++ b/src/bosh-director/bin/bosh-director-migrate
@@ -25,7 +25,8 @@ unless config['db'] && config['db']['database']
   abort ('Director database config missing from config file')
 end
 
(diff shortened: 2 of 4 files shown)


pivotal-jamil-shamy:
thanks @peterellisjones for submitting the pull request !
We've just recently started working on enable this behavoir. 
Regarding the questions (in order):
1- We usually render these kinds of certificates through an ERB template, rather than dumping them into files in `config.rb`.
2- We're still thinking if we need integration tests for the TLS option or have a separate test suite for it.
3- Answer in first point

We'll keep you posted as we go through the changes, we may end up merging the pull request and doing some modifications.

thanks again for the pull request

pivotal-jamil-shamy:
@peterellisjones , btw have you tried connecting to a postgres instance over TLS? In our development we needed to bump the sequel `gem` version to be able to connect.

peterellisjones:
> @peterellisjones , btw have you tried connecting to a postgres instance over TLS? In our development we needed to bump the sequel gem version to be able to connect.

@pivotal-jamil-shamy I'm afraid I didn't try it with postgres, just MySQL

peterellisjones:
hi @pivotal-jamil-shamy,

I closed this since it sounds like you're already working on it and it looks like it's an inconsistent implementation with how you've done other certificates. Is there a public tracker where I can follow the progress of the Bosh team's implementation of this?

thanks,

Pete

PR #1827, opened 2017-11-06
As an operator, I expect to be able to retrieve exact same manifest t…

barthy1:
…hat was submitted.

Manifest text, which is sent to director, is stored at database as raw_manifest and retrieved from it without modifications.

Several variables were renamed to show whether they are text or hash.
New :manifest_file option was added to deploy method of IntegrationExampleGroup to support ability to pass manifest files from fs directly.
Usage of manifest hash which is based on manifest text in the actual deployment process is out of the scope

(#150869116)[https://www.pivotaltracker.com/story/show/150869116]


barthy1:
https://github.com/cloudfoundry/bosh/pull/1809#issuecomment-337423063 comment was taken into account. Fix is done, integration test is added,

PR #1828, opened 2017-11-08
Add rake task to dump postgresql schema

manno:
This rake task should help with analyzing the database.
No more running integration tests and putting in a sleep statement or breakpoint, finding the right director database (the names are all hashes) and finally figuring out credentials just to look at the current schema SQL.

Instead this task creates a database `director_latest_tmp`, runs migrations and dumps the schema to STDOUT and the file `https://github.com/manno/bosh/tree/performance-integration-test`.

The resulting database is missing test data of course, so for running `EXPLAIN ANALYZE` we  might still want to mess with integration test databases.


dpb587-pivotal:
Tracker story: https://www.pivotaltracker.com/story/show/144061967

tjvman:
Thanks for the contribution! 

We have merged this locally, with some additional changes - we added very simple unit tests and removed the `schema` namespace. We also added another rake task which outputs the table descriptions for a human-readable way to see the current tables/columns.

PR #1874, opened 2018-01-12
Set ownership of blobstore to vcap even when it is empty

ebeer:
We were seeing our `/var/vcap/store/blobstore/store` dir being created with ownership set to `root`.

We traced it down to it being created after a BBR restore happened. 
This scenario will happen if the blobstore is empty at the time of backup/restore.


dpb587-pivotal:
Thanks!

PR #1877, opened 2018-01-22
as a director user, i expect to create configs such that only my team…

barthy1:
…s are affected by them (can see, can update, can list)

- admin and team admin can create/delete new configs
- admin, reader, team admin and team reader can list configs
- team admin and team reader can list only congis of the corresponding team
- admin can see teams for each config
- so far each config has information only about one team

[#154135425](https://www.pivotaltracker.com/story/show/154135425)


barthy1:
The PR requires also https://github.com/cloudfoundry/bosh-cli/pull/386

PR #1891, opened 2018-02-10
Enable pure IPv6 Director/UAA/Credhub integration without use of DNS

cppforlife:
Pipeline passed with int and unit tests: https://main.bosh-ci.cf-app.com/teams/main/pipelines/ipv6


cppforlife:
cc @cunnie

dpb587-pivotal:
Related upstream pull request which fixes the need for monkey-patching - https://github.com/ruby/openssl/pull/185

Issue #1900, opened 2018-02-20
attach_disk sets disk size as 1


muralisc:
`attach_disk` currently sets the size of [attached disk as 1](https://github.com/cloudfoundry/bosh/blob/8b375db201031c3256570d39df9d435d61a0a6c3/src/bosh-director/lib/bosh/director/jobs/attach_disk.rb#L74) and the cloud_properties empty.
This leads to a disk copy every time a `bosh start` follows a disk attach operation on a deployment, even when the disk size of the newly attached disk is same as that of the existing disk.

For disks with data already filled the `bosh start` operation takes time as long as the copy.
This leads to the deployment being unavailable for so long.

In cases where the newly attached disk is of same size and user expects a copy,
`bosh attach` could have a flag `--copy` or `--no-copy` which decides weather to copy the data or just attach the disk as is.

cppforlife:
@muralisc yup, i agree, we could probably copy over and persistent cloud properties to avoid unnecessary copy operation on restore.

muralisc:
@cppforlife Took a stab at this issue. Can you take a look and let me know the comments.

cppforlife:
@muralisc returning this information from cpi is not necessary, since director should already know what it sends to the cpi. you should be able to grab it from persistent disk model (holds disk size, and disk cloud properties in its db table)

muralisc:
@cppforlife The issue we are trying to address is not only for an orphaned disk, but any arbitrary disk created by infrastructure.
Initial solution was to read the details of the new disk from the infrastructure and create persistent_disk entry.
The issue link with the AWS CPI pull reqs and comments : https://github.com/cloudfoundry/bosh/issues/1900

voelzmo:
Closing in favor of #2015

muralisc:
@voelzmo opened #2037

PR #1916, opened 2018-03-19
Copy size and cloud properties to new disk

muralisc:
Set PersistentDisk with the actual size of the disk attached.
Also update the cloud properties as reported by the cpi.
This prevents unnecessary copy of disk after a disk attach.

Issue: cloudfoundry/bosh#1900

### DIFF
--- a/src/bosh-director/lib/bosh/director/jobs/attach_disk.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/attach_disk.rb
@@ -17,6 +17,8 @@ module Bosh::Director
         @job_name = job_name
         @instance_id = instance_id
         @disk_cid = disk_cid
+        @size = 1
+        @cloud_properties = {}
         @transactor = Transactor.new
         @disk_manager = DiskManager.new(logger)
         @orphan_disk_manager = OrphanDiskManager.new(logger)
@@ -58,6 +60,8 @@ module Bosh::Director
       def handle_previous_disk(instance)
         previous_persistent_disk = instance.managed_persistent_disk
         previous_persistent_disk.update(active: false)
+        @size = previous_persistent_disk.size
+        @cloud_properties = previous_persistent_disk.cloud_properties
 
         if instance.state == 'stopped'
           @disk_manager.detach_disk(previous_persistent_disk)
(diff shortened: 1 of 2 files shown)


cppforlife:
@muralisc would you mind signing cla or showing sap github org (sap-cloudfoundry).

muralisc:
@cppforlife i have exposed sap-cloudfoundry on my profile.

dpb587-pivotal:
@muralisc I see @SAP listed as your organization, but we technically need your GitHub org membership [publicly visible](https://help.github.com/articles/publicizing-or-hiding-organization-membership/). If you could give that a second look, I'd appreciate it.

muralisc:
@dpb587-pivotal hi, i have updated the visibility.

dpb587-pivotal:
@cppforlife please prioritize this ([#156892576](https://www.pivotaltracker.com/story/show/156892576)); @muralisc was asking

abh1kg:
@cppforlife : Can this be merged?

ashishjain14:
@cppforlife @dpb587-pivotal  this is an important fix for us and would help if we an expedite the review and merge.

cppforlife:
@muralisc @ashishjain14 i think i misunderstood the change that yall did. in the above change, it appears that we are just copying metadata from an unrelated disk (previously used disk) to the new disk. i dont think that this makes sense as these two disks are not related. 

what i was proposing in one of the issues is that we copy orphan's disk metadata onto new disk, but i see that already happens correctly in the case when disk is "unorphaned": https://github.com/cloudfoundry/bosh/blob/ce5b73ff8d821237cb614392d78154bf3e70dc7b/src/bosh-director/lib/bosh/director/orphan_disk_manager.rb#L50-L51

@voelzmo what do you think about this change? should we do it?

muralisc:
@cppforlife The issue we are trying to address is not only an orphaned disk, but an arbitrary disk created by infrastructure. Initial solution was to read the details of the new disk from the infrastructure and create persistent_disk entry.
The issue link with the AWS CPI prs and comments : https://github.com/cloudfoundry/bosh/issues/1900

muralisc:
Closing this and raising https://github.com/cloudfoundry/bosh/pull/2015

PR #1950, opened 2018-05-09
Lock releases individually to reduce lock contention on deploy

henryaj:
We're seeing a number of deployment failures when we make several deployments simultaneously if those deployments all consume the same release, along the lines of:

`Error: Failed to acquire lock for lock:release:consul`

We believe that locking releases individually during `bind_model` will reduce the contention on these locks.

Thanks!

@henryaj and @tinygrasshopper


henryaj:
@cppforlife - what are your thoughts on this approach?

jfmyers9:
Sorry that we never reviewed this PR in a reasonable time frame. At this point I'm going to close this PR. Please feel free to resubmit the PR against the master branch if you would like us to consider this change.

Best,

@jfmyers9, CF BOSH

PR #1959, opened 2018-05-17
Adding the Network Lifecycle feature

kreamyx:
This is the first PR aimed at enabling network lifecycle management.

The PR is based on the discussions in the following proposal:
https://docs.google.com/document/d/1akr17Iep99fCYNqZQv9uoxar7qF04U40O6cYoAsg3lA/

- [x] the appropriate tables to the bosh database were added
- [x] feature enabling/disabling via the director configuration file
- [x] automatic creation of subnet(s) by invoking the CPI's create_subnet

Missing from the proposal:
- [ ] the network garbage collector
- [ ] the new api endpoints for the new proposed cli commands


cppforlife:
few overall changes to do:
- clean up whitespace & tab indentation (rubocop might help with this)
- change cpi level naming from create/delete_subnet to create/delete_network
- add necessary locking for creating/delete subnets so that multiple deployments can be done in parallel

additionally, after thinking a bit more i think we need to switch to a different way of how we match existing subnets. instead of doing two level matching (based on director network name and numeric index) we should switch to using ipv4/v6 ranges for manual networks. if ranges change then we need to create new networks and disassociate old ones from the deployment.

subnet matching for dynamic networks would have to be done a bit differently i think. still thinking how to best do it for them...

src/bosh-director/lib/bosh/director/api/controllers/info_controller.rb
```diff
@@ -26,6 +26,9 @@ def requires_authentication?
               'status' => @powerdns_manager.dns_enabled?,
               'extras' => {'domain_name' => @powerdns_manager.root_domain}
             },
+            'network_lifecycle' => {
```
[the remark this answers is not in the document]
medvedzver: @cppforlife How do you know that director supports it?

kreamyx:
@cppforlife All your comments have been addressed except for those related to the CPI input/output (for example, the netmask bits). I think we need to arrange a meeting to have an agreement on what the cpi input/output interface should look like

src/spec/gocli/integration/network_lifecycle_spec.rb
```diff
@@ -0,0 +1,387 @@
+require_relative '../spec_helper'
```
cdutra: Could you add a test to cover the case of adding a subnet to the network that already exists and will be used by the deployment. According to [the code it should fail](https://github.com/cloudfoundry/bosh/pull/1959/files#diff-de6ac6385f7bc7569a15ef6b5653d4f0R161).

cdutra:
Hi @kreamyx,

Thank you for submitting the PR. We're in the process of reviewing it and we left some comments.
Also we noticed that the commits could be merged and the commit message could better capture the proposed changes.

In the PR you mentioned you should have a meeting with @cppforlife about changing the interface for `create_network` call. @cppforlife suggested we should allow users to have the ability specify only netmask in case the user doesn't care about the actual network address, only the size of it; this way the cpi could pick one based on the IaaS current state. Did you you all come to a conclusion?

cc @belinda-liu

kreamyx:
@cdutra Thanks. Appreciate your feedback and working on it. I will also squash the commits and tidy up the commit messages.

Yes, we already had the meeting with @cppforlife and we agreed to include `netmask_bits` in the subnet definition. It is already included in the PR

Thanks

src/spec/gocli/integration/network_lifecycle_spec.rb
kreamyx: Oh this is a case of `updating` the network which we have decided to address later (we have discussed that with the rest of the team and @cppforlife).

src/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
```diff
@@ -24,12 +27,23 @@ def self.parse(network_spec, availability_zones, global_network_resolver, logger
           subnets << new_subnet
         end
         validate_all_subnets_use_azs(subnets, name)
-        new(name, subnets, logger)
+        manual_network = new(name, subnets, logger)
+        manual_network.set_managed(managed)
+        manual_network
+      end
+
+      def set_managed(managed)
```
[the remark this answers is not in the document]
kreamyx: Sounds good! Thanks

src/bosh-director/spec/unit/deployment_plan/stages/create_network_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+          allow(job_network).to receive(:deployment_network).and_return(network)
+          allow(network).to receive(:managed?).and_return(true)
+          allow(Config).to receive(:network_lifecycle_enabled?).and_return(true)
+          allow(AZCloudFactory).to receive(:create_with_latest_configs).and_return(cloud_factory)
+          allow(cloud_factory).to receive(:get).and_return(cloud)
+          allow(Config).to receive(:current_job).and_return(job)
+          allow(job).to receive(:task_id).and_return(1)
+          allow(job).to receive(:username).and_return('name')
+        end
+        it 'creates all subnets successfully' do
```
jrussett: We should probably have a unit-level test to drive out the behavior of `create_network` being idempotent; specifically, we should probably have a test that verifies it will re-use a pre-existing network by name.

src/spec/gocli/integration/network_lifecycle_spec.rb
```diff
@@ -0,0 +1,387 @@
...
+  let(:cloud_config_hash) { Bosh::Spec::NewDeployments.simple_cloud_config_with_multiple_azs }
+  let(:manifest_hash) { Bosh::Spec::NewDeployments.simple_manifest_with_instance_groups }
+  before do
+    bosh_runner.reset
+    manifest_hash['instance_groups'].first['instances'] = 1
+    manifest_hash['instance_groups'].first['azs'] = ['z1']
+    manifest_hash['instance_groups'].first['networks'] = [{ 'name' => 'a' }]
+  end
+  context 'when deploying a manifest with a managed network' do
+    it 'should have a subnet name' do
```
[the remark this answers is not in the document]
kreamyx: Agreed.

src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
```diff
@@ -0,0 +1,171 @@
...
+                # add relation between deployment and network
+                begin
+                  @deployment_plan.model.add_network(db_network)
+                rescue Sequel::UniqueConstraintViolation
+                  @logger.info('deployment to network relation already exists')
+                end
+                # fetch the subnet cloud properties from the database
+                network.subnets.each do |subnet|
+                  db_subnet = db_network.subnets.find { |sn| sn.name == subnet.name }
+                  raise('cannot find the subnet in the database') if db_subnet.nil?
```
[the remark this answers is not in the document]
kreamyx: @jrussett Can you please give me a scenario where the subnet name can change? Are you referring to a cloud config update that would update the subnet name? If yes, then all cases involving subnet changes, adding subnets, or deleting subnets are not handled by this PR as these actions are related to `network update`. We don't have an agreement yet on how to handle `updates` but we believe that we might need an additional CPI method (`update_network` for example)

src/bosh-director/lib/bosh/director/deployment_plan/manual_network_subnet.rb
```diff
@@ -4,91 +4,96 @@ class ManualNetworkSubnet
...
-        end
-
-        legacy_reserved_ranges.each do |cidr_range|
-          cidr_range.range(0, nil, Objectify: true).each do |ip|
-            restricted_ips.add(ip.to_i) unless static_ips.include?(ip.to_i)
-          end
-        end
-
-        new(network_name, range, gateway, name_servers, cloud_properties, netmask, availability_zone_names, restricted_ips, static_ips)
+        new(network_name, range, gateway, name_servers, cloud_properties, netmask, availability_zone_names, restricted_ips, static_ips, subnet_name = sn_name, netmask_bits = nm_bits)
```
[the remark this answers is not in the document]
kreamyx: OK

src/bosh-director/spec/unit/deployment_plan/stages/create_network_spec.rb
kreamyx: Sounds good!

src/bosh-director/db/migrations/migration_digests.json
```diff
@@ -145,5 +145,6 @@
   "20180411165035_migrate_update_strategy_to_vm_strategy": "116a03febc41f331316bb0d14385093ed941bc68",
   "20180427184448_add_historical_metadata_to_orphaned_vms": "405597203ab69f062481a9896a667617bc3207f7",
   "20180515204145_add_vm_metadata_spec": "3a094f32feef551247aeb82432cb551d4804ac16",
-  "20180613190204_links_api_longtext_support": "555ab4f70cc81a7a5889429f8ddadb08dcd432f0"
-}
\ No newline at end of file
+  "20180613190204_links_api_longtext_support": "555ab4f70cc81a7a5889429f8ddadb08dcd432f0",
+  "20180327200711_create_network_table": "167d33bbd968ad671f2c1f7bac5d5046e77edd4f"
```
luan: Note to merger: rename this migration to be the most recent one (use current date).

kreamyx:
@jrussett , @cdutra , and @jaresty Thank you very much for your feedback and your comments. The PR was updated to address all your comments. I also squashed the commits into one.

src/bosh-director/lib/bosh/director/orphan_network_manager.rb
```diff
@@ -0,0 +1,91 @@
...
+        {
+          'name' => network.name,
+          'type' => network.type,
+          'created_at' => network.created_at.to_s,
+          'orphaned_at' => network.orphaned_at.to_s,
+        }
+      end
+    end
+
+    def delete_network(network)
```
GarfieldIsAPhilosopher: which component will be calling delete_network?It might be a good idea to add network lock before deleting the network to avoid race condition where the same network is being reused by create_networks function

kreamyx: @GarfieldIsAPhilosopher For this PR, no components are calling delete network. Ina future PR, it will be called. Good point about locking!

GarfieldIsAPhilosopher:
Hi @kreamyx ,

Thanks for your PR updates. We have a question about enabling this feature when there are already existing manual networks. 
Since `enable_cpi_management` is a global property on the director, how are we differentiating a new network that should be managed versus an existing manual network that should not be managed? Our concern is that the `managed` property is not stored on the network table, but rather is applied to all networks that are parsed and configured to be `true` or `false` based on the config spec. 

Thanks
@GarfieldIsAPhilosopher && @mikexuu, BOSH team

kreamyx:
Hi @GarfieldIsAPhilosopher 

Thanks for your question.

For a network to be managed, two things have to happen:

1. The global `enable_cpi_management` has to be enabled (it is disabled by default)
2. The network spec in the cloud config has to have the `managed` property set to true (it is false by default)

So enabling the global `enable_cpi_management` in itself won't do anything as no networks will be managed after enabling the feature unless you explicitly specify that you want a network to be managed in the cloud config.

Another thing, the networks table in the database _only_ stores managed networks so there is no need to store the managed property in the table.

Let me know if you have further questions

PR #1985, opened 2018-06-21
Enable trim mode (<% -%>) for Template.render

pc-jedi:
Enable trim mode for BOSH ERB templating if Template.render is been used.


voelzmo:
Closing in favor of #2004 as it had tests. Thanks!

PR #1993, opened 2018-06-28
Refactor instance_updater

jaresty:
- Chose this class because it had a high flog score and high churn
- Reduced complexity of all methods
- Broke into three classes -- InstanceUpdater, RecreateHandler, and
UpdateProcedure
- Passes all unit and integration tests

Signed-off-by: Josh Russett <jrussett@pivotal.io>


src/bosh-director/lib/bosh/director/instance_updater/update_procedure.rb
```diff
@@ -0,0 +1,208 @@
...
+      def perform
+        # Optimization to only update DNS if nothing else changed.
+        if dns_change_only?
+          @logger.debug('Only change is DNS configuration')
+          update_dns
+          return
+        end
+
+        unless instance_plan.already_detached?
+          handle_already_detached_instance_plan
```
mikexuu: Maybe a misnamed method here? Did y'all mean `handle_not_detached_instance_plan` ?

src/bosh-director/lib/bosh/director/instance_updater/update_procedure.rb
```diff
@@ -0,0 +1,208 @@
...
+            return
+          end
+        end
+
+        if instance.state == 'detached'
+          handle_detached_instance
+          return
+        end
+
+        finish_up
```
mikexuu: Is there a better name for `finish_up`? It looks like the method is handling a lot of different responsibilities (recreation, networking, dns, links, state, etc.) Maybe this is a sign to break up `finish_up` even further? The refactor in RecreateHandler.perform looks great and very easy to follow. I can see similar pattern being used in `finish_up`. What are your thoughts?

src/bosh-director/lib/bosh/director/instance_updater/update_procedure.rb
```diff
@@ -0,0 +1,208 @@
...
+          # desired state
+          if instance.state == 'stopped'
+            # Command issued: `bosh stop`
+            instance.update_state
+            return
+          end
+        end
+
+        if instance.state == 'detached'
+          handle_detached_instance
```
mikexuu: It looks like some of the logic in `handle_detached_instance` is repeated in `finish_up` (i.e. releasing network plans, links, dns, updating variable set, etc.). Is there a way we can only have this in one place?

src/bosh-director/lib/bosh/director/instance_updater/update_procedure.rb
jaresty: Yep.

PR #2007, opened 2018-07-17
Clean up unused properties in the HM job

pivotal-jamil-shamy:
Looks like these properties were left behind with the changes to the NATS properties in HM months ago.


mfine30:
@pivotal-jamil-shamy is there a story in the Toronto backlog tracking this? Is this something that would make more sense for SF or Toronto to address? cc @dpb587 @freddesbiens

PR #2011, opened 2018-07-26
use the TrueClass, because it works in more database type.

andyliuliming:


dpb587-pivotal:
We have not seen any errors when using the ruby and database versions that we support. Can you provide more context on why this change would be necessary?

andyliuliming:
@dpb587-pivotal Hi Danny, I'm thinking about to support the mssql for the bosh.
so prepare the database migration for that first.
in case you don't mind making it more general : )

dpb587-pivotal:
Thanks for the additional context. Right now we don't have `mssql` support on our roadmap, but if you/someone want to prove it out, feel free to continue using your fork for testing and developing this. For now, I'm going to close this until there's a bit more confidence in the need for these and other changes (I suspect there are several other changes that will end up being necessary) - it'll be great to see another, more comprehensive PR to discuss if it's something we can support.

FYI, since migration scripts are a little fragile, there are some dependencies and tests relying on checksums of these files which would have changed. Be sure to update those and double check tests for the next PR.

PR #2024, opened 2018-08-09
Export path for psql binary for deprecated restore methods

voelzmo:
Such that calling https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/api/controllers/restore_controller.rb#L7 for just restoring the DB with an existing dump still works.

Wdyt? Can we pull this in or should we rather remove this controller and action completely and tell people with this use-case to rely on their own scripts doing `scp` and `ssh -c`?

### DIFF
--- a/jobs/director/templates/restore-db
+++ b/jobs/director/templates/restore-db
@@ -70,6 +70,7 @@ done
 if [ "$adapter" = 'postgres' ];then
     export PGPASSWORD=$pass
     export PGCLIENTENCODING=UTF8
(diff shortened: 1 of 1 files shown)


jfmyers9:
Seeing as we have removed the old director backup and restore functionality, I'm going to close this PR and [add a chore to remove this template](https://www.pivotaltracker.com/story/show/165300701).

Best,

@jfmyers9, CF BOSH

PR #2068, opened 2018-10-22
Allow custom scopes in local identity provider

keymon:
What is this change about?
----------------------------

We want to add readonly users for bosh for monitoring, but we do
not want to have to configure a full UAA for that.

What tests have you run against this PR?
--------------

```
bundle exec rspec \
    spec/unit/api/user/config_user_manager_spec.rb \
    spec/unit/api/local_identity_provider_spec.rb
```

-----------

When using local authentication, local users can have an additonal
field `scopes` to allow override the assigned scopes. By default the
scope is still `bosh.admin`

Example:

    director:
      user_management:
        local:
          users:
            - name: admin
              password: admin123
            - name: hm
              password: hm123
            - name: prometheus
              password: prometheus123
              scopes: [bosh.read]

-----

No

Tag your pair, your PM, and/or team!
-----

Gov UK PaaS team.

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/local_identity_provider.rb
+++ b/src/bosh-director/lib/bosh/director/api/local_identity_provider.rb
@@ -26,15 +26,14 @@ module Bosh
             raise AuthenticationError
           end
 
-          LocalUser.new(*auth.credentials)
+          username, password = auth.credentials
+          scopes = @user_manager.user_scopes(username)
+
+          LocalUser.new(username, password, scopes)
         end
       end
 
-      class LocalUser < Struct.new(:username, :password)
-        def scopes
-          ['bosh.admin']
-        end
-
+      class LocalUser < Struct.new(:username, :password, :scopes)
         def username_or_client
           username
         end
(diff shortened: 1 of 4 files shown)


mfine30:
Hey @keymon I'm somewhat hesitant to add support for more custom scopes in the local bosh identity provider. Anything auth related increases surface area for security risks and the local users are pretty lightweight, to the point that I think it is obvious that they are not intended for serious use cases. 

I'm also trying to understand your rationale behind this request, along with the Prometheus PR you just closed. Not having admin (write) users for prometheus makes sense, but if you're in an environment where something like that is a concern, I would say using UAA over the changes you're suggesting makes more sense.

keymon:
Yes it is true that you can implement this with UAA. But let me disagree here:

This change allows you to restrict the scopes for local users, with a minimal fingerprint. The code change is quite small, and any local user is a bosh.admin which means all the permissions anyway. 

Instead, having to install a complete UAA, with all the scaffolding and leverage it does actually increase the surface area: new service, new port, new complex auth flow. Also makes it more heavy to run in small and test environments. 

This PR does not add too much logic. It simply allows parametrise a previously hardcoded value (`bosh.admin`) using that value as default.

PR #2077, opened 2018-11-04
custom tags may be attached to DataDog healthmonitor metrics

rosenhouse:

DataDog allows tags to be attached to metrics.  BOSH healthmonitor attaches some tags, e.g. the deployment that originated the metric.  This is great.

But if you've got multiple directors, each with similar-looking deployments, all forwarding metrics to the same DataDog account, then all those metrics get mixed together: There isn't a straightforward way to distinguish metrics based on the originating director.  So, for example, it is tough to make a dashboard showing metrics from only one director, or to compare the same metrics across different directors.

This change allows an operator to set custom tags on the director, which get set on all metrics that the director emits.

This way, you can set different tags on different directors, and thus distinguish between those directors in your datadog dashboards.

We've got [this story](https://www.pivotaltracker.com/story/show/161104272).  The only way we see to make it work today is to discover the director's `cid` value after it is deployed, and then populate our datadog templates with that dynamically-generated value.  The `cid` changes on every re-deploy of the director (which our CI pipelines do every night).   So that means we'd need to update our datadog templates after every deploy.  Not great.

```
docker run --rm -it -e RUBY_VERSION=2.4.4 -e DB=sqlite -v $PWD/:/bosh-src bosh/main /bosh-src/ci/tasks/test-unit.sh
```

Maybe the commit message?

> Custom tags may be attached to DataDog healthmonitor metrics

No.

### DIFF
--- a/jobs/health_monitor/spec
+++ b/jobs/health_monitor/spec
@@ -185,6 +185,11 @@ properties:
     description: Health Monitor Application Key for DataDog
   hm.datadog.pagerduty_service_name:
     description: Service name to alert in PagerDuty upon HM events
+  hm.datadog.custom_tags:
+    description: Tags, as key/value pairs, to add to all metrics sent to DataDog.  See https://docs.datadoghq.com/tagging/
+    example: |
+      env: prod
+      region: eu
 
   # Send metrics to Graphite
   hm.graphite_enabled:
--- a/jobs/health_monitor/templates/health_monitor.yml.erb
+++ b/jobs/health_monitor/templates/health_monitor.yml.erb
@@ -188,6 +188,10 @@ if p('hm.datadog_enabled')
     datadog_plugin['options']['pagerduty_service_name'] = pagerduty_service_name
   end
 
+  if_p('hm.datadog.custom_tags') do |custom_tags|
+    datadog_plugin['options']['custom_tags'] = custom_tags
+  end
+
   params['plugins'] << datadog_plugin
 end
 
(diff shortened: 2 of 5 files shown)


voelzmo:
Some context from my POV: We're not using datadog, but graphite instead. What's working for us is [setting a prefix for graphite metrics](https://github.com/cloudfoundry/bosh/blob/master/jobs/health_monitor/spec#L197) that's either the 'landscape name' or the 'director name' (which already is in your deployment manifest, anyways). Both of these can be populated during deployment time, do not change on re-deployment, and should be unique per BOSH director.

Looking at your use-case and comparing it to the change, I'm not sure if we would need a generic list of key-value pairs to be forwarded to datadog, or if one, well-defined tag (like 'director-name' or 'environment-name', or similar) might suffice.

wdyt?

dpb587-pivotal:
I think this pull request makes sense. I like the hash syntax with support for more than just an env/landscape name in it.

Some historical, related requests around this sort of functionality (I don't they should block a merge of this):

 * forward deployment-level tags thru these plugins
 * a shared health monitor tags property which applies to all plugins

@mfine30 this seems like a fairly cheap thing for us to merge/prioritize.

mfine30:
Thanks for the context and conversations. Let's work and prioritize this in IPM today

rosenhouse:
@mfine30 did this get prioritized?  routing team is interested in using the feature.

mfine30:
@rosenhouse thanks for the poke. We IPM'd on it on Monday and I apparently forgot to prioritize it. It should be prioritized now though!

rosenhouse:
Thanks!

Issue #2030, opened 2018-08-15
BOSH assumes uniqueness of IPs over all networks


MatthiasWinzeler:
**Describe the bug**
In our setup, we have some networks that are different IaaS networks (i.e. NSX networks) but reuse the same IP ranges and talk to the same director. 
The agents communicate to the director using SNAT, so connectivity from the agents to the BOSH director is possible.
Example: We have a cloud config with two networks `service-snat-1` and `service-snat-2`. Both networks are backed by a different IaaS network, i.e. `service-snat-1-ls` and `service-snat-2-ls`. However, they use the same IP range (i.e. 192.168.1.0/24). `service-snat-1` does not have to communicate with `service-snat-2`, only the agent in the service networks communicate to the BOSH director. This works fine, and we can live with the limitation that we can not talk from the director to the VMs, i.e. for `bosh ssh`. 

But: BOSH makes sure each IP is only used once in the whole installation. So when we deploy jobs to these networks, `service-snat-1` gets 192.168.1.2, while `service-snat-2` gets 192.168.1.3 etc.
We would like that BOSH allows reuse of IPs and only checks uniqueness in the context of a cloud config network.

**To Reproduce**
1. Deploy a bosh director on any IaaS
2. Upload stemcell and cloud config with these networks:
```
- name: service-snat-1
  subnets:
  - azs:
    - az1
    - az2
    - az3
    cloud_properties:
      name: service-snat-1-ls
    dns:
    - 172.16.1.5
    - 172.16.1.6
    gateway: 192.168.1.1
    range: 192.168.1.0/24
    reserved:
    - 192.168.1.1
    static: []
  type: manual
- name: service-snat-2
  subnets:
  - azs:
    - az1
    - az2
    - az3
    cloud_properties:
      name: service-snat-2-ls
    dns:
    - 172.16.1.5
    - 172.16.1.6
    gateway: 192.168.1.1
    range: 192.168.1.0/24
    reserved:
    - 192.168.1.1
    static: []
  type: manual
```
3. Deploy any release, once in network `service-snat-1` and once in `service-snat-2`

**Expected behavior**
As you will see, BOSH will successfully bring up the VMs in both networks. However, it will make sure that the used IPs are distinct (i.e. `service-snat-1` gets 192.168.1.2, `service-snat-2` gets 192.168.1.3), even though the networks are completely separated and each IP could be used twice.
[...]

PR #2081, opened 2018-11-08
Overlapping ips

MatthiasWinzeler:

This PR allows overlapping IPs in one BOSH director by removing the global uniqueness constraint on IPs. It addresses the issues described in #2030. I think uniqueness of IPs should only be enforced in context of a single network.

As far as I could find out, BOSH did not enforce global uniqueness of IPs until some time ago, it was introduced with this story: https://www.pivotaltracker.com/n/projects/956238/stories/109680324

I do not fully understand why it was introduced - I guess it was to ease some kind of migration. Maybe this constraint is not needed anymore and we can scope the uniqueness to networks, therefore enabling networks with overlapping IPs. But it nevertheless needs a thorough look by an experienced maintainer whether it breaks something. 

Two tests are broken by my changes:
https://github.com/MatthiasWinzeler/bosh/blob/overlapping-ips/src/bosh-director/spec/unit/deployment_plan/ip_provider/database_ip_repo_spec.rb#L145
https://github.com/MatthiasWinzeler/bosh/blob/overlapping-ips/src/bosh-director/spec/unit/deployment_plan/ip_provider/database_ip_repo_ipv6_spec.rb#L145

I'm not sure the use cases tested are still needed...

In our service provider setup, our customers can choose their own network ranges for their services (similar to AWS). They can overlap (i.e. Coca Cola likes to use 10.0.0.0/24 for their services and Pepsie uses 10.0.0.0/24 for their services). We like to serve all customers with one BOSH director, so BOSH director should allow overlapping IPs. We will make sure that BOSH agents can communicate with the director (i.e. by using SNAT as described in #2030 or by adding a second mgmt network interface that's routed to the director).

Local unit tests. 

Allows overlapping IPs in one BOSH director.

Should not.


mrosecrance:
I have some concerns on how this PR will play with the current state of our networking solution. For instance, we need to support keeping the same manual and vip IP addresses when an instance group decides to switch networks and the move is considered "valid". As part of this move, we aren't always updating the network name saved in the database. Specifically we save the IP/Network combo as part of these calls to [add](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_provider.rb#L112) but may switch network in the [reconciler](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/deployment_plan/network_planner/reservation_reconciler.rb#L50)

Another issues is that [the database_ip_repo](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/database_ip_repo.rb#L11) ignores network name on deletion and would clobber potentially the wrong ip. I'm not sure that respecting network name here would be completely get around the issue as to support manual and vip networks renames we may be exacerbating the issues of not having the correct network name attached to the ip.

MatthiasWinzeler:
Thanks for taking a look! I'm sorry for not letting you know earlier, but we're not needing this functionality anymore. We solved our specific use case with NAT to overcome the overlap limitation.

mrosecrance:
That's great to hear. I'm sorry it took us so long to look at this PR.

PR #2090, opened 2018-11-27
Backfill rspec tests for UpdateDeployment 

jaresty:
Previous description:

> Before this change, no matter what config IDs were being passed into
UpdateDeployment, it was using an empty array for the cloud config
models and runtime config models when preparing the update deployments.
This change adds test coverage for that code path and fixes the loading
of those models at the beginning of UpdateDeployment#perform.

Updated:
This PR backfills unit tests that were missing from the UpdateDeployment unit tests.

### DIFF
--- a/src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
@@ -35,7 +35,7 @@ module Bosh::Director
           @event_log.warn_deprecated(warning)
           cloud_config_models = nil
         else
-          cloud_config_models = Bosh::Director::Models::Config.where(id: [@cloud_config_ids]).to_a
+          cloud_config_models = Bosh::Director::Models::Config.where(id: @cloud_config_ids).to_a
           if cloud_config_models.empty?
             logger.debug('No cloud config uploaded yet.')
           else
@@ -43,7 +43,7 @@ module Bosh::Director
           end
         end
 
-        runtime_config_models = Bosh::Director::Models::Config.where(id: [@runtime_config_ids]).to_a
+        runtime_config_models = Bosh::Director::Models::Config.where(id: @runtime_config_ids).to_a
         if runtime_config_models.empty?
           logger.debug("No runtime config uploaded yet.")
         else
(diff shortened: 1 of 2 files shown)


src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
```diff
@@ -35,15 +35,15 @@ def perform
           @event_log.warn_deprecated(warning)
           cloud_config_models = nil
         else
-          cloud_config_models = Bosh::Director::Models::Config.find_by_ids(@cloud_config_ids)
+          cloud_config_models = Bosh::Director::Models::Config.where(id: @cloud_config_ids).to_a
```
jfmyers9: Is there a reason to change this from using `find_by_ids`? From the definition on the class it seems to be doing a very similar thing.

jaresty: I thought it should be working as is, but the tests showed it was not actually working.  Without this change, I was unable to get the `cloud_config_ids` to result in cloud_config objects passed to the test doubles.

jaresty: I see the issue now - `find_by_ids` is actually a custom find method defined on our model (I had assumed it was provided by Sequel).  I'll investigate a bit further to see why it's not finding the models we're creating for test.

jaresty: We were stubbing its return value in the test.  I removed the stub and restored the original code.  Now this PR is just adding some new test coverage.

jaresty:
@jfmyers9 I backfilled tests for the orphaned network cleanup and saw some results that seemed surprising.  It's possible that I made a mistake in my test setup, but it looks like we are removing unmanaged networks if they are on the deployment model (where being on means that the network names match). That behavior sounds incorrect to me, but maybe I'm wrong.

jfmyers9:
@jaresty Do you think it's valuable to break this out into a chore or a bug to explore further so that we can get a pair to spend some time digging into it during working hours? Otherwise I'll remind the community pair to take a look at this tomorrow.

PR #2093, opened 2018-11-30
Refactor UpdateDeployment

jaresty:
I attempted to reduce the complexity of the #perform method by relying
on the class to encapsulate its dependencies.  Since this class is only
used to call #perform once, this is similar to extracting a method
object for the #perform method.  All unit tests pass.  I made all the
new methods private to prevent clients from calling them directly.  I
also made an attempt to order them by relevance, though I am open to
suggestions for reordering.  This class has a lot of responsibility and
in future refactorings it might be a good idea to break it up into
smaller objects with fewer responsibilities.

### DIFF
--- a/src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
@@ -25,126 +25,70 @@ module Bosh::Director
       end
 
       def perform
-        logger.info('Reading deployment manifest')
-        manifest_hash = YAML.load(@manifest_text)
-        logger.debug("Manifest:\n#{@manifest_text}")
-
-        if ignore_cloud_config?(manifest_hash)
-          warning = "Ignoring cloud config. Manifest contains 'networks' section."
-          logger.debug(warning)
-          @event_log.warn_deprecated(warning)
-          cloud_config_models = nil
-        else
-          cloud_config_models = Bosh::Director::Models::Config.find_by_ids(@cloud_config_ids)
-          if cloud_config_models.empty?
-            logger.debug('No cloud config uploaded yet.')
-          else
-            logger.debug("Cloud config:\n#{Bosh::Director::CloudConfig::CloudConfigsConsolidator.new(cloud_config_models).raw_manifest}")
-          end
-        end
-
-        runtime_config_models = Bosh::Director::Models::Config.find_by_ids(@runtime_config_ids)
-        if runtime_config_models.empty?
-          logger.debug("No runtime config uploaded yet.")
-        else
-          logger.debug("Runtime configs:\n#{Bosh::Director::RuntimeConfig::RuntimeConfigsConsolidator.new(runtime_config_models).raw_manifest}")
-        end
-
-        @deployment_name = manifest_hash['name']
-
         previous_releases, previous_stemcells = get_stemcells_and_releases
+
         context = {}
-        parent_id = add_event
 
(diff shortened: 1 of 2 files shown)


jrussett:
I went over this with @jaresty when we were community pair and talked through some of the design decisions and agreed with them. I really like the idea of extracting smaller portions of logic into digestible functions and then making the "main" function essentially a large composition of helper methods, especially considering so many of these dense classes are difficult to reason about and keep a working mental model of their logic.

I don't feel comfortable merging this PR with him because I think other people on the team should have a chance to look at it and ask questions if they have any, so I'll let someone else merge it in.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
```diff
@@ -344,6 +290,165 @@ def event_context(next_releases, previous_releases, next_stemcells, previous_ste
...
+          logger.debug("No runtime config uploaded yet.")
+        else
+          logger.debug("Runtime configs:\n#{Bosh::Director::RuntimeConfig::RuntimeConfigsConsolidator.new(runtime_config_models).raw_manifest}")
+        end
+        @runtime_config_models
+      end
+
+      def deployment_name
+        manifest_hash['name']
+      end
```
jrussett: Some of these helper methods seem to clearly be helpful in understanding the true nature/type of an object, e.g. there is a more generic variable named `@manifest_text` but the function `raw_manifest_text` better conveys the intent of the object. Why was this one useful?

belinda-liu:
Looks much better to Max and me! We are big fans. Dragging it into the prioritization section of the backlog to give other people a chance to look at it and a pair to properly run integration tests on it.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
```diff
@@ -25,126 +25,70 @@ def dry_run?
...
+      ensure
+        deployment_plan.template_blob_cache.clean_cache!
+      end
 
-            @notifier.send_end_event
-            logger.info('Finished updating deployment')
-            add_event(context, parent_id)
+      def prepare_deployment
+        event_log_stage.advance_and_track('Preparing deployment') do
+          # that's where the link path is created
```
xtreme-andrew-su: This comment can be removed.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
```diff
@@ -25,126 +25,70 @@ def dry_run?
...
-        end
-      rescue Exception => e
-        begin
-          @notifier.send_error_event e unless dry_run?
-        rescue Exception => e2
-          # ignore the second error
-        ensure
-          add_event(context, parent_id, e)
-          raise e
+          # that's where the links resolver is created
```
xtreme-andrew-su: Similarly, this can also be removed.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
jaresty: Removed

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
jaresty: Removed

xtreme-andrew-su:
Overall it looks good and I do like the change into smaller components. Some parts feel like they can still be broken down or even moved to their own class.

PR #2100, opened 2018-12-08
Make (most) tests transactional

luan:
Some time in the past we changed everything to be incidentally truncating the director's database after each unit test. While this generates a completely consistent running environment for the test, it causes the entire suite to slow quite a bit down.

With the changes in this PR I was able to get units to run in about 15 minutes (as opposed to over 1h currently)

### DIFF
--- a/src/.rubocop.yml
+++ b/src/.rubocop.yml
@@ -20,7 +20,7 @@ Metrics/AbcSize:
 
 Metrics/BlockLength:
   Max: 25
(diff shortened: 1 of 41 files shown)


jfmyers9:
This seems beneficial. Since this change touches a large number of files and has the potential for merge conflicts, I'm going to merge this now.

dpb587-pivotal:
In addition to the original motivation behind this change, I'm very happy to see the formatting/stylistic/consistency changes as well. Thanks for taking that extra time!

PR #2107, opened 2018-12-21
Speed up recursive chown

ansd:

Only execute chown for files whose user or group is not already vcap.
$STORE_DIR stores all director tasks. With for example 300,000 files, before this change, this command takes multiple seconds until up to 30 minutes to complete depending on the persistent disk type used.
After this change, this command runs within less than 1 second.
This reduces downtime caused by bosh updates by seconds to minutes.

There are no tests for pre start scripts? I just ran the command on a director and compared how long it takes before and after this change.

@voelzmo @friegger 

We were also discussing whether we even need the recursion?
Why not just `chown vcap:vcap $STORE_DIR`?

### DIFF
--- a/jobs/director/templates/pre-start.erb
+++ b/jobs/director/templates/pre-start.erb
@@ -8,7 +8,7 @@ DB_MUTUAL_TLS_PRIVATE_KEY=/var/vcap/jobs/director/config/db/client_private_key.k
 TMPDIR=/var/vcap/data/director/tmp
 
 mkdir -p $STORE_DIR
-chown -R vcap:vcap $STORE_DIR
+find $STORE_DIR \( -not -user vcap -or -not -group vcap \) -execdir chown vcap:vcap {} \+
 
 chmod -R 0640 $CONFIG_DIR
 find $CONFIG_DIR -type d | xargs -n1 chmod 0750


voelzmo:
I'm happy to merge this to improve the execution time of `pre-start`! The question from above still remains:
> We were also discussing whether we even need the recursion?
Why not just chown vcap:vcap $STORE_DIR?

Can anyone remember a specific reason why we should need to `chown` recursively and not just the directory itself?

dpb587-pivotal:
Oops, I missed those original questions. It looks like the mkdir/chown was introduced for [director](https://github.com/cloudfoundry/bosh/commit/48d7567696cfa99a17713a5decc273b1c269d8a8#diff-30def77cc531683b6e63c9b5e8b77ab6R23) and [worker](https://github.com/cloudfoundry/bosh/commit/48d7567696cfa99a17713a5decc273b1c269d8a8#diff-bf4b44d387be01908ba087597c5c15bfR23) in 2013 as part of a "fix stemcell creation" changeset. I assume that meant running things more strictly as `vcap` and perhaps the recursive part was necessary for migration?

So, good thinking. I don't think we need to recurse, given we've been running as `vcap` for such a long time. I'd hope that nobody is trying to upgrade from v1.3262.25.0 to v268.4+ nowadays :)

voelzmo:
Incredible code archeology, thanks! I've merged this change for now, but we can just remove the recursive part, if we feel we don't need it anymore, I guess.

PR #2112, opened 2019-01-10
Added update feature to managed networks

keramix:

This allows managed networks to be updated in the cloud config. Currently, managed networks are only allowed to be created and deleted. But once a managed network is created, there is no way to update the subnets of that particular network (adding, removing, or updating existing subnets). This PR introduces this feature.

Unit tests, Integration test, and I manually tested the update process.

No

The changes in this PR are based on the following document
https://docs.google.com/document/d/1E72c5tetQkxXd0zRIpmFQVE2aQ-_kEPMs11bmrCkb5Q/edit?usp=sharing

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_deleter.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_deleter.rb
@@ -48,6 +48,8 @@ module Bosh::Director
             if network.deployments.size == 1
               OrphanNetworkManager.new(Config.logger).orphan_network(network)
             end
+
+            remove_unused_subnets(deployment_model, network)
           end
         end
       end
@@ -55,6 +57,53 @@ module Bosh::Director
       event_log_stage.advance_and_track('Destroying deployment') do
         @logger.info('Destroying deployment')
         deployment_model.destroy
+      end
+    end
+
+    private
+
+    def unused_subnets(new_subnets, old_subnets)
+      old_subnets.select do |subnet|
+        new_subnets.find { |e| e['name'] == subnet.name }.nil?
+      end
+    end
+
+    def latest_cloud_config(deployment)
+      Models::Config.latest_set_for_teams('cloud', *deployment.teams).max(&:id).raw_manifest
+    end
+
+    def used_cloud_config_state(deployment)
+      latest = Models::Config.latest_set_for_teams('cloud', *deployment.teams).map(&:id).sort
+      if deployment.cloud_configs.empty?
+        'none'
+      elsif deployment.cloud_configs.map(&:id).sort == latest
+        'latest'
+      else
+        'outdated'
+      end
+    end
+
+    def delete_subnet(cloud_factory, subnet)
+      cpi = cloud_factory.get(subnet.cpi)
+      begin
+        @logger.info("deleting unused subnet #{subnet.name}")
+        cpi.delete_network(subnet.cid)
+      rescue StandardError => e
+        @logger.error("failed to delete subnet #{subnet.name}: #{e.message}")
+      end
+
+      subnet.destroy
+    end
+
+    def remove_unused_subnets(deployment_model, network)
+      other_deployments = network.deployments.delete_if { |x| x == deployment_model }
+      return if other_deployments.empty?
+      return unless other_deployments.all? { |deployment| used_cloud_config_state(deployment) == 'latest' }
+      cloud_config_hash = latest_cloud_config(other_deployments.first)
+      network_spec = cloud_config_hash['networks'].find { |x| x['name'] == network.name }
+      unused_subnets(network_spec['subnets'], network.subnets).each do |subnet|
+        cloud_factory = AZCloudFactory.create_with_latest_configs(deployment_model)
+        delete_subnet(cloud_factory, subnet)
       end
     end
   end
--- a/src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
@@ -16,36 +16,146 @@ module Bosh::Director
 
         private
 
+        def network_update_plan(network_name, network_subnets, subnet_models, deployment_cloud_config)
+          old_network = deployment_cloud_config['networks'].find { |x| x['name'] == network_name }
+          old_subnets = old_network['subnets']
+
+          subnets_to_add = network_subnets.select do |subnet|
+            subnet_models.find { |e| e.name == subnet.name }.nil?
+          end
+
+          subnets_to_remove = subnet_models.select do |subnet|
+            network_subnets.find { |e| e.name == subnet.name }.nil?
+          end
+
+          # Exclude all subnets that only had name changes
+          subnets_to_remove.delete_if do |subnet_model|
+            old_subnet = old_subnets.find { |x| x['name'] == subnet_model.name }
+            # old_subnet will be nil only when a subnet has actually been modified (not name change).
+            # A new database entry will be created with "subnet-outdated-<uuid>"
+            next if old_subnet.nil?
+            next unless old_subnet.key?('range')
+
+            old_range = NetAddr::CIDR.create(old_subnet['range'])
+            old_gateway = NetAddr::CIDR.create(old_subnet['gateway'])
+
+            s = subnets_to_add.find do |new_subnet|
+              new_subnet.range == old_range && \
+                new_subnet.gateway == old_gateway && \
+                new_subnet.cloud_properties == old_subnet['cloud_properties']
+            end
+
+            unless s.nil?
+              subnet_model.name = s.name
+              subnet_model.save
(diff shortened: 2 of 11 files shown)


src/bosh-director/lib/bosh/director/deployment_deleter.rb
```diff
@@ -57,5 +59,52 @@ def delete(deployment_model, instance_deleter, vm_deleter)
...
+      end
+    end
+
+    def delete_subnet(cloud_factory, subnet)
+      cpi = cloud_factory.get(subnet.cpi)
+      begin
+        @logger.info("deleting unused subnet #{subnet.name}")
+        cpi.delete_network(subnet.cid)
+      rescue StandardError => e
+        @logger.error("failed to delete subnet #{subnet.name}: #{e.message}")
```
pivotal-dale-wick: It seems unusual to attempt to remove the subnet, and then if it fails, remove it from the bosh database, and never retry deleting it.  

Maybe it shouldn't destroy it from the bosh database in the failure state, so that the next time it does a cleanup, bosh can try again.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
```diff
@@ -89,6 +89,60 @@ def prepare_deployment
...
+
+      # removes a subnet model in the iaas and from the database
+      def delete_subnet(cloud_factory, subnet)
+        cpi = cloud_factory.get(subnet.cpi)
+        begin
+          @logger.info("deleting unused subnet #{subnet.name}")
+          cpi.delete_network(subnet.cid)
+        rescue StandardError => e
+          # failing to delete the subnet in the iaas shouldn't fail the deployment
+          @logger.error("failed to delete subnet #{subnet.name}: #{e.message}")
```
pivotal-dale-wick: It seems unusual to attempt to remove the subnet, and then if it fails, remove it from the bosh database, and never retry deleting it.  

Maybe it shouldn't destroy it from the bosh database in the failure state, so that the next time it does a cleanup, bosh can try again.

pivotal-dale-wick: Does the CPI result give a hint of whether the IaaS isn't responding versus the subnet is already deleted?  Maybe that could be the difference of whether the subnet db model should be purged or not.

kreamyx: Currently, the CPI doesn't give any hints regarding why a failure could have occurred. I was initially thinking of removing the subnet from the IaaS as a best effort attempt. Meaning that Bosh will attempt to remove the subnet, if it fails for any reason, Bosh would remove this subnet from the database and stop managing it and then it will be left to the cloud operator to clean things up. But you are right, another approach could be to leave it in the database. This way, Bosh will attempt to remove it again the next time the deployment is updated. I have no strong opinions either way.

src/bosh-director/lib/bosh/director/deployment_deleter.rb
kreamyx: Commented below.

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
kreamyx: One problem I see in the second approach is:
1- deployment created with a managed network with two subnets
2- cloud config updated, and the second subnet is removed from the config
3- IaaS returns an error "can not delete subnet" and the deployment update proceeds
4- The cloud operator manually cleans up the no longer used subnet
5- any further update of the deployment will attempt to remove the subnet from the iaas (which no longer exists) but will fail
In this scenario, the subnet will never be removed from the database.

I have another PR ready to be submitted for this approach if this is the desired behavior.

jfmyers9: Hi @kreamyx,

Sorry we haven't responded in a while.

With regards to the second approach, is it possible for us to detect whether or not the subnet is already deleted? If so, then when deleting a subnet, if the subnet does not exist we can safely move on. This would be similar to the behavior for VMs.

Thoughts?

Also it appears that this PR has gotten somewhat stale due to our inactivity. Is there any chance you could rebase this on PR with the master branch?

Best,

@jfmyers9 && @belinda-liu, CF BOSH

kreamyx: Hi @jfmyers9 and @belinda-liu 

Thanks for reviewing the PR.

Currently, there is no way to find out if a subnet doesn't exist in the IAAS (we currently only support `create_network` and `delete_network` in the CPI). But I think the idea of having an additional `has_network` in the future would be very useful in this case (and perhaps other cases as well), and it will make it consistent with the VM-related CPI methods as well.

I have also rebased my branch on top of master.

Thanks,
Karim

src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
```diff
@@ -16,37 +16,147 @@ def perform
 
         private
 
+        def network_update_plan(network_name, network_subnets, subnet_models, deployment_cloud_config)
+          old_network = deployment_cloud_config['networks'].find { |x| x['name'] == network_name }
+          old_subnets = old_network['subnets']
```
belinda-liu: We're having a hard time understanding why there are three sets of subnets to consider. Our current understanding is that `subnet_models` are what are currently saved in the database, and therefore existing subnets, and `network_subnets` are the desired subnets. What do `old_subnets` mean? If there are the subnets that are defined in the cloud config, why are they "old"?

src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
```diff
@@ -16,37 +16,147 @@ def perform
...
+                       subnet.range.contains?(old_range)
+
+                      raise NetworkOverlappingSubnets, "Subnet '#{subnet.name}' in managed network '#{network.name}' cannot be modified to an overlapping subnet" if to_remove_subnet.name.start_with?(subnet.name)
 
-                # fetch the subnet cloud properties from the database
-                network.subnets.each do |subnet|
-                  db_subnet = db_network.subnets.find { |sn| sn.name == subnet.name }
-                  raise Bosh::Director::SubnetNotFoundInDB, "cannot find subnet: #{subnet.name} in the database" if db_subnet.nil?
-                  populate_subnet_properties(subnet, db_subnet)
+                      raise NetworkOverlappingSubnets, "Updating managed network '#{network.name}' doesnot support overlapping subnets. Subnet '#{subnet.name}' overlaps with a subnet that is scheduled for removing"
```
belinda-liu: "doesnot" -- fix spacing

src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
```diff
@@ -16,37 +16,147 @@ def perform
...
+                subnets_to_add, subnets_to_remove = network_update_plan(
+                  network.name,
+                  network.subnets,
+                  db_network.subnets,
+                  deployment_cloud_config,
+                )
+
+                subnets_to_add.each do |subnet|
+                  subnets_to_remove.each do |to_remove_subnet|
+                    next unless subnet.range
```
belinda-liu: Why aren't we nil checking the `to_remove_subnet.range` as well?

src/bosh-director/lib/bosh/director/deployment_plan/stages/create_network.rb
```diff
@@ -115,25 +225,34 @@ def create_subnet(subnet, network_model, rollback)
...
+            predeployment_cloud_properties: JSON.dump(subnet.cloud_properties),
+            netmask_bits: subnet.netmask_bits,
             name: subnet.name,
             range: range,
             gateway: gw,
             reserved: JSON.dump(reserved_ips),
             cpi: cpi_name,
           )
+
+          sn.type = if subnet.netmask_bits
```
belinda-liu: Could you change this to a ternary operator?

src/bosh-director/lib/bosh/director/jobs/update_deployment.rb
belinda-liu: Since we do not currently have a `has_network` method, would it be possible to distinguish between when the deletion fails due to the fact that the subnet no longer exists? This would allow BOSH to safely retry even if the operator has already manually cleaned up the IaaS. 
If that's possible, then let's keep the subnet in the database so that BOSh can retry when deleting fails.

miguelverissimo:
Hi @kreamyx recent changes have cause conflicts with your PR, let us know if you need more info on those changes when merging back master!

kreamyx:
@miguelverissimo Thanks for the heads up! I will look at this PR and the comments as soon as I have some bandwidth.

mrosecrance:
I'm going to close this as it's stale.

Issue #2113, opened 2019-01-12
Manifest values starting with colons are considered symbols


dpb587:
This probably isn't a common issue, but when values start with a colon, they seem to be interpreted as a symbol. Depending on how those values are serialized into a template (e.g. with `JSON.generate`) the first colon will be dropped and can easily cause runtime errors.

**To Reproduce**
Given a release template with:

```erb
json: <%= JSON.generate({ "value" => p('value') }) %>
yaml: <%= YAML.dump({ "value" => p('value') }) %>
```

When it is deployed with:

```yaml
...snip...
  jobs:
  - name: test
    properties:
      value: "::1"
```

Then it [currently] renders:

```
json: { "value": ":1" } # only a single colon
yaml: { value: :":1" }  # symbol syntax
```

**Expected behavior**
It seems like deployment manifests are parsed by YAML with support for symbols. I would expect the values in the manifest to be treated as regular strings since there's no value to symbols in BOSH configurations. Then it would render as:

```
json: { "value": "::1" }
yaml: { value: "::1" }
```

**Versions:**

 * bosh/268.4.0
 * bosh-cli/5.4.0

**Deployment info:**
I ran into this while configuring a release with an IPv6 localhost reference as a value property. It starts with double colon, but then failed in runtime because a single colon is an invalid IPv6/CIDR. For now I worked around it in the release template by manually converting symbols in the related field, but it wouldn't fix other fields. Roughly...

```erb
both: <%= JSON.generate({ 'value' => p('value') { |v| v = ":#{v}" if v.is_a?(Symbol); v } }) %>
```

An alternative hack would probably be to add a third colon to the value in the manifest.

PR #2115, opened 2019-01-14
Fix bosh backup scripts for external DBs

aclevername:

Changing the bosh directors backup script to produce an empty artifact when configured with an external DB

When you configure your bosh director to use an external DB and attempt to backup using [bbr](https://github.com/cloudfoundry-incubator/bosh-backup-and-restore) it would produce an artifact that was invalid. This is due to the ruby gem used to create the backup artifact does not support external DB's, only internal. If you tried to restore with this artifact it would `succeed` falsely, leaving the director in a bad state. To negate this we want it to produce an empty backup artifact instead and for restore to no-op when configured with an external DB.

Story that discovered this issue: https://www.pivotaltracker.com/story/show/161379117
Story to fix this issue: https://www.pivotaltracker.com/story/show/162088113

Manual testing with the following configuration:
`director.db.host` set to `localhost` - resulted in backup script being correctly templated
`director.db.host` set to `127.0.0.1` - resulted in backup script being correctly templated
`director.db.host` set to `<exteranl-db>` - resulted in backup script being templated empty (as desired)

Bug Fix- BOSH Director backup no longer produces an artifact when configured with an external DB

No

@gmrodgers

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -19,7 +19,6 @@ templates:
   mime.types: config/mime.types
   config_server_ca.cert.erb: config/config_server_ca.cert
   uaa_server_ca.cert.erb: config/uaa_server_ca.cert
-  sync_dns_ctl.erb: bin/sync_dns_ctl
   bbr_backup.erb: bin/bbr/backup
   bbr_restore.erb: bin/bbr/restore
   nats_server_ca.pem.erb: config/nats_server_ca.pem
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -19,8 +19,9 @@ templates:
   mime.types: config/mime.types
   config_server_ca.cert.erb: config/config_server_ca.cert
   uaa_server_ca.cert.erb: config/uaa_server_ca.cert
-  bbr_backup: bin/bbr/backup
-  bbr_restore: bin/bbr/restore
+  sync_dns_ctl.erb: bin/sync_dns_ctl
+  bbr_backup.erb: bin/bbr/backup
+  bbr_restore.erb: bin/bbr/restore
   nats_server_ca.pem.erb: config/nats_server_ca.pem
   nats_client_certificate.pem.erb: config/nats_client_certificate.pem
   nats_client_private_key.erb: config/nats_client_private_key
--- a/jobs/director/templates/bbr_backup.erb
+++ b/jobs/director/templates/bbr_backup.erb
@@ -31,4 +31,6 @@ export BUNDLE_GEMFILE=/var/vcap/packages/director/Gemfile
 export GEM_HOME=/var/vcap/packages/director/gem_home/ruby/2.4.0
 
 /var/vcap/packages/director/bin/bosh-backup
+<% else %>
+echo "Skipping director backup because an external database is configured."
 <% end %>
--- a/jobs/director/templates/bbr_backup.erb
+++ b/jobs/director/templates/bbr_backup.erb
@@ -2,7 +2,7 @@
 
 set -eu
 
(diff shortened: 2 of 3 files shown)


aclevername:
Still attempting to test when the internal db is configured with `::1`, see here https://pivotal.slack.com/archives/C3M0B3TV3/p1547134391134800

Please do not merge yet.

aclevername:
A issue has been opened regarding the inability to configure `::1` here: https://github.com/cloudfoundry/bosh/issues/2113

aclevername:
After talking to the BOSH team we are happy that it will work correctly for `::1`.

gmrodgers:
Hi @dpb587-pivotal 
### How do you see BBR users dealing with configuring database backups?
Not quite sure about this question.  The current external database backups produced by the restore script in this version are invalid and will restore the director to an unhealthy state when used. We would like to remove this attempt to backup external databases as it was not promised by `bbr` until the BBR SDK was introduced in BOSH v267.2.  External databases would have to be backed up outside of `bbr`.

### Should they need to know to differentiate between internal and external databases?
Yes as the internal database is backed up using `bbr` but the external was never backed up for this version we are PRing into.

### Should they be expected to add `bbr-database-X` backup jobs if they're using an external database? How does that get communicated?
The external database backup feature was never promised or communicated to the consumer of the BOSH release until the usage of the BBR SDK release in BOSH v267.2.  If we were to communicate this, we believe something similar to [this matrix](https://docs.pivotal.io/pivotalcf/2-3/customizing/backup-restore/backup-pcf-bbr.html#-external-storage-support-across-pcf-versions) would be useful.  We believe that they would have to backup the external database outside of using `bbr` and therefore not create a new `bbr-database-X` job

### Should the backup/restore functionality instead be removed from director and into the database-specific jobs which provide the local postgres database?
Not sure about this question either with regards to the PR.  This is up to the BOSH team as `bbr` will run the `backup` and `restore` scripts if they are found in the `bin/bbr/` directory of any job so the functionality should be similar if you decide to implement this.  The fix would be good regardless we are giving a false impression to the consumers of BOSH that we back up external databases for this version.

### Should the current director backup/restore functionality be fixed to work for both local and remote databases?
No as this is not a fix for backing up external databases but a fix to stop it looking like we back up external databases.

Thanks Glen

alamages:
Hello @dpb587-pivotal, is there any update on this?

dpb587-pivotal:
@miguelverissimo while it would be nice to provide more awareness into what might be happening in the backup, I don't believe standard output is streamed to the client where they would see that sort of message.

cc @mfine30 @jfmyers9 for prioritizing/continuing this.

mfine30:
@miguelverissimo @dpb587 my interpretation of reading this thread is there is some potential UX improvements we could make, but they may be challenging based on current architecture/code behavior. I'd be inclined to say we should merge this and we can improve the output if/when we get feedback that we should add more.

dpb587-pivotal:
@mfine30 accurate summary.

jamesjoshuahill:
Hi hi

Thank you for your comments @dpb587 and @mfine30.

@miguelverissimo we have added the `else` you suggested. This skip message will be visible in `bbr` output when the `--debug` flag is set.

Josh and @alamages

jamesjoshuahill:
Hi @h4xnoodle 

Thank you for merging our PR. How can we track the next release of v266.x? Do you know when the next patch release is due?

Josh and @alamages

h4xnoodle:
Hi @jamesjoshuahill, we're tracking in this story: https://www.pivotaltracker.com/story/show/163197005 and once @mfine30 can accept then we'll cut a release, and the next teams will be notified/pick it up automatically.

PR #2116, opened 2019-01-15
Add ability to supply link address in template test

tylerschultz:

This change allows users to fake out 'link.address' when writing template rendering tests. Without this change, templates that contain 'link.address' cause the tests to fail during rendering.

NA

This test only changes the test folder of the bosh-template gem. The test folder contains classes that are only used when writing tests for a template. There is no change to the production template classes. Therefore, I've only ran the spec folder in the template-gem.

Improved template test support - 'link.address' can now be specified and verified in rendered templates.

No

@christianang @xanderstrike @ameowlia @oozie

### DIFF
--- a/src/bosh-template/lib/bosh/template/test/link.rb
+++ b/src/bosh-template/lib/bosh/template/test/link.rb
@@ -1,11 +1,12 @@
 module Bosh::Template::Test
   class Link
-    attr_reader :instances, :name, :properties
+    attr_reader :instances, :name, :properties, :address
 
-    def initialize(name:, instances: [], properties: {})
+    def initialize(name:, instances: [], properties: {}, address: nil)
       @instances = instances
       @name = name
       @properties = properties
+      @address = address
     end
 
     def to_h
@@ -13,6 +14,7 @@ module Bosh::Template::Test
         'instances' => instances.map(&:to_h),
         'name' => name,
         'properties' => properties,
+        'address' => address,
       }
     end
   end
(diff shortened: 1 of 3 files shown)


tylerschultz:
It occurred to me this change will also require a new version of bosh-template gem is created/published.

dpb587-pivotal:
Thanks! Do you have an example of a test where you're using this? Given that `link(...).address` is technically a method which accepts arguments, I'm wondering how we could help make those calls fully testable (e.g. that a release is specifically looking for "healthy" instances).

Have any ideas about how to solve that, or examples where it might be useful in your template?

tylerschultz:
You're right, the changes I've made here only cover `link.address` (with no args). I'm not sure what would happen if args were passed to link.address, as it is being treated as a ManualLink by the EvaluationContext. 

Despite this shortcoming, I'd argue there's still value here. This change enables users of `link.address` with no args to run template tests, which is not possible without this change. Tests fail with 'must be run with director' error when `link.address` is accessed.

I think I've convinced Danny to spend some flex time with me to come up with ways we might go about handling arguments to `link.address`. 

We can do that as a separate PR, or we can improve this PR. LMK what you'd like.

belinda-liu:
Hey @tylerschultz,

Sorry for letting this sit for a while. Being able to use `link.address` with no args seems like a valuable thing to support, and the change is straightforward enough. If you've worked on allowing `link.address` with args, we'll be happy to take that as a separate PR so we can close out this one. Thanks!

tylerschultz:
HI @belinda-liu, Thanks for merging. 

I've recently tried to accommodate arguments to the address method, but it's proving non trivial.

Friendly reminder: this change won't get published until a new version of the template gem is publish.

mfine30:
@tylerschultz we have a new story here to cut a new gem: https://www.pivotaltracker.com/story/show/165892023

We'll IPM on it next week

PR #2121, opened 2019-01-22
Pre fetch agent

voelzmo:
Pre-fetch the `agent_id` for all instances selected for a dns update to avoid unnecessary DB requests. Previously, much time was spent in the `agent_broadcaster#sync_dns` method, which is even logged at the end of the method:
```
INFO -- DirectorJobRunner: agent_broadcaster: sync_dns: attempted 1619 agents in 53211ms (555 successful, 0 failed, 1167 unresponsive)
```

Even for 1600 VMs, this method should not take 53 seconds, but rather 10 seconds + `x`, for `x` in the range of ms.

There were thousands of DB requests logged like 
```
DirectorJobRunner: (0.000409s) (conn: 47449216593120) SELECT * FROM "vms" WHERE (("instance_id" = 3683) AND ("active" IS TRUE)) LIMIT 1
```
because each access to `instance.agent_id` triggers a request like this. This was called during the regular 10-second `sync-dns` job as well as during all forced syncs during the update of instances.

`bundle exec rake spec:unit`
Deploy with a Director dev release.

Probably not release note relevant?

no

/cc @cloudfoundry/cf-bosh-europe

### DIFF
--- a/src/bosh-director/lib/bosh/director/agent_broadcaster.rb
+++ b/src/bosh-director/lib/bosh/director/agent_broadcaster.rb
@@ -14,12 +14,12 @@ module Bosh::Director
       @logger.info("deleting arp entries for the following ip addresses: #{ip_addresses}")
       instances = filter_instances(vm_cid_to_exclude)
       instances.each do |instance|
-        agent_client(instance.agent_id, instance.name).delete_arp_entries(ips: ip_addresses)
+        agent_client(instance[:fetched_agent_id], "#{instance[:job]}/#{instance[:uuid]}").delete_arp_entries(ips: ip_addresses)
       end
     end
 
     def sync_dns(instances, blobstore_id, sha1, version)
-      @logger.info("agent_broadcaster: sync_dns: sending to #{instances.length} agents #{instances.map(&:agent_id)}")
+      @logger.info("agent_broadcaster: sync_dns: sending to #{instances.length} agents #{instances.map { |i| i[:fetched_agent_id]}}")
 
       lock = Mutex.new
 
@@ -33,18 +33,19 @@ module Bosh::Director
 
       instances.each do |instance|
         pending.add(instance)
-        instance_to_request_id[instance] = agent_client(instance.agent_id,
-                                                        instance.name).sync_dns(blobstore_id, sha1, version) do |response|
+        agent_id = instance[:fetched_agent_id]
+        instance_to_request_id[instance] = agent_client(agent_id,
+                                                        "#{instance[:job]}/#{instance[:uuid]}").sync_dns(blobstore_id, sha1, version) do |response|
           valid_response = (response['value'] == VALID_SYNC_DNS_RESPONSE)
 
           if valid_response
-            updated_rows = Models::AgentDnsVersion.where(agent_id: instance.agent_id).update(dns_version: version)
+            updated_rows = Models::AgentDnsVersion.where(agent_id: agent_id).update(dns_version: version)
 
             if updated_rows == 0
               begin
-                Models::AgentDnsVersion.create(agent_id: instance.agent_id, dns_version: version)
+                Models::AgentDnsVersion.create(agent_id: agent_id, dns_version: version)
               rescue Sequel::UniqueConstraintViolation
-                Models::AgentDnsVersion.where(agent_id: instance.agent_id).update(dns_version: version)
+                Models::AgentDnsVersion.where(agent_id: agent_id).update(dns_version: version)
               end
             end
           end
@@ -54,7 +55,7 @@ module Bosh::Director
               num_successful += 1
             else
               num_failed += 1
-              @logger.error("agent_broadcaster: sync_dns[#{instance.agent_id}]: received unexpected response #{response}")
+              @logger.error("agent_broadcaster: sync_dns[#{agent_id}]: received unexpected response #{response}")
             end
             pending.delete(instance)
           end
@@ -81,14 +82,15 @@ module Bosh::Director
 
         unresponsive_agents = []
         pending_clone.each do |instance|
-          agent_client = agent_client(instance.agent_id, instance.name)
+          agent_id = instance[:fetched_agent_id]
+          agent_client = agent_client(agent_id, "#{instance[:job]}/#{instance[:uuid]}")
           agent_client.cancel_sync_dns(instance_to_request_id[instance])
 
           lock.synchronize do
             num_unresponsive += 1
           end
 
-          unresponsive_agents << instance.agent_id
+          unresponsive_agents << agent_id
         end
         if num_unresponsive > 0
           @logger.warn("agent_broadcaster: sync_dns: no response received for #{num_unresponsive} agent(s): [#{unresponsive_agents.join(', ')}]")
@@ -102,9 +104,10 @@ module Bosh::Director
     end
 
     def filter_instances(vm_cid_to_exclude)
-      Models::Instance
+      Config.db[:instances]
         .inner_join(:vms, Sequel.qualify('vms', 'instance_id') => Sequel.qualify('instances','id'))
         .select_append(Sequel.expr(Sequel.qualify('instances','id')).as(:id))
+        .select_append(Sequel.expr(Sequel.qualify('vms','agent_id')).as(:fetched_agent_id))
         .where { Sequel.expr(Sequel.qualify('vms','active') => true) }
         .exclude { Sequel.expr(Sequel.qualify('vms','cid') => vm_cid_to_exclude) }
         .where(compilation: false)
--- a/src/bosh-director/lib/bosh/director/dns/dns_version_converger.rb
+++ b/src/bosh-director/lib/bosh/director/dns/dns_version_converger.rb
@@ -3,9 +3,9 @@ module Bosh::Director
 
     ONLY_OUT_OF_DATE_SELECTOR = lambda do |current_version, logger|
       logger.info('Selected strategy: ONLY_OUT_OF_DATE_SELECTOR')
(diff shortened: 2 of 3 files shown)


src/bosh-director/lib/bosh/director/dns/dns_version_converger.rb
```diff
@@ -3,9 +3,9 @@ class DnsVersionConverger
 
     ONLY_OUT_OF_DATE_SELECTOR = lambda do |current_version, logger|
       logger.info('Selected strategy: ONLY_OUT_OF_DATE_SELECTOR')
-      Models::Instance.inner_join(:vms, Sequel.qualify('vms', 'instance_id') => Sequel.qualify('instances', 'id'))
+      Config.db[:instances].inner_join(:vms, Sequel.qualify('vms', 'instance_id') => Sequel.qualify('instances', 'id'))
```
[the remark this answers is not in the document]
pivotal-dale-wick: I wouldn't use an inner_join, since ruby sequel makes a magic foreign key variable called: instance.vms: https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/models/instance.rb#L9

If you have a given instance, you can just say: `instance.vms.each` to iterate through the matching vms for the instance, or `instance.vms.all` for an array of the vms.  And it should be super quick.  Alternately maybe an index would make the query resolve quicker if that isn't an option.

From the snippet I can't tell if that would help or not though.

src/bosh-director/lib/bosh/director/agent_broadcaster.rb
```diff
@@ -79,24 +83,25 @@ def sync_dns(instances, blobstore_id, sha1, version)
           pending_clone = pending.clone
         end
 
+        start_time_cancelling = Time.now
         unresponsive_agents = []
         pending_clone.each do |instance|
-          agent_client = agent_client(instance.agent_id, instance.name)
-          agent_client.cancel_sync_dns(instance_to_request_id[instance])
+          @nats_rpc.cancel_request(instance_to_request_id[instance])
```
degaurab: replacing `cancel_sync_dns` with direct rpc call kind of confusing as it don't explain what we are actually canceling until you look around the code. For maintainability aspect the code it look more cleaner approach, but kind of hiding the intent as the rpc method is very generic (as the agent method is just a wrapper around the rpc call). Just a view...

voelzmo: Good catch, this seems to have gone wrong during squashing the changes.

src/bosh-director/lib/bosh/director/dns/dns_version_converger.rb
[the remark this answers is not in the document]
charleshansen: Not fully understanding the situation all the way, It looks to me like you want to use eager here 
https://sequel.jeremyevans.net/rdoc/classes/Sequel/Model/Associations/DatasetMethods.html#method-i-eager

(We it for similar reasons in some complicated lookups in cloud controller, like https://github.com/cloudfoundry/cloud_controller_ng/blob/63fa0c5815ab536aaa2cb5baed13452ca14ced6c/lib/cloud_controller/copilot/sync.rb#L70)

I'll try to take a closer look at the data model here if you think this is a useful direction.

src/bosh-director/lib/bosh/director/agent_broadcaster.rb
```diff
@@ -102,9 +104,10 @@ def sync_dns(instances, blobstore_id, sha1, version)
     end
 
     def filter_instances(vm_cid_to_exclude)
-      Models::Instance
+      Config.db[:instances]
         .inner_join(:vms, Sequel.qualify('vms', 'instance_id') => Sequel.qualify('instances','id'))
         .select_append(Sequel.expr(Sequel.qualify('instances','id')).as(:id))
+        .select_append(Sequel.expr(Sequel.qualify('vms','agent_id')).as(:fetched_agent_id))
         .where { Sequel.expr(Sequel.qualify('vms','active') => true) }
```
voelzmo: This line is what I'm trying to do: specifically add the `agent_id` to an `instance`, without having to load the `vm` for each `instance` with a separate `SELECT * FROM "vms" WHERE (("instance_id" = 3683) AND ("active" IS TRUE)) LIMIT 1`

I'm not sure why the query in place is already using a lot of sequel internals like specific `inner_join`, `qualify`, and `select_append`, it seems pretty much all of it could be done with a single `eager_graph` like `Artist.eager_graph(albums: proc{|ds| ds.where{year > 1990}}).all`

I'll give that a try and see what happens, thanks!

src/bosh-director/lib/bosh/director/dns/dns_version_converger.rb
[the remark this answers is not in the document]
voelzmo: Thanks @charleshansen for the suggestion. I've added a second comment above where the change is maybe a bit clearer. I'll try to switch to `eager_graph` (I think that's necessary because we're filtering on `instance.vm.active`?) and see what happens.

src/bosh-director/lib/bosh/director/agent_broadcaster.rb
voelzmo: @charleshansen ok, I give up. I tried wrestling sequel into what's necessary, but couldn't do it. Here is what I'm unsure how this would work in sequel
* `instance` has a `one_to_many` with `vms`
* we want to achieve that for an `instance` `i.agent_id` *does not lazily do `SELECT * FROM "vms" WHERE (("instance_id" = 3683) AND ("active" IS TRUE)) LIMIT 1`
* At the same time, the above method needs to filter out instances based on fields in the associated `vm`. This is what I was unable to figure out in sequel. Any claused passed to `eager_graph` only filters on what to eagerly load, not what to keep from the list of instances.
  * examples are: a `vm.cid` which gets passed into the method and VMs with `vm.active != true` 

Not sure how/if that is possible in sequel.

charleshansen: I looked into the query with @xtreme-behrouz-soroushian 

The existing query is more complicated than it needs to be, doing sql calculations across multiple tables. It should actually be both faster and simpler to do the filtering in memory instead of using sql queries. Here is our crack at it (all the unit tests pass)
```
    def filter_instances(vm_cid_to_exclude)
      instances_from_db = Models::Instance
        .where(compilation: false)
        .eager(:vms)
        .all

      instances_from_db.select do |instance|
        instance.vms.any? do |vm|
          vm.active && vm.cid != vm_cid_to_exclude
        end
      end
    end
```

One complicating factor: `instance.agent_id` goes through [instance.active_vm](https://github.com/cloudfoundry/bosh/blob/e163565666c03e81e72481c630ecf1491b827039/src/bosh-director/lib/bosh/director/models/instance.rb#L146-L148) which appears to go around anything we could do with eager loading. If `active_vm` used (self.)`vms.find` instead of `Vm.first`, it would work, but it has slightly different caching behavior and I'm not sure if this is a safe change to make.

charleshansen: We looked at it a little more, and tried to fix the `active_vm` reloading behavior (https://github.com/cloudfoundry/bosh/commit/e53000862b4df58c7e7442125be202a967780f83), but this breaks a number of integration tests.

My recommendation at this point is to directly use `instance.vms.find(&:active).agent_id` instead of `instance.agent_id` if you need caching behavior. It might be worth it to add an instance method to Models::Instance, something like `cached_agent_id` or `cached_active_vm`. If this doesn't work, let me know.

voelzmo: > It should actually be both faster and simpler to do the filtering in memory instead of using sql queries.

This might sound stupid, but how did you validate this?

charleshansen: On master, we changed `filter_instances` to the function we wrote above and the unit tests all passed. We got the query slightly wrong originally and the tests caught it, so I'm pretty sure this is ultimately the same function. I personally find it much easier to understand without all the extra sql logic.

As far as validating "faster", we didn't run any performance tests or anything, it was just more of a gut feeling. I've found in optimizing other queries that Ruby in memory is normally just as fast as sql unless you are using really large tables (~million rows) or if you are searching for something and utilizing database indexing. We aren't indexing on the "active" field in the vms, and I've never heard of anything approaching a million VMS in a bosh deployment, so it seems safe to me. 

That said, If you have a performance environment set up, I'd like to know how it fares. You would still need to stop using `instance.agent_id` so that we don't hit the database for every instance. I didn't try running it, so I'm not 100% sure this will fix it.

charleshansen:
@voelzmo thanks for the PR, it seems like a good idea and we ended up just re-implementing it after thinking about it for a bit 6a09e3cd61852c2ca595e8a1ed86934865192411. It is basically the same as your PR. We verified that it now has fewer calls to the database on a small deployment, but did not try anything against 1600 VMs like you were talking about.

Let us know if this looks good to you and if this improves your performance. Thanks!

--

@charleshansen @dpb587-pivotal

voelzmo:
Hey @charleshansen and @dpb587-pivotal, thanks for following up on this!

The change looks great, I'll try to draw some graphs for NATS performance using a bigger setup we have here. We also try to do some even bigger scaling to 2k instances and see if this change makes it possible to deploy at all for us without disabling `sync_dns` after instance update.

jfmyers9:
Hi @voelzmo,

We are going to close this PR as it seems like the above change may work for you all. Please feel free to re-open this PR or an issue if the changes still aren't sufficient. Thanks!

Best,
@jfmyers9  && @belinda-liu, CF BOSH

Issue #2120, opened 2019-01-22
v268.5.0 incorrectly detects configuration change when none occurred


aeijdenberg:
**Describe the bug**

v268.5.0 bosh director seems to always want to update all instances, even when nothing has changed.

**To Reproduce**

A few days ago one of our CI/CD environments updated to the latest `bosh-deployment` which was a bump to `v268.5.0` of bosh director.

Since then, each time our pipeline runs (which amongst other things, deploys v7.0.0 of cf-deployment) we see it update all instances, even when the manifest has not changed.

**Expected behavior**

It to churn for about 2 minutes, determine nothing has changed, and continue (as it did with previous version of bosh director).

**Logs**

Normal output of deploy task:

```
Using deployment 'cf'

Release 'cf-smoke-tests/40.0.44' for stemcell 'ubuntu-xenial/170.19' already exists.
...
Task 465081 | 06:20:52 | Verifying remote release: Verifying remote release (00:00:00)
Task 465081 | 06:20:52 | Extracting release: Extracting release (00:00:00)
Task 465081 | 06:20:52 | Verifying manifest: Verifying manifest (00:00:00)
Task 465081 | 06:20:52 | Resolving package dependencies: Resolving package dependencies (00:00:00)
Task 465081 | 06:20:52 | Processing 2 existing jobs: Processing 2 existing jobs (00:00:00)
Task 465081 | 06:20:52 | Compiled Release has been created: bosh-dns-aliases/0.0.3 (00:00:00)

Task 465081 Started  Thu Jan 17 06:20:51 UTC 2019
Task 465081 Finished Thu Jan 17 06:20:52 UTC 2019
Task 465081 Duration 00:00:01
Task 465081 done

Task 465082

Task 465082 | 06:20:54 | Preparing deployment: Preparing deployment (00:00:32)
Task 465082 | 06:22:14 | Preparing package compilation: Finding packages to compile (00:00:02)

Task 465082 Started  Thu Jan 17 06:20:54 UTC 2019
Task 465082 Finished Thu Jan 17 06:22:16 UTC 2019
Task 465082 Duration 00:01:22
Task 465082 done
```

Current behavior:

```
Using deployment 'cf'

Release 'cf-smoke-tests/40.0.44' for stemcell 'ubuntu-xenial/170.19' already exists.
...
Task 472091 | 00:39:12 | Downloading remote release: Downloading remote release (00:00:01)
[...]

PR #2122, opened 2019-01-23
Ensure JSON is emitted deterministically

aeijdenberg:
So that we don't cause unnecessary instance update churn

It looks like an unintended side-effect of https://github.com/cloudfoundry/bosh/commit/39748765525aa13f084a27f0903c65671f978ff8 is that the `.bosh/links.json` file is now generated in what appears to be an unpredictable order.

This means that each deploy of a manifest will incorrectly think that each job has updated configuration, and thus drain, update config, and start the jobs again, which is inefficient.

Fixes issue https://github.com/cloudfoundry/bosh/issues/2120 (I hope!).

None yet, opening PR early as this is tricky to test (if you don't do much Ruby, which I don't), and probably fairly urgent as v268.5.0 (which inadvertently introduces this issue) was only just released.

No

### DIFF
--- a/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb
+++ b/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb
@@ -117,7 +117,7 @@ module Bosh::Director::Core::Templates
         }
       end
 
-      JSON.pretty_generate(data) + "\n"
+      JSON.pretty_generate(data.sort_by{|e| [e['name'], e['type'], e['group']]}) + "\n"
     end
   end
 end


aeijdenberg:
2nd attempt with correct Ruby syntax appears to work when applied to our bosh director.

Note that the first deploy after updating the director will cause all instances to be updated, but at least after that this part will be stable, and subsequent deploys will be clean.

It may be possible with a more complex patch to preserve the order returned in prior version of bosh director, but I'm not going to attempt that, and it didn't seem to be well defined - rather it seems to be deterministic by accident.

aeijdenberg:
Have verified a 2nd deploy to the same directory of the same manifest now correctly does not attempt to update any of the instances.

h4xnoodle:
Hey @aeijdenberg thanks for taking a look at this and submitting the PR!

Would you mind adding a unit test for this? Let me know if you'd like help with this. 

- Rebecca, BOSH

xtreme-behrouz-soroushian:
@aeijdenberg thanks for the PR.
We might want to add some additional changes because the `link_provider_intents` dataset coming from DB is also non-deterministic in sequence and will cause the map sequence output to generate different SHA. So we will add additional sorts to make sure its deterministic.

/cc @degaurab

jfmyers9:
@aeijdenberg Due to the impact on user experience that this issue has. We are going to go ahead and make the suggested changes above after merging your pull request. Thanks for the submission.

PR #2144, opened 2019-02-22
Scan and fix request containing all missing VMs of deployment, resurrect VMs  in parallel

ansd:

This PR is implementing parallel resurrection:
* Aggregated scan and fix request for all instances of a single deployment having missing VMs (see [tracker story](https://www.pivotaltracker.com/story/show/163648897))
* Resurrect in parallel (see [tracker story](https://www.pivotaltracker.com/story/show/163648915))

see tracker stories

Unit tests and integration tests ran green.
Manually tested that resurrection works on bosh lite.

see stories

Hopefully not: The instance_manager continues to raise all old alerts (i.e. alert per missing VM) in addition to the new deployment aggregated alert. The new alert lacks the alert attributes `job` and `instance_id` (and has the attribute `jobs_to_instance_ids` instead). However, we understand that all plugins can handle these missing attributes.

@friegger @langered @videlov @NautiluX

### DIFF
--- a/src/bosh-director/lib/bosh/director/problem_resolver.rb
+++ b/src/bosh-director/lib/bosh/director/problem_resolver.rb
@@ -31,16 +31,19 @@ module Bosh::Director
 
       begin_stage('Applying problem resolutions', problems.count)
 
-      problems.each do |problem|
-        if problem.state != 'open'
-          reason = "state is '#{problem.state}'"
-          track_and_log("Ignoring problem #{problem.id} (#{reason})")
-        elsif problem.deployment_id != @deployment.id
-          reason = 'not a part of this deployment'
-          track_and_log("Ignoring problem #{problem.id} (#{reason})")
-
-        else
-          apply_resolution(problem)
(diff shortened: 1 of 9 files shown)


ansd:
@BeckerMax and I just noticed that by parallelising [apply_resolutions](https://github.com/cloudfoundry/bosh/blob/03f78682cc9475b0ab3ec3f7c34af57d1cc3f587/src/bosh-director/lib/bosh/director/problem_resolver.rb#L28), when running `bosh cck`, not only are VMs resurrected in parallel, but also disks get reattached in parallel since the CLI sends a single request to the [/problems](https://github.com/cloudfoundry/bosh/blob/03f78682cc9475b0ab3ec3f7c34af57d1cc3f587/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb#L376-L379) endpoint.

That's a side effect we didn't intend to do by this PR. But it also shouldn't harm? Are there cases when we don't want to apply resolutions in parallel when executing `bosh cck`?

Issue #1869, opened 2017-12-28
`attach_disk` failures result in corrupt state


beyhan:
When BOSH creates a disk it also creates the corresponding disk model with the disk CID. https://github.com/cloudfoundry/bosh/blob/dbe2e8a4bb41901b51ae350d73db34d12d6246d5/src/bosh-director/lib/bosh/director/disk_manager.rb#L183-L191 Afterwards BOSH tries to attach the disk to the VM. If it's successful the CPI updates at the end the registry with the disk CID. If it's not successful the registry will not contain a CID but the model will still contain the CID . This is a mismatch which results in the following error when redeploying:
```
Task 30 | 12:09:50 | Deprecation: Ignoring cloud config. Manifest contains 'networks' section.
Task 30 | 12:09:50 | Preparing deployment: Preparing deployment (00:00:00)
Task 30 | 12:09:50 | Preparing package compilation: Finding packages to compile (00:00:00)
Task 30 | 12:09:50 | Updating instance dummy: dummy/4d0ee51a-7881-4719-95fc-1035bebff3a3 (0) (canary) (00:00:05)
                   L Error: 'dummy/4d0ee51a-7881-4719-95fc-1035bebff3a3 (0)' has invalid disks: agent reports '' while director record shows '29d1dced-d008-4301-8c44-385871381751'
Task 30 | 12:09:55 | Error: 'dummy/4d0ee51a-7881-4719-95fc-1035bebff3a3 (0)' has invalid disks: agent reports '' while director record shows '29d1dced-d008-4301-8c44-385871381751'
```

Even if the CPI `attach_disk` call is successful and agent `mount_disk` call fails, we end up in the same situation because `unmount_disk` call is executed. https://github.com/cloudfoundry/bosh/blob/dbe2e8a4bb41901b51ae350d73db34d12d6246d5/src/bosh-director/lib/bosh/director/disk_manager.rb#L137

damzog:
Did you find a way to repair the bosh config?

voelzmo:
@damzog I think we default to calling `bosh delete-vm` followed by `bosh deploy` to fix these issues.

mfine30:
@voelzmo @beyhan it seems like this bug is directly related because of the registry? Since we're working on removing it, do you think this bug will still exist? 

I'm going to close it, but if you believe it will, please re-open.

voelzmo:
[...]

PR #2140, opened 2019-02-15
Update BOSH cck for CPI API V2 and mount the disk in case of reboot option

beyhan:
### What is this change about?

This change adds CPI API V2 support to the BOSH `cck` command.
Also will mount the disk when BOSH `cck` with the option `reattach disk and reboot instance`
is executed.

### Please provide contextual information.

fixes #2139 
fixes #2138 

### What tests have you run against this PR?

Director Unit Tests and all ITs. Also manual tests has been executed.

PR #2146, opened 2019-02-27
Reattach persistent disk if mismatch

ansd:
If the agent reports no disk, but the director model includes a disk,
re-attach it.

see [#164276966](https://www.pivotaltracker.com/story/show/164276966) and https://github.com/cloudfoundry/bosh/issues/1869 for context

Director unit tests and integration tests.

Nope.

@friegger @beyhan @langered

### DIFF
--- a/src/bosh-director/lib/bosh/director/disk_manager.rb
+++ b/src/bosh-director/lib/bosh/director/disk_manager.rb
@@ -10,7 +10,7 @@ module Bosh::Director
     def update_persistent_disk(instance_plan)
       @logger.info('Updating persistent disk')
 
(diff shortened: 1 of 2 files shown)


beyhan:
@dpb587-pivotal and @jfmyers9 this is the PR for attach disk issue we talked about. Also the PR https://github.com/cloudfoundry/bosh/pull/2140 for `bosh cck` is related.

jfmyers9:
@beyhan We've scheduled both of these PRs to be reviewed soon. Thanks for the submission.

friegger:
@jfmyers9 The PR for cck has already been merged, when are you going to have a look at this one. Asking since they are so related and context should be similar.

PR #2147, opened 2019-02-28
Speedup performance of job template rendering

s4heid:

Deploying a large number of instances comes with long rendering times of job templates. The reason behind the poor performance can be tracked down to the frequent initialization of the `EvaluationContext` for each template combined with frequent deep copies of the spec.

We discovered two hotspots in the [`JobTemplateRenderer`](https://github.com/cloudfoundry/bosh/blob/641ea9493cd4e32361d8496457f8cb33afc2175e/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb), which become very expensive with a growing number of instances. Those are the initialization of [`Bosh::Template::EvaluationContext`](https://github.com/cloudfoundry/bosh/blob/8fbf0457d48fbf6ab1223ef5e9134585cbf43f57/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb#L50) and the frequent usage of `Bosh::Common::DeepCopy.copy()` on the spec.

According to our benchmark tests we estimate that this change will speed up the job template rendering by factor ~3. Disclaimer: this has only been tested in a VBox environment with 1500 Zookeeper instances. 

[Tracker story #163459423](https://www.pivotaltracker.com/story/show/163459423)

We ran unit and integration tests on director concourse pipeline.

Improve job template rendering performance when deploying. 

No

@friegger @voelzmo @mfine30 @KaiHofstetter @NautiluX @BeckerMax

### DIFF
--- a/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb
+++ b/src/bosh-director-core/lib/bosh/director/core/templates/job_template_renderer.rb
@@ -30,6 +30,8 @@ module Bosh::Director::Core::Templates
     end
 
     def render(spec)
+      spec = Bosh::Common::DeepCopy.copy(spec)
+
       if spec['properties_need_filtering']
         spec = remove_unused_properties(spec)
       end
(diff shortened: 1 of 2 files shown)


NautiluX:
@dpb587 We ended up not changing the behavior of the current implementation. Templates still render in isolation, meaning if one template changes the spec then the next one gets the unmodified version.

dpb587-pivotal:
Hi - we were taking a look at the latest changes and had a couple thoughts.

First, is there a reason that `EvaluationContext.==` does not need to consider if `@properties` has changed? Seems like we check all the other publicly-accessible values. Semi-related, any way we can ensure we don't introduce new properties and then forget to include them in our custom `==` logic?

Second, can you add a test for the `EvaluationContext.==` change to verify the behavior is working as expected? Want to make sure that we don't regress on the behavior in the future.

@dpb587-pivotal && @charleshansen

NautiluX:
Hi @dpb587-pivotal, thanks for your comment!

Unfortunately, I think it's common in ruby to implement == yourself if you need it for your class and need to keep its existence in mind when adding properties to a class. Maybe we could do something like `shared_examples` to make sure the interface doesn't change unnoticed, but I think this only works the other way round: You will notice when a function is removed.

We didn't consider adding tests for == as the behaviour of the whole thing shouldn't change, but I think you're right, it's part of the public interface of `EvaluationContext`. We will add those.

And yes - I think we just overlooked `@properties`.

PR #2149, opened 2019-03-01
Enhancement for encryption of persistent disk

zts-pash:
For redeployment of release with an encrypted disk partition, the file
system is mounted differently. Therefore agent disk cid will be out of
sync with cid stored with director.
This change bypasses that check only if it finds that there is an encryption
package available to enable successful redeployment.
This change will not affect any other scenatio except the one mentioned above.

Co-authored-by: Deepak Shetty <dshetty@zettaset.com>
Co-authored-by: Brian Metzger <bmetzger@zettaset.com>

For an encrypted storage partition, the mount convention is different.

We were seeing issue during redeployment of the release which does encryption of the persistent disk and then mounts it. Therefore agent disk cid will be out of sync with cid stored with director. This bypasses that check to enable successful redeployment.

We are implementing encryption of persistent storage which requires the filesystem to be mounted differently

N/A

N/A

We deployed an addon to allow encryption of persistent storage. We ran a sample application with encrypted partition and redeployed with encrypted partition.

Don't track cid if the persistent disk is encrypted

_Something brief that conveys the change and is written with the Operator audience in mind. See [previous release notes](https://github.com/cloudfoundry/bosh/releases) for examples._
No operator intervention is required. Encryption will be implemented automatically once the proper addon is installed.

N/A

No

@zts-bmetzger @zts-myankovskiy @zts-dshetty

### DIFF
--- a/src/bosh-director/lib/bosh/director/disk_manager.rb
+++ b/src/bosh-director/lib/bosh/director/disk_manager.rb
@@ -10,7 +10,10 @@ module Bosh::Director
     def update_persistent_disk(instance_plan)
       @logger.info('Updating persistent disk')
 
-      check_persistent_disk(instance_plan) unless has_multiple_persistent_disks?(instance_plan)
+      current_packages = instance_plan.instance.current_packages.to_s
+      if !current_packages["xcrypt"]
+        check_persistent_disk(instance_plan) unless has_multiple_persistent_disks?(instance_plan)
+      end
 
       return unless instance_plan.persistent_disk_changed?
 


mrosecrance:
Changing the agent Disk CID without updating the record in the Director means that the Director can no longer send disk changes to the agent. I wonder if getting these CID's to match is a better solution than skipping the checking altogether. Without them matching none of the behavior [here](https://bosh.io/docs/persistent-disks/#changing-persistent-disk) is supported although we'd need to think about what data migration on an encrypted disk looks like. Currently it seems like recreates of the VM will lose track of the disk.

Is the desired behavior here that the bosh director do nothing with the disk and it's managed completely by the operator?

As for comments on the code, test's would be great. Also nesting a not if in a not if is not very elegant. Could it be rewritten as `if !current_packages["xcrypt"] || !has_multiple_persistent_disks?(instance_plan)`

dpb587-pivotal:
Closing - we discussed this in a call with the following suggestions. Future changes along these lines should result in different PRs in other repositories.

1. This change is not generic enough for reuse outside of a single implementation.
1. The underlying issue was from bosh-agent reading `mount` for the `/var/vcap/store` and not recognizing the `/dev/mapper...` path, reporting the disk mismatch errors. This change doesn't address that underlying resource modeling issue, so would likely occur in other places.
1. This should be investigated and probably addressed at the bosh-agent level to try and make it aware of more complex device mapping scenarios. One area of investigation would be to have agent try to resolve device paths (virtual `/dev/mapper...` or physical `/dev/xvdf`), potentially through `lsblk` or `dmsetup` commands.
1. We suggested experimenting with bosh-agent changes and running it on a local environment to test if this will work before sending a pull request for us to integrate into the regular stemcells.
1. This change was a surprise to use, so we'd definitely appreciate more up front discussion around broad changes like this. It will hopefully help save both of our teams time to have more of a plan.
1. There will likely be additional discussions that are necessary because there are some underlying job prioritization which will eventually be required (i.e. to ensure the disk-encryption jobs are starting before other disk-needing jobs). This is broad-reaching change, so we'll definitely want to chat about this.

zts-pash:
Yes, as discussed in the call, we will be opening a different pull request soon for bosh-agent repository in which we will address this particular issue of agent going out of sync with director after disk encryption causing redeployment to fail.

PR #2152, opened 2019-03-07
Add unit tests for EvaluationContext.== operator

BeckerMax:

* Also compare the `properties` member
* Add tests that modify member variables
* Add test to ensure new public properties are noticed when they are
added. The test acts as a reminder to not forget to add new properties
to the == operator

This is a Follow-up to https://github.com/cloudfoundry/bosh/pull/2147
See also [tracker story](https://www.pivotaltracker.com/story/show/163459423)

bosh unit and integration tests

No

@NautiluX @dpb587-pivotal

### DIFF
--- a/src/bosh-template/spec/unit/bosh/template/evaluation_context_spec.rb
+++ b/src/bosh-template/spec/unit/bosh/template/evaluation_context_spec.rb
@@ -2,6 +2,7 @@ require 'spec_helper'
 require 'bosh/template/evaluation_context'
 require 'bosh/template/evaluation_link_instance'
 require 'bosh/template/evaluation_link'
+require 'common/deep_copy'
 
 module Bosh
   module Template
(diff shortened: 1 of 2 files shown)


src/bosh-template/spec/unit/bosh/template/evaluation_context_spec.rb
```diff
@@ -109,7 +110,86 @@ def eval_template(erb, context)
...
+          using MakeIndexAccessible
+
+          it 'returns false' do
+            evaluation_context.modify_index
+            expect(evaluation_context.index).to equal(42)
+            expect(evaluation_context == other_evaluation_context).to equal(false)
+          end
+        end
+
+        context 'when instance variables are modified' do
```
BeckerMax: @dpb587-pivotal @charleshansen 
We implemented a [test|] which will check whether a new member is added to the EvaluationContext class which is not yet part of `EvaluationContext.==`. With this kind of test, members which are intentionally not part of `EvaluationContext.==` need to be listed explicitly in the test. We think the test provides additional value but are unsure if it is worth the added complexity. What is your opinion?

dpb587-pivotal: Verbose, but dynamically defensive; clever. I'm okay with it.

PR #2157, opened 2019-03-13
Spike: Parallel resurrection respects `serial` and `max_in_flight`

ansd:
**Do not merge this PR. This PR is for discussion only.**

This is a follow up to https://github.com/cloudfoundry/bosh/pull/2144.

The question is:
Should parallel resurrection respect the `update` properties:
* `serial` ? (see 1st commit)
* `max_in_flight`? (see 2nd commit)

With this PR, both properties are respected. In [#163648935](https://www.pivotaltracker.com/story/show/163648935), we describe a test setup with expected behavior.

If we decide, it makes sense to respect one or both of these properties, we close this PR and re-implement it.

cc @beyhan @KaiHofstetter @friegger

### DIFF
--- a/src/bosh-director/lib/bosh/director/problem_resolver.rb
+++ b/src/bosh-director/lib/bosh/director/problem_resolver.rb
@@ -1,5 +1,6 @@
 module Bosh::Director
   class ProblemResolver
+    include LegacyDeploymentHelper
 
     attr_reader :logger
 
@@ -31,17 +32,21 @@ module Bosh::Director
 
       begin_stage('Applying problem resolutions', problems.count)
 
-      ThreadPool.new(max_threads: Config.max_threads).wrap do |pool|
-        problems.each do |problem|
-          pool.process do
-            if problem.state != 'open'
-              reason = "state is '#{problem.state}'"
-              track_and_log("Ignoring problem #{problem.id} (#{reason})")
-            elsif problem.deployment_id != @deployment.id
-              reason = 'not a part of this deployment'
-              track_and_log("Ignoring problem #{problem.id} (#{reason})")
-            else
-              apply_resolution(problem)
+      # TODO: here we just assume :type recreate VM,
+      # make sure that other problem types like re-attach disk still work
+      parallelizable_problems_for_each(problems.all) do |probs|
+        ThreadPool.new(max_threads: Config.max_threads).wrap do |pool|
+          probs.each do |problem|
+            pool.process do
+              if problem.state != 'open'
+                reason = "state is '#{problem.state}'"
+                track_and_log("Ignoring problem #{problem.id} (#{reason})")
+              elsif problem.deployment_id != @deployment.id
+                reason = 'not a part of this deployment'
+                track_and_log("Ignoring problem #{problem.id} (#{reason})")
+              else
+                apply_resolution(problem)
+              end
             end
           end
         end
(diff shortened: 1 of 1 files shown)


xtreme-behrouz-soroushian:
This changes lead to failure of some unit tests. Also the PR itself don't add any test (unit test or integration test) pertinent to the new functionality. Can you please consider checking the failing tests and propose fixes for them? Also it will be appreciated if you could add appropriate tests for coverage of the new feature if needed?

belinda-liu:
Hi @ansd,

@charleshansen and I spent some time talking about this. We think that it makes sense for parallel resurrection to respect `serial` and `max_in_flight`; it seems reasonable that in certain scenarios, the number of rolling VMs should use the same configuration as a deployed update. 
That said, operators may want to configure a larger `max_in_flight` for resurrection (ex. all instances are down and they want to recover as quickly as possible). It would be nice to have this as a configurable property in the resurrection config that would override the property in the manifest if set. We're not sure what a current resurrection config looks like, is this possible? 
Other than that, this sounds pretty good to us, not sure if other folks have more opinions cc @mfine30 .

voelzmo:
Hey @belinda-liu and @charleshansen, 

Thanks for the reply and for looking at this change! My feeling is that figuring out the correct behavior for `max_in_flight` and `serial` on resurrection is difficult, as it seems to depend on what is currently happening:
* The VMs might really be lost, therefore you want to recover as quickly as possible. This is somehow similar to a first deployment, where you don't care about `max_in_flight` and (maybe also not) `serial` and just want to get stuff running
* The VMs might still be running and the processes on them doing their thing (e.g. Diego cells still run the applications on them), but due to whatever reason, heartbeats aren't created or cannot reach the HM. This is similar to an update deployment, which cares about `max_in_flight` and `serial`, as you need to plan for enough capacity to migrate the customer payload, such that they don't experience a downtime.

I'm not sure if I'm missing other cases in which you may or may not care about these two settings, please let me know if you have additional ones. Just by looking at those two, I'm struggling with the value of adding an extra property in the [resurrection config](https://bosh.io/docs/resurrector/#enable-with-resurrection-config), as this might be a choice that's very hard to make in an informed way for operators. If BOSH cannot decide which of the two scenarios above is really happening, it is probably a good idea to err on the safe side of things and respect `max_in_flight` and `serial` as defined in the manifest. Do we have indications that this would not meet operator's expectations?

belinda-liu:
Hey @voelzmo,

Good point, adding an extra property in the resurrection config might have limited value to operators once their VMs are already down. We were thinking that it would allow operators to potentially choose a slightly higher `max_in_flight` value as a middle ground between the two cases you described. However, it sounds good to respect the deployed `max_in_flight` and `serial` values for now as it seems like the safer option. We can always add resurrection config later if it would be helpful.
We also chatted with Morgan and a few other folks on our side and everyone seems on board with respecting these properties. Were there any other discussion points you wanted to cover?

@belinda-liu && @charleshansen

friegger:
@belinda-liu @charleshansen @mfine30 Thank you. No, we don't have additional points for discussion. Based on this, we'll provide a production-ready implementation.

jfmyers9:
Hi @friegger,

We are going to close this pull request in the mean time. Please open a new pull request with the production implementation and reference this PR. Thanks!

Best,

@jfmyers9 && @belinda-liu, CF BOSH

voelzmo:
Thanks @belinda-liu and @charleshansen for replying! Seems like I missed this during the summit.

PR #2162, opened 2019-03-19
Use FactoryBot to create DeploymentPlan Factories

charleshansen:

Create Factories for deployment plan objects in test

Tests that require DeplymentPlan objects can have hundreds of lines of setup just to create the objects, we should use factories. Machinist won't work with DeploymentPlan objects, but FactoryBot does (because of initialize_with). To discuss further perhaps in tech retro.

Unit tests pass

Testing change only, no notes

No


mrosecrance:
Didn't you already merge this in? Can we delete the pr?

jfmyers9:
Yep this change appears to already exist on the master branch. Going to close this PR.

jfmyers9:
@charleshansen Can you delete the branch if it is no longer in use?

charleshansen:
Actually still using the branch, but good to close the PR

PR #2184, opened 2019-05-27
Respect serial and max_in_flight property

s4heid:

Before this change, the resolution of detected problems by a scan and fix task was scheduled in parallel across all instance groups in a deployment. After this change, `max_in_flight` and `serial` are considered.

**Question**: Why does the regular deploy flow [not consider](https://github.com/cloudfoundry/bosh/blob/c73c7e6636a895b6315739ecad18fd8fc2228b56/src/bosh-director/lib/bosh/director/deployment_plan/multi_instance_group_updater.rb#L29) the [`max_threads`](https://bosh.io/jobs/director?source=github.com/cloudfoundry/bosh#p%3ddirector.max_threads) property during the update instance step although it is considered in the create vm step, c.f. [step_executor](https://github.com/cloudfoundry/bosh/blob/c73c7e6636a895b6315739ecad18fd8fc2228b56/src/bosh-director/lib/bosh/director/step_executor.rb#L15).

In this implementation, resurrection of instances also considers `max_threads` from the director config for the update instance step. This is the only difference between resurrection and the standard deploy flow and it would be nice if both would be aligned.

[#165062729](https://www.pivotaltracker.com/story/show/165062729) and discussion in https://github.com/cloudfoundry/bosh/pull/2157#issuecomment-476132012

Unit and integration tests.

No.

@friegger, @mfine30, @beyhan

### DIFF
--- a/src/bosh-director/lib/bosh/director/problem_resolver.rb
+++ b/src/bosh-director/lib/bosh/director/problem_resolver.rb
@@ -1,5 +1,9 @@
+require_relative 'jobs/update_deployment'
+
 module Bosh::Director
   class ProblemResolver
+    include DeploymentPlan
+    include Jobs
 
     attr_reader :logger
 
@@ -7,8 +11,12 @@ module Bosh::Director
       @deployment = deployment
       @resolved_count = 0
       @resolution_error_logs = StringIO.new
-
-      #temp
+      update_deployment = UpdateDeployment.new(
+        @deployment.manifest,
+        @deployment.cloud_configs.map(&:id),
+        @deployment.runtime_configs.map(&:id),
+      )
+      @instance_groups = update_deployment.parse_manifest.instance_groups
       @event_log_stage = nil
       @logger = Config.logger
     end
@@ -27,20 +35,20 @@ module Bosh::Director
 
     def apply_resolutions(resolutions)
       @resolutions = resolutions
-      problems = Models::DeploymentProblem.where(id: resolutions.keys)
+      all_problems = Models::DeploymentProblem.where(id: resolutions.keys, deployment_id: @deployment.id).all
+      begin_stage('Applying problem resolutions', all_problems.size)
 
-      begin_stage('Applying problem resolutions', problems.count)
+      if Config.parallel_problem_resolution && all_problems.size > 1
+        ig_to_problems = problems_by_instance_group(all_problems)
 
-      if Config.parallel_problem_resolution
-        ThreadPool.new(max_threads: Config.max_threads).wrap do |pool|
-          problems.each do |problem|
-            pool.process do
-              process_problem(problem)
-            end
+        problems_serially_ordered_by_job(ig_to_problems) do |ig_problems, max_in_flight|
+          n_threads = [ig_problems.size, max_in_flight, Config.max_threads].min
+          parallel_each(n_threads, ig_problems) do |problem|
+            process_problem(problem)
           end
         end
       else
-        problems.each do |problem|
+        all_problems.each do |problem|
           process_problem(problem)
         end
       end
@@ -52,16 +60,65 @@ module Bosh::Director
 
     private
 
-    def process_problem(problem)
-      if problem.state != 'open'
-        reason = "state is '#{problem.state}'"
-        track_and_log("Ignoring problem #{problem.id} (#{reason})")
-      elsif problem.deployment_id != @deployment.id
-        reason = 'not a part of this deployment'
-        track_and_log("Ignoring problem #{problem.id} (#{reason})")
+    def parallel_each(n_threads, ary)
+      if n_threads > 1
+        ThreadPool.new(max_threads: n_threads).wrap do |pool|
+          ary.each do |entry|
+            pool.process do
+              yield entry
+            end
+          end
+        end
       else
-        apply_resolution(problem)
+        ary.each do |entry|
+          yield entry
+        end
       end
+    end
+
+    def process_problem(problem)
+      if problem.open?
+        apply_resolution(problem)
+      else
+        track_and_log("Ignoring problem #{problem.id} (state is '#{problem.state}')")
+      end
+    end
+
+    def problems_serially_ordered_by_job(ig_to_problems, &block)
+      BatchMultiInstanceGroupUpdater.partition_jobs_by_serial(@instance_groups).each do |jp|
+        igs_with_problems = []
+        jp.each { |ig| igs_with_problems << ig.name if ig_to_problems.key?(ig.name) }
+        n_threads = [igs_with_problems.size, Config.max_threads].min
+        parallel_each(jp.first.update.serial? ? 1 : n_threads, igs_with_problems) do |ig_name|
+          process_ig(ig_name, ig_to_problems[ig_name], block)
+        end
+      end
+    end
+
+    def process_ig(ig_name, problems, block)
+      instance_group = @instance_groups.find do |plan_ig|
+        plan_ig.name == ig_name
+      end
+      max_in_flight = instance_group.update.max_in_flight(problems.size)
+      block.call(problems, max_in_flight)
+    end
+
+    def problems_by_instance_group(problems)
+      instance_group_to_problems = {}
+      problems.each do |p|
+        begin
+          if p.instance_problem?
+            instance = Models::Instance.where(id: p.resource_id).first
+          else
+            disk = Models::PersistentDisk.where(id: p.resource_id).first
+            instance = Models::Instance.where(id: disk.instance_id).first if disk
+          end
+          (instance_group_to_problems[instance.job] ||= []) << p if instance
+        rescue StandardError => e
+          log_resolution_error(p, e)
+        end
+      end
+      instance_group_to_problems
     end
 
     def apply_resolution(problem)
@@ -84,8 +141,7 @@ module Bosh::Director
       problem.state = 'resolved'
       problem.save
       @resolved_count += 1
-
-    rescue => e
+    rescue StandardError => e
       log_resolution_error(problem, e)
     end
 
--- a/src/bosh-director/lib/bosh/director/problem_resolver.rb
+++ b/src/bosh-director/lib/bosh/director/problem_resolver.rb
@@ -40,7 +40,7 @@ module Bosh::Director
 
       if Config.parallel_problem_resolution && all_problems.size > 1
         igs_to_problems = problems_by_instance_group(all_problems)
-        partitions_to_problem_igs = filter_instance_groups(igs_to_problems)
+        partitions_to_problem_igs = partition_instance_groups_by_serial(igs_to_problems.keys)
         process_partitions(partitions_to_problem_igs, igs_to_problems)
       else
         all_problems.each do |problem|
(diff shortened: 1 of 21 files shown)


src/bosh-director/lib/bosh/director/problem_resolver.rb
```diff
@@ -27,20 +35,20 @@ def track_and_log(task, log = true)
...
-      if Config.parallel_problem_resolution
-        ThreadPool.new(max_threads: Config.max_threads).wrap do |pool|
-          problems.each do |problem|
-            pool.process do
-              process_problem(problem)
-            end
+        problems_serially_ordered_by_job(ig_to_problems) do |ig_problems, max_in_flight|
+          n_threads = [ig_problems.size, max_in_flight, Config.max_threads].min
+          parallel_each(n_threads, ig_problems) do |problem|
+            process_problem(problem)
```
miguelverissimo: It feels confusing having`process_problem` being called in `process_ig`, which is being called by `problems_serially_ordered_by_job`. Cant this be refactored so that the ordering of the jobs, and their processing is done separately?
It seems that `apply_resolutions` should be the method calling the above methods.

src/bosh-director/lib/bosh/director/problem_resolver.rb
```diff
@@ -52,16 +60,65 @@ def apply_resolutions(resolutions)
...
+    end
+
+    def problems_serially_ordered_by_job(ig_to_problems, &block)
+      BatchMultiInstanceGroupUpdater.partition_jobs_by_serial(@instance_groups).each do |jp|
+        igs_with_problems = []
+        jp.each { |ig| igs_with_problems << ig.name if ig_to_problems.key?(ig.name) }
+        n_threads = [igs_with_problems.size, Config.max_threads].min
+        parallel_each(jp.first.update.serial? ? 1 : n_threads, igs_with_problems) do |ig_name|
+          process_ig(ig_name, ig_to_problems[ig_name], block)
+        end
```
miguelverissimo: This block feels very hard to read, what does `jp` represent? There's plenty of work being done within the iterators in this block. For the sake of readability, do you mind refactoring the variable names and expanding the blocks?

src/bosh-director/spec/unit/problem_resolver_spec.rb
```diff
@@ -9,12 +9,31 @@ module Bosh::Director
     let(:task) { Bosh::Director::Models::Task.make(id: 42, username: 'user') }
     let(:task_writer) { Bosh::Director::TaskDBWriter.new(:event_output, task.id) }
     let(:event_log) { Bosh::Director::EventLog::Log.new(task_writer) }
+    let(:update_deployment) { double }
+    let(:parallel_problem_resolution) { true }
+    let(:parallel_update_config) { instance_double('Bosh::Director::DeploymentPlan::UpdateConfig', serial?: false) }
+    let(:n_igs_with_problems) { 4 }
+    let(:n_problems_in_ig) { 3 }
```
belinda-liu: It's not obvious what `n_problems_in_ig` means, and it's easily confused with the above variable `n_igs_with_problems`. Would you consider renaming to something like `problems_per_instance_group` and `num_problem_instance_groups `?

src/bosh-director/spec/unit/problem_resolver_spec.rb
```diff
@@ -31,96 +50,201 @@ def inactive_disk(id, deployment_id = nil)
...
-          problems = []
+          context 'when max_in_flight is one' do
+            let(:max_in_flight) { 1 }
 
-          agent = double('agent')
-          expect(agent).to receive(:list_disk).and_return([])
-          expect(cloud).to receive(:detach_disk).exactly(1).times
+            it 'does not create a threadpool for processing the problem' do
+              test_instance_apply_resolutions
+              expect(ThreadPool).to have_received(:new).once.with(max_threads: n_igs_with_problems)
```
belinda-liu: I think this test description might be off, it seems like the code is asserting that ThreadPool is actually being created.

Also, shouldn't `ThreadPool` receive a new with a max_threads of what `max_threads` is set to? Why is the number of max_threads it's initialized with the number of instance groups with problems?

src/bosh-director/spec/unit/problem_resolver_spec.rb
```diff
@@ -31,96 +50,201 @@ def inactive_disk(id, deployment_id = nil)
...
+            end
 
-          expect(resolver.apply_resolutions(problems[0].id.to_s => 'delete_disk', problems[1].id.to_s => 'ignore'))
-            .to eq([2, nil])
+            it 'respects serial' do
+              test_instance_apply_resolutions
+              non_serial_igs_with_problems = 2
+              expect(ThreadPool).to have_received(:new).once.with(max_threads: non_serial_igs_with_problems)
+              expect(ThreadPool).to have_received(:new).exactly(
+                n_igs_with_problems,
```
belinda-liu: `n_igs_with_problems` is currently set to 1 in this context (line 89) but my understanding is that `ig-2` and `ig-3` would both need a ThreadPool, which means that shouldn't this be called twice? Do ThreadPools get reused across instance groups with parallel updates?

belinda-liu:
@s4heid as to your question about the max_threads setting in the ParallelMultiInstanceGroupUpdater: it seems like that has been set to `jobs.size` since it was introduced. My best guess is that it was intended as an optimization; if you had fewer jobs than `max_threads` then there was no need to spin up that many, and if there were more then perhaps operators didn't want to be limited by `max_threads`, which may be kind of an obscure setting. I don't have the full historical context behind this though.

src/bosh-director/spec/unit/problem_resolver_spec.rb
langered: Hi @belinda-liu ,

ThreadPools are getting reused across instance groups which should get updated in parallel.

Therefore, `ig-2` and `ig-3` will share one ThreadPool because they are configured with `serial = false`. This ThreadPool is created once and will use 2 threads to handle those instance groups in parallel.
This is what the first `expect` is testing. It checks if a ThreadPool is created once with `max_thread: 2`.

The second `expect` is testing the parallel resolution of problems inside a single instance group. Each instance group (4 in total) has 3 problems which should all get resolved in parallel. Therefore, we expect to create 4 ThreadPools with each having `max_thread: 3`.

`n_igs_with_problems` should be the default (4) again because line 89 is in another context.

src/bosh-director/spec/unit/problem_resolver_spec.rb
langered: Thanks for pointing this out! You are right, actually one ThreadPool should get created. But only the one which handles instance groups in parallel, therefore the number of `max_threads` is set to the number of instance_groups.

src/spec/gocli/integration/health_monitor/resurrector_spec.rb
```diff
@@ -59,4 +59,74 @@
...
+
+    director.wait_for_resurrection_to_finish
+    resurrection_task = director.tasks.filter(
+      username: 'hm',
+      description: 'scan and fix',
+      state: 'done',
+    ).order(:id).first
+
+    expect(resurrection_task).to be_truthy
+    expect(resurrection_task[:event_output]).to match(/Applying.*started.*\n.*Applying.*started.*\n.*Applying.*finished/m)
```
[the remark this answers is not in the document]
langered: You are right, `.any?` makes the assertion more readable.

To the second expect:
We want to make sure, that only the first two instances get recreated in parallel since `max_in_flight` is set to 2.  The next message should include `finished` to ensure that the third one did not start before. After that point in time, we cannot make assumption about the ordering of messages.

langered:
Hi @miguelverissimo,

regarding pull `update_deployment` into the `problem_resolver.rb`:

I agree with you that it's heavy to load the full `update_deployment` into the `problem_resolver`. We had an approach with a similar strategy where we used a lightweight `DeploymentConfig` instead of loading the manifest from the `DeploymentPlan`. However, we had to rewrite existing logic from the `BatchMultiInstanceGroupUpdater.partition_jobs_by_serial`, which resulted in duplicated code.

After some discussion in the team we decided to go with the approach of this PR and reuse code of the `DeploymentPlan`.

Best Regards,
@langered

/cc @ansd, @FlorianNachtigall, @s4heid

src/bosh-director/lib/bosh/director/problem_resolver.rb
```diff
@@ -52,18 +55,80 @@ def apply_resolutions(resolutions)
...
+      else
+        track_and_log("Ignoring problem #{problem.id} (state is '#{problem.state}')")
       end
     end
 
+    def filter_instance_groups(igs_to_problems)
+      ig_partition_to_ig_names_with_problems = {}
+      BatchMultiInstanceGroupUpdater.partition_jobs_by_serial(@instance_groups).each do |ig_partition|
+        igs_with_problems = []
+        ig_partition.each { |ig| igs_with_problems << ig.name if igs_to_problems.key?(ig.name) }
```
charleshansen: A minor nit, but these blocks are harder to read when using `each` instead of the more idiomatic ruby array methods like `map` or `select`/`reject` or even `reduce`. For instance, theses two lines could be replaced with
```rb
igs_with_problems = ig_partition.map(&:name).select { |ig_name| igs_to_problems.key?(ig_name) }
```
This is much more readable to me. There are similar patterns throughout the PR (most of the occurrences of `each`). I wouldn't want to block the PR on it, but would still appreciate if you could make them more idiomatic ruby.

mrosecrance:
It turns out it should be relatively simple to avoid using `update_deployment`, starting from a deployment model, you can get a deployment plan and the instance groups with the following code stolen and modified [from cck](https://github.com/cloudfoundry/bosh/blob/e814033e0fb9017019fc0568f960ce1df5456ee6/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb#L63)

```rb
deployment_plan = DeploymentPlan::PlannerFactory.create(Config.logger).create_from_model(deployment_model)
instance_groups = deployment_plan.instance_groups
```

mrosecrance:
`Config.max_threads` is something the team has some ambiguity on. There's a risk that operators will be confused when we ignore `max_in_flight` in favor of a usually unconfigured value (it's defaulted to 32) and with the multiple nested `parrallel_each` it may get more confusing. It's also been mainly used to throttle director workloads when the workload may crash the director. It doesn't seem like resurrection should fall into this case.

There's room for a more comprehensive discussion on the value and use of `max_threads` but this does not seem like the right time for it.

src/bosh-director/lib/bosh/director/problem_resolver.rb
friegger: It also feels much cleaner to have a proper return value via `map` instead of filling an array via a side effect from within a loop.

friegger:
> `Config.max_threads` is something the team has some ambiguity on. There's a risk that operators will be confused when we ignore `max_in_flight` in favor of a usually unconfigured value (it's defaulted to 32) and with the multiple nested `parrallel_each` it may get more confusing. It's also been mainly used to throttle director workloads when the workload may crash the director. It doesn't seem like resurrection should fall into this case.
> 
> There's room for a more comprehensive discussion on the value and use of `max_threads` but this does not seem like the right time for it.

Yes, for this PR I feel it should just be consistent with the deploy flow.

beyhan:
We applied all change requests. From our side this is ready for merge.

charleshansen:
Looked at the PR after the changes, it looks pretty good now. There was a minor fix and change I wanted so I ended up merging manually.

For more details on my change:

I removed the db truncation before test. It felt dangerous/brittle to rely on a specific index in the database and I didn't want to start a pattern where we need it. It wasn't clear to me why we actually needed it in the test where you were using it and everything passed when I took it out, so I thought it was safe to do. Let me know if I'm missing something and we actually need it.

charleshansen:
Followup, it turns out the tests do fail without truncate_before_test, but not with the sqlite unit tests which run by default. I'm still not thrilled about it, but I put it back in for now.

friegger:
> Followup, it turns out the tests do fail without truncate_before_test, but not with the sqlite unit tests which run by default. I'm still not thrilled about it, but I put it back in for now.

Yes, we had a short discussion around this in our daily and don't like to rely on a specific index here. We'll follow-up on this.

charleshansen:
closing since it is merged, we can talk about the db truncation later

PR #2188, opened 2019-06-07
Backfill tests for UpdateRelease#process_packages

jaresty:
This method is one of the most complex in the class. It is currently
public, but currently it is tested mostly via other methods in the file.
This commit backfills explicit tests for the current behavior of this
method.


charleshansen:
Just looking at it quickly, the method you are testing really shouldn't be public, maybe that is why there are no tests? I would only add tests for `process_packages` if you think the behavior is not tested by the tests for `process_release`. Alternatively, if you were planning on pulling out a collaborator or something, this would be an ok stepping stone.

jaresty:
@charleshansen That's exactly the plan -- I don't believe it is completely tested by the rest of the specs so adding these tests is intended to be a stepping stone to breaking up the method and possibly pulling out a collaborator without changing the existing functionality since I'm not confident that the the rest of the tests cover it well.  I'd expect that some of this coverage could be subsequently replaced in a better way.

jaresty:
I added an example of one possible collaborator extraction that still passes the existing tests (including the new ones).

charleshansen:
Had time to look at the tests in more detail. The text of the `describe` and `it` is welcome. I like the more explicit granularity of use cases. I think when the whole refactor is done, it will be much easier to understand how the packages are processed by reading the tests. 

On the other hand, the actual implementation of the tests relies heavily on mocking the subject itself and `instance_variable_set`, both of which are huge red flags to me. This means that the internal state of the subject is somehow part of it's api, and now the tests don't really tell me how to use the thing because a real consumer wouldn't be able to do this. I know it is an ask, but I'd rather see the whole thing extracted at once so that you don't need to touch private things just to write the tests.

As a random aside, I'd love your extracted collaborator to have a single `process` class method that returns a list of packages instead of manipulating a bunch of instance variables.

jaresty:
This feedback makes sense to me -- I can definitely work on extracting the whole thing (tests included). Could you explain a bit more what you mean here?

> As a random aside, I'd love your extracted collaborator to have a single process class method that returns a list of packages instead of manipulating a bunch of instance variables.

I'm happy to discuss in a call too if it would be easier to convey.

charleshansen:
Great. To expand on the random aside:

I've found in the bosh code base that instance variables are a real problem. Developers who did not create the instance variables originally think that they are part of the API of whatever class they are in and start manipulating them. I would much prefer that instance variables are only used when there is legitimately state that exists in between function calls. If you need to share data between private methods, just pass the data as arguments to the private methods. 

In this particular case, you never call any method more than once, so there shouldn't be any state.  You always just call `PackageProcesses.new(args).process`, it would be clearer that `process` is a one-time thing if you made `process` a class method and just called `PackageProcesses.process(args)`. This way when I am reading the code 4 months from now, I don't have to worry that the internal state is something surprising. 

example code I'm trying to prevent in the future, and generally not burn cognitive overhead on:
```rb
processor = PackageProcesses.new(args)
# ...
# in some other function, who knows where 
processor.manifest_packages << extra_mystery_packages
# ...
processor.process
```

jaresty:
Got it. I'll give that a shot and see how it goes.

jaresty:
@charleshansen Extracted the new file so now the tests don't need to interact with internal state of the subject.

jaresty:
In re-reading your feedback I understood that you were also asking me to eliminate state completely; i.e. to make this collaborator more of a function than an object. While I can see a way to do that I kind of prefer a sort of command pattern for this kind of refactor where there is still some lightweight internal state in the object since it makes it trivial to break up the methods and try out various arrangements to find one that is more intuitive than the first one you pick.  That said, I totally agree that the manipulation of internal variables in objects is a rampant issue in this code base so I do see value in trying to make it clear that this internal state is not meant to be altered in that way. Do you think that this middle ground, where a class method that creates the object and immediately uses it such that users would not be tempted to alter the internal state of the object is acceptable?

charleshansen:
Yeah, I'm mostly good with it. The tests are much better at least, and I'd say the PR improves the code as a whole at this point. It is possible that the PackageProcessor actually needs more state than I expect because of the recursive package compilation part.

PR #2189, opened 2019-06-14
Extract package persistence from update release

jaresty:
This extracts a secondary collaborator that takes the output of
PackageProcessor as input to persist the results.

There may yet be tests in the UpdateRelease spec that are testing
PackagePersister behavior and would be better to move over. The
PackagePersister tests that were moved were only those that were
discovered because the existing UpdateRelease tests were setting mock
expectations on UpdateRelease itself to validate its behavior and thus
were difficult to leave where they were.


charleshansen:
As a first appraisal, this seems like a coherent enough object to be worth extracting. The only red flag for me are the couple tests that are mocking/expecting on the subject `job` in this case. If you can't write the tests without doing that, maybe this isn't the right code to extract.

Also for what it is worth, I like leaving at least some tests in UpdateRelease that are implicitly testing the PackagePersister. It is the only reliable way to actually test your wiring.

jaresty:
I agree with the assessment of the tests that are mocking the subject. Fwiw-those tests aren't new (I just moved them from UpdateRelease). I haven't had a chance to expand test coverage on the new object yet.

jaresty:
Updated tests to avoid mocking the test subject.

charleshansen:
Cool, the tests look better now. I have a few more requests for you:

- In the package persister specs, can you put tests for `create_package` and `persist` in their own separate contexts that are clearly labeled?
- Can you make the class methods other than `create_package` and `persist` private? I had to do some detective work to figure out they were not used outside of the PackagePersister.
- For the public class methods, can you use named arguments? There are too many arguments to just use ordered args. This is especially noticeable in the tests for UpdateRelease that look like `expect(Jobs::UpdateRelease::PackagePersister).to have_received(:persist).with.........` I'm not sure, but I'd hope you can change those tests to use `hash_including` if you have named arguments.

jaresty:
- made class methods private where possible.
- changed to use keyword arguments instead of long argument lists

I couldn't trivially make all class methods private since there are places where they are called from instance methods.

PR #2191, opened 2019-06-22
Refactor PackagePersister

jaresty:
- Moved non-fix logic to the top of the method since that is the more
frequent usage
- extracted safe_delete to make fix_package more linear
- moved compiled_package deletion into fix_package since it should
always happen in that case
- moved helper methods down in class definition since they are less
important
- reuse fix_package for both compiled and uncompiled packages
- inline low-value register_package method
- extract create_package_from_bits method


jaresty:
Hold off on merging this -- I've got a broken test to investigate.

jaresty:
Ok, fixed now.

jaresty:
Seeing some suspicious test failures. Closing this PR for the time being until I can be confident they've been addressed.

jaresty:
Was able to fix this and reopened as [PR 2192](https://github.com/cloudfoundry/bosh/pull/2192)

PR #2203, opened 2019-09-10
Introduce use-nats-pure property to director job

videlov:
- New director job property 'use_nats_pure' with values 'true'/'false'
- Switch NATS client library used at runtime based on it
- Use of a common NATS client wrapper for both clients in nats_rpc.rb

[#167423026](https://www.pivotaltracker.com/story/show/167423026)

Co-authored-by: Kai Hofstetter <kai.hofstetter@sap.com>
Co-authored-by: Denis Langer <denis.langer@sap.com>
Co-authored-by: Beyhan Veli <beyhan.veli@sap.com>


beyhan:
The unit tests and ITs were green with this change. ITs required the execution of `src/bosh-dev/assets/sandbox/nats_server/certs/generate.sh`.  We didn't add the generated sandbox certs to this change. Is this something outdated in the sandbox setup?

xtreme-conor-nosal:
The sandbox certs are still used. They expired earlier in the month leading to some maintenance and were regenerated on Sept 11.

xtreme-conor-nosal:
@videlov @beyhan Can you provide additional context on the motivation for this option? The chain of tracker stories mentioned dropped NATS messages and a spike on performance testing, but I couldn't see the initial troubleshooting or the performance testing results.

xtreme-conor-nosal:
@h4xnoodle @jfmyers9 given NATS' integral role, should there be a full integration suite run against pure nats instead of a single deployment test? (similar to the parallel postgres/mysql jobs, we would add a integration-pure-nats job)

h4xnoodle:
+1 for running the entire suite with the NATS gem change, instead of a test specifically for the change.

videlov:
@xtreme-conor-nosal , @h4xnoodle ,
1) about the background for this and why we want to introduce this option, you may take a look the findings document - https://docs.google.com/document/d/1kMeMnwWnG37sm19mZycHjremG5hY7wXujR24OvXVfIM/edit?usp=sharing
2) performance-wise we tested same conditions like with the nats.rb gem and we were able to go further. Here is the first story - https://www.pivotaltracker.com/n/projects/1456570/stories/166978331 , with nats-pure.rb we managed to deploy 1300 diego cell instances with 50 max-in-flight, additionally we did analyse bumping the max-in-flight to 100 and we managed to deploy 1000 diego cells in 10 deployments. About the update time for a single instance, it may be just a bit slower than the nats.rb version, but it run stable and it solves our issue with the NATS timeouts.
3) since this use-pure-nats options defaults to false, we think it should be treated more or less as an experimental option. we would like to test it extensively on our many landscapes and then provide more feedback.
4) regarding running whole integration tests with the nats-pure.rb gem, would you suggest we need to define new concourse job now before merging this or create extra story for it and in the meantime get more feedback on how this scales in our landscapes?

friegger:
@xtreme-conor-nosal @h4xnoodle 
As @videlov mentioned, the flag is intended to be an experimental flag that we will use to get more experience in big landscapes. While it certainly makes sense to run the whole test suite with this option, I feel it is not a pre-requisite for this PR to be merged.

friegger:
@videlov Did you check the comment regarding the sandbox certificates?

videlov:
The sandbox certificates are not related to this change / impl.

h4xnoodle:
I agree that this PR is fine as it is. What concerns me is the future, and the present landscape that doesn't demonstrate a need for this besides these scale tests carried out. That is, we don't actually see these problems where bosh is heavily used in real settings (at least that I've heard of so far!)

Bosh already has some experimental features that have been left and don't really see much light of day, yet will crop up when others try to use them and it fails. I don't think replacing the NATS gem will suffer the same maintenance/support burden as other features, but it is still something more to add to the project that may not be used with the exception of this extreme scale cases so far.

I would argue that we should either not include this at all, and it can remain experimental as a guide/patch not present in master, or the new gem should completely replace the current NATS client solution. I think a half-way/experimental feature will add more potential cruft to the project.

@mfine30 @jfmyers9 what do you think?

friegger:
First of all I also see value in reducing the complexity of BOSH and to avoid features and experimental flags that might never get cleaned-up in the future and that increase the maintenance burden.

The assumption that the NATS timeout errors only happen during scale tests is not correct. They happen in real settings and the sizing tested on scale were only slightly bigger than what runs in real settings. There is a limit where BOSH is not able to handle bigger deployments and the answer is probably to scale out to multiple CF foundations instead. This limit is of course fuzzy and depends on many different factors. From you perspective what is the number of instances BOSH can and should support?

Since this limit is so fuzzy, we offer to gain more insights in the context of our real setting landscapes, but to keep our efforts reasonable it would help not to have the burden of maintaining a fork with such a big change. If those insights lead to the conclusion that NATS pure does not yield other problems we are totally d'accord with settling on a single library going forward.

h4xnoodle:
Hi @friegger,

After discussion with @mfine30 and @jfmyers9 we feel that it would be better to include the pure nats gem _as a replacement_ for the current eventmachine-based library. 

We would rather avoid including more parallel & optional features/libraries, so we feel including it fully is best. We feel confident in the testing already done, and more testing will naturally be done when this is the only gem used.

Could you please update the PR such that the pure-nats gem is now the only nats gem used?

ssapra:
We found several integration tests failing from the changes and we are unsure if you've already addressed the new behavior in how the NATS client connection is created.

Specifically, the NATS rpc client was [lazily connected](https://github.com/cloudfoundry/bosh/blob/69e5f97fa95a7fa733bdc31bc02b1b2974ce97c2/src/bosh-director/lib/bosh/director/nats_rpc.rb#L20-L27), and with the switch to nats-pure, the code cannot proceed in the case that the client cannot connect to the NATS server. Given the non-lazy nature of the connection, it is also unclear what happens if an initially successful NATS connection is ever lost.

The [keep_unreachable_vms_spec.rb](https://github.com/cloudfoundry/bosh/blob/69e5f97fa95a7fa733bdc31bc02b1b2974ce97c2/src/spec/integration/keep_unreachable_vms_spec.rb) wants to see the error message "Timed out pinging" based on the `RpcTimeout` that is raised when the ping message fails to reach the NATS server. With nats-pure, we see `An error has occurred while connecting to NATS: Connection refused`. 

This connection behavior seems too risky to merge at this time. Have you seen this behavior or thought about any mitigations?

It should be possible to write a connection manager to wrap the nats client and solve the problem, but it is not a priority for us at the moment. Feel free to continue working on it and we can take another look once the integration test suite is passing.

You can view the current state of changes in https://github.com/cloudfoundry/bosh/tree/wip/fix-nats-pure. 

Thanks,
@ssapra and @charleshansen

friegger:
We will look into it, thank you.

beyhan:
We created a new pr https://github.com/cloudfoundry/bosh/pull/2254. The new pr reimplemented the `nats-pure.rb` integration and  is not based this one. Please use the new pr for any further discussions.

Issue #1908, opened 2018-03-09
bosh vms only display static ip address for vms configured with both dynamic and static networks


ywysuibian:
```
Version   263.3.0 (00000000)
CPI       openstack_cpi
```

I have some vms configured with both dynamic and static networks, but `bosh vms` only display static network ips for these vms.

I checked bosh director database, seems only static network ips stored in table `ip_addresses`.

voelzmo:
Hey @ywysuibian 

It seems you're trying to use multiple networks on a single VM, which is only supported in a very specific case: http://bosh.io/docs/openstack-multiple-networks.html

Due to missing functionality in the bosh-agent, other cases involving dynamic networking or DHCP are not supported: https://github.com/cloudfoundry/bosh/issues/1129#issuecomment-186286247

ywysuibian:
@voelzmo thank you for your kindly response.

I am not sure about `multiple networks`, maybe there is some misunderstanding in my previous description. 
I was using this kind of configuration for almost two years, and I just upgraded bosh director from a very lower version to 263.3, then upgraded cf from 253 to 270. During cf upgrading, all VMs were recreated (maybe due to stemcell upgrading),  and then `bosh vms` cannot display **private ip** of VMs which also configured with an **floating ip**.
 
My deployment yml looks like this
```yaml
instance_groups:
...
- instances: 1
  jobs:
  - name: aaaa
    release: aaaa-release
  - name: bbbb
    release: bbbb-release
  - name: cccc
    release: cccc-release
  name: xxxx
  networks:
  - default:
    - dns
    - gateway
    name: default
  - name: floating
    static_ips:
    - xxx.xxx.xxx.ipb
```

In the response from bosh_agent to director, I foud
```json
...
{
  "networks": {
    "default": {
      "cloud_properties": {

      },
      "default": [
        "dns",
        "gateway"
      ],
      "dns": [
        "xxx.xxx.xxx.xxx",
        "xxx.xxx.xxx.ipa"
      ],
      "gateway": "xxx.xxx.xxx.1",
      "ip": "xxx.xxx.xxx.ipa",
      "netmask": "255.255.255.0",
      "type": "dynamic"
    },
    "floating": {
      "cloud_properties": {
        
      },
      "ip": "xxx.xxx.xxx.ipb",
      "type": "vip"
    }
  }
...
}
```

But in bosh director database, I can only find one record in ip_addresses table related to this VM. After I delete this record, `bosh vms` became to display both ips.

[...]

PR #2206, opened 2019-09-15
Director API will return both static & dynamic IPs

tlwr:

Related (closed) issue: https://github.com/cloudfoundry/bosh/issues/1908

(More info in commit message)

Currently when an instance is configured with both manual/vip and dynamic networks, the Director API returns IP addresses only from vip/manual networks, ignoring any IPs that have been allocated dynamically via the IaaS.

As a user of the AWS CPI, I want to be able to have networks in my cloud config like:
```
networks:
  - name: public
    type: vip

  - name: internal
    type: dynamic
    subnets:
      - az: z1
        cloud_properties:
          subnet: subnet-123456
```

and then have an instance group like:

```
instance_groups:
  - name: my-instance
    instances: 1
    azs: [z1]
    networks:
      - name: public
        static_ips: [1.2.3.4]
      - name: internal
        default: [dns, gateway]
```

In this example my instance has two IP addresses, e.g. `10.0.0.10` (dynamically allocated by AWS) and `1.2.3.4` which is an AWS Elastic IP.

BOSH should:

1. display both of these IPs to me when I do `bosh instances` at the CLI, right now only `1.2.3.4` will be displayed
2. return both of these IPs to consumers of the BOSH API (including the CLI), so that things like the BOSH exporter can get dynamic internal IPs as well as external ones.

Currently a workaround exists for this, which is to configure all networks manually and replicate the CIDR ranges of the subnets defined in AWS in `manual` network definitions. This works, but in the case where the subnet is shared between BOSH and something else, BOSH expects to manage the whole range, and asks the CPI to use an address which is already in use (e.g. `Address 10.0.0.10 is in use` from the AWS CPI) - which is AFAICT the point of dynamic networks.

I have an even more concrete use case of why this is important but it is probably outside the purview of this PR.

I have run the unit tests (`bundle exec rake spec:unit:director`)

Edit: I have deployed this to a real BOSH director running the AWS CPI and it works as expected.

> The Director will now return all IP addresses that are assigned to an instance, regardless of network type. Previously BOSH would return only IP addresses would only return IP addresses from manual/vip networks, if both dynamic and static networks were configured. Now all IP addresses are returned, statically defined IPs followed by dynamically defined IPs.

If there are API consumers which rely on only statically defined IP addresses being returned by the API then this will be disruptive, however this change does not affect the API schema.

Solo - this is my first PR to bosh so I might have done something egregious

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
@@ -611,15 +611,21 @@ module Bosh::Director
       end
 
       def ips(vm)
+        manual_or_vip_ips(vm) + dynamic_ips(vm)
+      end
+
+      def manual_or_vip_ips(vm)
         return [] if vm.nil?
 
-        # For manual and vip networks
-        result = vm.ip_addresses.map { |ip| NetAddr::CIDR.create(ip.address).ip }
-
-        # For dynamic networks
-        result = vm.network_spec.map { |_, network| network['ip'] } if result.empty? && !vm.network_spec.empty?
-        result
+        vm.ip_addresses.map { |ip| NetAddr::CIDR.create(ip.address).ip }
       end
+
+      def dynamic_ips(vm)
+        return [] if vm.nil?
+
+        vm.network_spec.map { |_, network| network['ip'] }
+      end
+
     end
   end
 end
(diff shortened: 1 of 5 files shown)


tlwr:
I've got this running in my BOSH director on AWS CPI and it works as expected...

```
Instance                                               Process State  AZ  IPs
concourse-worker/5e9f96be-43f5-48e5-8ec4-b3c831cd5811  running        b1  10.0.0.242
concourse/5a54723d-cb7f-4c8c-9ade-2297796682cc         running        b1  redacted-public-ip
                                                                          10.0.0.135
```

PR #2217, opened 2019-10-29
Make brats easier to run locally

jaresty:
- Sometimes you want to run brats against a dirty git repository
- Sometimes blobs are in an inconsistent state so it's safer to run
reset-release first
- When re-running test-brats, start-bosh.sh will have already run so
it's not necessary to start the outer bosh director again
   - Moreover, start-bosh fails the second time since the heuristic it uses
   to determine the IP address results in broken yml and fails


xtreme-conor-nosal:
An additional consideration would be to run brats on a currently targeted director instead of starting a docker-bosh (decouple the director setup from the test run)

PR #2218, opened 2019-11-07
Make bosh-deployment overrideable for brats

jaresty:
This change makes bosh-deployment overrideable as a concourse input.

### DIFF
--- a/ci/tasks/test-brats.sh
+++ b/ci/tasks/test-brats.sh
@@ -22,7 +22,12 @@ DNS_RELEASE_PATH="$(realpath "$(find "${PWD}"/bosh-dns-release -maxdepth 1 -path
 export DNS_RELEASE_PATH
 CANDIDATE_STEMCELL_TARBALL_PATH="$(realpath "${src_dir}"/stemcell/*.tgz)"
 export CANDIDATE_STEMCELL_TARBALL_PATH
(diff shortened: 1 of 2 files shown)


ci/tasks/test-brats.sh
```diff
@@ -22,7 +22,12 @@ DNS_RELEASE_PATH="$(realpath "$(find "${PWD}"/bosh-dns-release -maxdepth 1 -path
 export DNS_RELEASE_PATH
 CANDIDATE_STEMCELL_TARBALL_PATH="$(realpath "${src_dir}"/stemcell/*.tgz)"
 export CANDIDATE_STEMCELL_TARBALL_PATH
-export BOSH_DEPLOYMENT_PATH="/usr/local/bosh-deployment"
+export OVERRIDDEN_BOSH_DEPLOYMENT=$(realpath "$(dirname $0)/../../../bosh-deployment")
```
charleshansen: I am surprised it is necessary to use relative paths here. I believe when this code runs on concourse, you should be able to just hit "${PWD}/bosh-deployment". Am I missing something that made the relatively paths necessary?

jaresty: It's not really necessary I think; force of habit perhaps.  If you were to use setup-brats and then try to run test-brats.sh after cd'ing into the directory this part would still catch the override (which it would not with absolute paths).

PR #2219, opened 2019-11-09
there are no required credentials sources for s3cli

bengerman13:

This removes the required credentials settings for s3cli blobstores

s3cli blobstores can use `static` or `env_or_profile` sources for credentials, and each has a different set of config keys. There is no set of credential properties that is required in all cases. In it's current form, the validation breaks deployments using `env_or_profile`.

An alternate fix I considered is to provide an override to the `validate` method for the s3cli blobstore that checks to make sure that either `credential_source` is set to `env_or_profile` or that `access_key_id` and `secret_access_key` are set - let me know if that's preferred and I can look at it.

I ran `./scripts/test-unit` without no failures

no

### DIFF
--- a/src/bosh-director/lib/bosh/blobstore_client/s3cli_blobstore_client.rb
+++ b/src/bosh-director/lib/bosh/blobstore_client/s3cli_blobstore_client.rb
@@ -126,7 +126,8 @@ module Bosh::Blobstore
     end
 
     def required_credential_properties_list
-      []
+      return [] if @s3cli_options[:credentials_source] == 'env_or_profile'
+      %w[access_key_id secret_access_key] 
     end
 
     # @param [String] path path to file which will be stored in S3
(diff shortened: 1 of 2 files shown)


src/bosh-director/lib/bosh/blobstore_client/s3cli_blobstore_client.rb
```diff
@@ -126,7 +126,7 @@ def sign_url(object_id, verb, duration)
     end
 
     def required_credential_properties_list
-      %w[access_key_id secret_access_key]
+      []
```
h4xnoodle: This needs to be conditional on whether `credentials_source` is `static` or not. We need to validate credentials exist/DNE if the source is `static` still.

bengerman13: thanks @h4xnoodle 
I added the logic to validate `static` back in, and added a couple tests.

PR #2240, opened 2020-02-18
Update deprected rubocop definition

beyhan:
Executing rubocop in the bosh repo fails with the error:
  The `Layout/IndentFirstHashElement` cop has been renamed to `Layout/FirstHashElementIndentation


h4xnoodle:
Thanks @beyhan. That's interesting, no one on the team saw this problem. I'm using 0.79.0 and with rubocop running it was able to find no problems, so I assume this evaluated the rules still.

PR #2254, opened 2020-03-24
Switch nats.rb to nats-pure.rb

beyhan:

This pr replaces the `nats.rb` client library with the `nats-pure.rb` library in `bosh` and in `health-monitor`.

The context for this pr can be found in pr: https://github.com/cloudfoundry/bosh/pull/2203
Also we implemented all the improvement suggestions provided in: https://github.com/cloudfoundry/bosh/pull/2203#issuecomment-577922249
We plan to use this change in production environments and will provide more feedback later. We open this pr to start the discussions around this change as early as possible.

all unit tests
all integration tests
manual tests 

No braking changes.

@videlov 
@langered 
@friegger


h4xnoodle:
Looks good to me, I'll merge in and bring it through the pipeline with BATS/BRATS. 

What are your thoughts on the versioning for this release? While it's technically not a _breaking_ change, I'm thinking of cutting a major version so if there are issues for others, they can easily switch to a version that gets backports for other CVEs/bugs, avoiding the new NATS  gem. I don't foresee major issues with this PR, but we don't know the details of all bosh-deployed landscapes.

friegger:
Thanks @h4xnoodle! Let us know in case you see issues with BATS/BRATS.

Regarding the versioning, cutting a new major makes sense for the reasons you mentioned. I'd suggest to release the few current fixes on master as minor first. What do you think?

Issue #2255, opened 2020-03-27
clean up tasks fails with wrong number of arguments (given 0, expected 1)


friegger:
**Describe the bug**
The `clean up tasks` tasks fails every time.

**To Reproduce**
Steps to reproduce the behavior (example):
1. Deploy a bosh director `270.12.0`
2. Wait until the first scheduled task clean up runs

**Expected behavior**
Don't error but clean up tasks.

**Logs**
```
E, [2020-03-22T00:00:02.054392 #28825] [task:104624] ERROR -- DirectorJobRunner: wrong number of arguments (given 0, expected 1)
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb:18:in `initialize'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/job_runner.rb:84:in `new'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/job_runner.rb:84:in `perform_job'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/job_runner.rb:34:in `block in run'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh_common-0.0.0/lib/common/thread_formatter.rb:50:in `with_thread_name'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/job_runner.rb:34:in `run'
/var/vcap/data/packages/director/4de68307cefbbb6cbb7e4a39d40adef9bf335e6a/gem_home/ruby/2.6.0/gems/bosh-director-0.0.0/lib/bosh/director/jobs/base_job.rb:9:in `perform'
[...]

PR #2256, opened 2020-03-30
Add default value to ScheduledTasksCleanup initialize argument

beyhan:
Fixes: #2255

unit tests
manual tests


voelzmo:
Looking at https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/jobs/scheduled_events_cleanup.rb I'm wondering if ScheduledTasksCleanup#initialize should look the same, i.e. getting the number of tasks to cleanup as parameter?

h4xnoodle:
@voelzmo the number to delete is calculated at the time of perform, and `has_work` will use the same calculation. https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb#L25

Issue #2253, opened 2020-03-23
Resurrection of VMs with multiple disk in not working


rm75:
Resurrection of VMs with multiple persistent disks are not working.

Seems like it supports only single persistent disk resurrection.
https://github.com/cloudfoundry/bosh/blob/17921ed6431d126f5a029392e0498c9013d26148/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb#L73

Opening this request based on the conversation in the slack channel.

voelzmo:
Assuming this is the slack thread, just for reference and cross-linking: https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1578339832051000

h4xnoodle:
Multiple persistent disks are semi-supported, at best. The feature is there for extreme cases where you absolutely need two classes of persistent disk. In most cases, one disk is enough.

Does your use-case only work with multiple disks attached or is this some kind of convenience? Note that we do not have any documentation on bosh.io for multiple disks, therefore not supporting any complex use-cases. We make no guarantees that multiple persistent disks work in all cases. 

As you've already identified the area of code, we are welcoming contributions :)

Thanks

PR #2258, opened 2020-04-22
Bosh multiple resurrection support

rm75:

This PR is for the resurrection of VMs with multiple persistent disks.

Currently VMs with multiple persistent disks does not start after resurrection. 

https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1578339832051000

Resurrection of VMs with multiple disk in not working (#2253)

https://www.pivotaltracker.com/story/show/171919518

Ran Director unit tests

Also tested the following scenarios with the changes in the Lab BOSH Director
1) Resurrection of VMs with no persistent disks
2) Resurrection of VMs with single persistent disk
3) Resurrection of VMs with single multiple persistent disks

No.

### DIFF
--- a/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb
+++ b/src/bosh-director/lib/bosh/director/cloudcheck_helper.rb
@@ -70,7 +70,7 @@ module Bosh::Director
           .create_for_instance_plan(
             instance_plan_to_create,
             planner.ip_provider,
-            Array(instance_model.managed_persistent_disk_cid),
+            Array(instance_model.active_persistent_disks.collection.map { |d| d.model.disk_cid}),
             instance_plan_to_create.tags,
             true,
           )
--- a/src/bosh-director/lib/bosh/director/deployment_plan/instance_plan_from_db.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/instance_plan_from_db.rb
@@ -57,7 +57,7 @@ module Bosh
         end
 
         def needs_disk?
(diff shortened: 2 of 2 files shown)


mrosecrance:
We'll need to have the CLA signed to merge this - it used to be the check was only run on pr open, so you may need to close and reopen this after signing. 

Unit tests would be good possibly [here](https://github.com/cloudfoundry/bosh/blob/9944537ce5dbd03df1b42da3af5bcfa958919bcb/src/bosh-director/spec/unit/cloudcheck_helper_spec.rb#L261). There's also a method that lists out the [disk cids](https://github.com/cloudfoundry/bosh/blob/27ec7157468638b9ceababddc85c3b59da6e7211/src/bosh-director/lib/bosh/director/models/instance.rb#L34) already which cleans up the code you've submitted a bit.

Multi persistent disks are sort of a half implemented feature which your PR improves but your PR also introduces a change to how normal cck operations will work (resurrection which is used by restore and cck share the same code path) so integration tests showing that would be good if you're feeling ambitious. [perhaps here](https://github.com/cloudfoundry/bosh/blob/9944537ce5dbd03df1b42da3af5bcfa958919bcb/src/spec/integration/cli_cck/unresponsive_agents_spec.rb#L122)

mrosecrance:
Poking this again about the CLA - it needs to be signed to merge this.

mrosecrance:
I'm going to close this due to lack of follow up. Feel free to reopen when you've addressed the comments above.

PR #2270, opened 2020-07-27
Prestart director job without recursive chown on storage folder

videlov:

Change director pre-start job to just `chown` the root store folder and do not do it recursively.

On one of our landscapes we noticed that this usage of `find + chown` causes few minutes to complete. Before merging let discuss if this change could cause some implications.

We have tested newly deployed bosh and an updated bosh with the change and it worked fine.

Not sure.

It should not.

@friegger @c0d1ngm0nk3y

### DIFF
--- a/jobs/director/templates/pre-start.erb
+++ b/jobs/director/templates/pre-start.erb
@@ -8,7 +8,7 @@ DB_MUTUAL_TLS_PRIVATE_KEY=/var/vcap/jobs/director/config/db/client_private_key.k
 TMPDIR=/var/vcap/data/director/tmp
 
 mkdir -p $STORE_DIR
-find $STORE_DIR \( -not -user vcap -or -not -group vcap \) -execdir chown vcap:vcap {} \+
+chown vcap:vcap $STORE_DIR
 
 chmod -R 0640 $CONFIG_DIR
 find $CONFIG_DIR -type d | xargs -n1 chmod 0750


mrosecrance:
It's hard for me to reason about what the effect of this might be. Have you tested this out? I assume that perhaps some child directories would be owned by root and with `-rw-r-----` permissions, that would mean that vcap could only read (I assume it's in the same group). I don't know if those child directories are important or if everyone of value is created after the pre-start. It does look like an attempt was made it speed it up: https://github.com/cloudfoundry/bosh/commit/2c3ad0339ebc2b0e8ea2ec82f672fcef9af93550 previously. Is your system significantly larger than what was tested?

friegger:
Yes we've tested this. Directors with this change already run on 4 different infrastructures. Pre-start is down to 1s. Biggest impact is on Azure landscapes, where disk performance seems to be slower. Here we previously had ~4-6min for this step. It's real downtime.

Sample Usage:
```
/var/vcap/store# df . -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/sdd1       252G  188G   52G  79% /var/vcap/store

/var/vcap/store# du . -h -d 1
4.0K	./health_monitor
19G	./director
165G	./blobstore
16K	./lost+found
1.6G	./postgres-9.4
3.0G	./postgres-10
188G	.
```

We also tested the scenario with a new disk and then updating the Director with that disk.

The previous change did not really improve the situation here, probably `find` is rather slow the very first time it runs. Afterwards it seems to rely on cache.

The recursive chown is there since the project has moved to the cloudfoundry org. We assumed it might be there, because in ancient times things where not executed as vcap, but were also wondering whether there could be another reason.

We'll report back, when the change has reached all our systems.

PR #2275, opened 2020-08-31
[WIP] Add list persistent disks end point

beyhan:

This is useful for:
- debugging of disk issues. E.g. a disk, which is not attached
to any instance, but it appears in the IaaS console as bosh managed disk. Some CPIs added the
tagging of disks too late and there are disks, which don't have any tags.
- collect information for the disks managed by bosh. This information
could be used to calculate costs...

`bosh disks` command can only list `orphaned` disks. It will be very useful to extend the command to list persistent disks in use also.

X

X

no

X

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/disks_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/disks_controller.rb
@@ -9,6 +9,12 @@ module Bosh::Director
         json_encode(orphan_json)
       end
 
+      get '/persistent_disks' do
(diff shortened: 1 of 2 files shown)


beyhan:
@mrosecrance do you think that this is an useful feature, which you would like to add to the director? Current state is without any tests just for discussion.

mrosecrance:
This seems like a sane feature. Stylistically it seems a little neater to add an optional param to the '/' disk controller endpoint as long as we keep backwards compatibility but I don't feel strongly.

PR #2279, opened 2020-09-18
Resolve blobstore logging warning [#174649741]

rkoster:

The blobstore log contained a warning which has confused people while debugging other blobstore issues.
The log line:
```
nginx: [alert] could not open error log file: open() "/var/vcap/packages/nginx/logs/error.log" failed (30: Read-only file system)
```
This was caused by the fact that the default prefix `/var/vcap/packages/nginx` was not writable.
By changing the prefix (using the `-p` flag) to `/var/vcap/data/blobstore` (which is writable), this issue was resolved.

This change also increases the log level to include warning.
Which resulted in the following log line:
```
 [warn] 6#0: 8192 worker_connections exceed open file resource limit: 1024
```
This was fixed by increasing the open_files limit via bpm to 8192.

While working on this I also noticed nginx was running in daemon mode (which makes sense under bpm).
However it still had a pid location configured which seemed unnecessary. 

https://www.pivotaltracker.com/story/show/174649741

Manual testing to verify warnings are ending up in the correct file:
```
==> /var/vcap/sys/log/blobstore/error.log <==
2020/09/18 10:08:33 [warn] 6#0: 8192 worker_connections exceed open file resource limit: 1024
```
This warning should not be displayed anymore after this PR is merged since the limit in bpm is increased.
Lowering the limit manually in the bpm config and restarting the blobstore should work.

Improved blobstore error logging

No

### DIFF
--- a/jobs/blobstore/templates/bpm.yml
+++ b/jobs/blobstore/templates/bpm.yml
@@ -3,12 +3,21 @@
 blobstore_config = {
   "name" => "blobstore",
   "executable" => "/var/vcap/packages/nginx/sbin/nginx",
-  "args" => ["-c", "/var/vcap/jobs/blobstore/config/nginx.conf"],
+  "args" => [
+    "-c",
+    "/var/vcap/jobs/blobstore/config/nginx.conf",
+    "-p",
+    "/var/vcap/data/blobstore"
+  ],
+  "limits" => {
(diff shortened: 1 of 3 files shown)


cunnie:
@rkoster Thanks for the excellent work!

Issue #2271, opened 2020-07-28
Director http metrics path label contains concrete values for variable segments


friegger:
**Describe the bug**

The Director http metrics collected contain concrete values for variable values in the path label.

```
# bosh -d real-deployment-name1 instances
...
http_server_requests_total{code="404",method="get",path="/deployments/real-deployment-name1/instances"} 1.0
# bosh -d real-deployment-name2 instances
...
http_server_requests_total{code="404",method="get",path="/deployments/real-deployment-name2/instances"} 1.0
```

This can lead to a huge amount of metric series which are not very useful. In addition the metric endpoint gets very slow, with a lot of requests that contain different deployment names.

The middleware used to collect Director http metrics replaces path elements for ids and uuids:
```
# bosh task 5
http_server_requests_total{code="416",method="get",path="/tasks/:id/output"} 1.0
```

**To Reproduce**
Steps to reproduce the behavior (example):
1. Deploy a bosh director on any IaaS with [metrics enabled](https://github.com/cloudfoundry/bosh-deployment/blob/master/experimental/enable-metrics.yml)
2. Exexute sample commands from above
5. Check metrics by hitting Director's /metric endpoint

**Expected behavior**
All variable path elements should be replaced by a place holder.

```
# bosh -d real-deployment-name1 instances
...
http_server_requests_total{code="404",method="get",path="/deployments/:deployment/instances"} 1.0
# bosh -d real-deployment-name2 instances
...
http_server_requests_total{code="404",method="get",path="/deployments/:deployment/instances"} 2.0
```

This can be achieved by implementing a custom Collector specific for the Director as described in the [examples](https://github.com/prometheus/client_ruby/blob/master/examples/rack/README.md#collector).

**Versions (please complete the following information):**
 - Infrastructure: any
 - BOSH version 270.1.0
 - BOSH CLI version 6.x

friegger:
#2284 has been merged.

PR #2284, opened 2020-10-26
Strip individual deployment name for Puma metrics

videlov:
Fixes issue: https://github.com/cloudfoundry/bosh/issues/2271

### DIFF
--- a/src/bosh-director/lib/bosh/director/custom_middleware_collector.rb
+++ b/src/bosh-director/lib/bosh/director/custom_middleware_collector.rb
@@ -1,0 +1,14 @@
+require 'prometheus/middleware/collector'
+
+module Bosh
+  module Director
+    class CustomMiddlewareCollector < Prometheus::Middleware::Collector
+      def strip_ids_from_path(path)
+        path
+          .gsub(%r{/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(/|$)}, '/:uuid\\1')
+          .gsub(%r{/\d+(/|$)}, '/:id\\1')
+          .gsub(%r{/deployments/[a-z0-9_-]+(/|$)}, '/deployments/:deployment\\1')
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


friegger:
@dlresende From our perspective this can be merged. Can someone provide feedback?

src/bosh-director/spec/unit/custom_middleware_collector_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+describe Bosh::Director::CustomMiddlewareCollector do
```
klakin-pivotal, line 3: The `CustomMiddlewareCollector` class name is very vague and nondescript. Would it be reasonably possible for you to change the name of this class to one that describes its purpose and function?

src/bosh-director/lib/bosh/director/custom_middleware_collector.rb
```diff
@@ -0,0 +1,14 @@
+require 'prometheus/middleware/collector'
+
+module Bosh
+  module Director
+    class CustomMiddlewareCollector < Prometheus::Middleware::Collector
+      def strip_ids_from_path(path)
+        path
+          .gsub(%r{/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}(/|$)}, '/:uuid\\1')
+          .gsub(%r{/\d+(/|$)}, '/:id\\1')
+          .gsub(%r{/deployments/[a-z0-9_-]+(/|$)}, '/deployments/:deployment\\1')
```
klakin-pivotal, line 10: Is there a reason that the default behavior of `strip_ids_from_path` is copied from the superclass? Are you particularly concerned that the behavior of the superclass will change in the future? If not, this should be changed to call the superclass, then chain your `deployment` stripping code. (This also happens to be the way recommended by the docs: <https://github.com/prometheus/client_ruby/blob/master/examples/rack/README.md#collector>)

I bring this up because -prior to this change- we were using the default behavior of the `Prometheus::Middleware::Collector`'s `strip_ids_from_path` function... so if we ever upgraded that library and that function's behavior ever changed, we would automatically get the new code.

ystros:
Since the feedback elements are somewhat small, we're going to go ahead and merge this and make the requested changes to naming / super class behavior. Thanks!

PR #2286, opened 2020-11-03
enable bpm again for health-monitor

rkoster:
This PR addresses: https://www.pivotaltracker.com/story/show/175544179


jobs/health_monitor/templates/bpm.yml
```diff
@@ -0,0 +1,35 @@
...
+    "BUNDLE_GEMFILE" => "/var/vcap/packages/health_monitor/Gemfile",
+    "GEM_HOME" => "/var/vcap/packages/health_monitor/gem_home/ruby/2.6.0",
+  },
+  "additional_volumes" => [{
+    "path" => "/var/vcap/data/jobs",
+    "allow_executions" => true,
+  }],
+  "unsafe" => {
+    "unrestricted_volumes" => [{
+      "path" => "/var/vcap/jobs",
```
voelzmo, line 15: Two questions about the volume configuration here
* why do we have two mount configurations for the `jobs` directory? `/var/vcap/data/jobs` and `/var/vcap/jobs` have the same contents, one is just a symlink to the other
* shouldn't we rather use a more specific glob for mounting the json plugin executables only? Something like `/var/vcap/jobs/*/bin/bosh-monitor` should only mount the stuff the HM needs, not the other, sensitive files as well. There's a reason why BPM prevented access to other job directories per default: they contain secrets most of the time.

voelzmo (commented):
Sorry for commenting on a merged PR, just saw this now.

jobs/health_monitor/templates/bpm.yml
rkoster: Did not know bpm could use globs, but just found the test for it: https://github.com/cloudfoundry/bpm-release/blob/fdb06e2e7d18ce51c01ca330deffcbfdf0296e56/src/bpm/integration/start_test.go#L409-L440. Will make a story make this improvement.
Just to be fair. The current change it not making things worse with regards to security. As before this change health monitor did not use bpm at all. So already had all access. The change was made to improve the reliability of the health_monitor start/stop sequence. Mostly to prevent multiple instances running at the same time (which some clients have seen).

Issue #2303, opened 2021-06-30
`attach-disk` fails in multi-CPI envs


Justin-W:
**Describe the bug**
It seems that the `bosh attach-disk` command does not support multi-CPI configurations, since it seems to use the "default CPI" when attempting to attach the disk to the VM, resulting in a (CPI-dependent) error if/when the  default CPI cannot successfully attach the disk to the VM.

**To Reproduce**

We encountered this bug when attempting to (re-)attach a persistent disk that we restored from a Bosh-created snapshot. This same process previously worked for us in a single-CPI Bosh environment, but failed this time in an environment here we use 2 different AWS CPIs.

1. Deploy a bosh director on AWS in Region A (CPI-dir)
2. Add a cloud-config and cpi-config for an AWS CPI (CPI-B) for Region B
3. Deploy a VM (VM-1) to Region B using CPI-B
4. Have Bosh take a snapshot (SNAP-1) of VM-1's persistent disk
5. Use AWS to manually restore a snapshot SNAP-1 as disk DISK-RESTORED
6. Run `bosh attach-disk <VM-1> <DISK-RESTORED>` to see the behavior (attach fails; presumably because it uses CPI-dir instead of CPI-B)

Note: Based on previous results, if CPI-dir and CPI-B have similar configuration, such that CPI-dir is able to perform IaaS API operations on VM-1, then Step 6 succeeds without error. But if the 2 CPIs have significantly different configuration, then CPI-dir fails during step 6 when calling the IaaS APIs.

**Expected behavior**

Expected Step 6 (above) to succeed, using the correct CPI (CPI-B) for the VM and disk being targeted.

Per the docs, restoring a restored disk from a snapshot _should_ work:

From: https://bosh.io/docs/persistent-disks/#orphaned-disks:

> `attach-disk` command can also attach available disks found in the IaaS. They don't have to be listed in the orphaned disks list.

From: https://bosh.io/docs/snapshots/:

> Take a disk snapshot of a persistent disk before deploying major updates or for other important events. If the changes corrupt persistent disk, promote a disk snapshot to be a persistent disk and attach it to the VM to restore data prior to your changes.

**Logs**

Attaching the restored disk failed with this log output:

```
D, [2021-06-28T23:31:07.192430 #14517] [task:1015950] DEBUG -- DirectorJobRunner: Using cpi_version 2 for CPI 
[...]

PR #2304, opened 2021-06-30
Fix Issue 2303

Justin-W:

Fixes the bug described in issue #2303.

See bug #2303.

Added and updated unit tests.

Fixed `bosh attach-disk` to correctly support multi-CPI configurations.

No.


rkoster (approved):
This behaviour matches all other uses of PersistentDisk.create:
- [disk_manager.rb](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/disk_manager.rb#L208)
- [oprphan_disk_manager](https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/director/orphan_disk_manager.rb#L52)

rkoster:
@Justin-W Thanks!

PR #2305, opened 2021-07-06
adding build chroot docker images

ramonskie:
the ubuntu-chroot docker images is used by the stemcell builder ci and the one that we reffered to was: https://github.com/cloudfoundry/bosh/blob/master/ci/old-docker/main-ubuntu-chroot/Dockerfile

this was still using the main-ruby-go:trusty and was not being build or updated.

this PR wil enable building the docker image again


ramonskie:
this is not relevant anymore for the stemcell builder tests.
as we are moving away from this container
see reference https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-bionic/master/bosh-stemcell/spec/support/os_image.rb#L9

PR #2306, opened 2021-07-11
Send custom tags in DataDog health monitor events

rcw5:
Extend the DataDog Health Monitor plugin to include custom tags when creating DataDog events for BOSH HM alerts

Before this change the DataDog Health Monitor plugin supported sending custom tags for metrics, but not events. This change also sends the same custom tag set whenever an event is generated due to a Health Monitor alert.

Operators often use user-friendly names for their CloudFoundry deployments, which are set as custom tags in DataDog for the firehose, OS agent, etc. These tags are then used as filters in dashboards, monitors, incidents, the event stream, etc.

Attaching these custom tags to DataDog events created by the Health Monitor allows for consistent tagging across an entire foundry. 

Tags were added to DataDog metrics in https://github.com/cloudfoundry/bosh/pull/2077. 

 - RSpec tests against the `bosh-monitor` folder
 
 - Created a development release and deployed a Director with DataDog `custom_tags` set as:

```
    custom_tags:
      custom_tag_1: hello
      custom_tag_2: world
```
Verified these tags are visible inside the DataDog event stream
Screenshot 2021-07-11 at 17 42 05

Custom tags can be attached to DataDog events generated by the Health Monitor

No

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/plugins/datadog.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/plugins/datadog.rb
@@ -94,6 +94,7 @@ module Bosh::Monitor
         [].tap do |tags|
           tags << "source:#{data[:source]}"
           tags << "deployment:#{data[:deployment]}"
+          custom_tags.each { |key, value| tags << "#{key}:#{value}" }
         end
       end
     end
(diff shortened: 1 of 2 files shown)


rkoster:
LGTM! @rcw5 Thanks!

PR #2310, opened 2021-08-23
Ruby 3.0

rkoster:
Original story:
```
2.7 won't EOL until 2023 (likley, it's not even set yet) but now that 3.0 has been out for a few months and probably doesn't have major bugs we should bump to something that supposedly 3x as fast.
```

https://www.pivotaltracker.com/story/show/178224013


rkoster:
This PR was already reviewed by @jpalermo

Issue #2276, opened 2020-09-11
[Feature Request] External CPI calls should eventually timeout


mrosecrance:
**Is your feature request related to a problem? Please describe.**
There's a few reasons CPI calls may hang. A call between the cpi and IAAS may have network issues or the initial connection to the IAAS from the CPI may fail. As a result, the bosh task backing the call hangs indefinitely in a "running" state until manually cancelled. There's also a small chance the director -> cpi communication channel has an error. It's done using Open3 which spawns a process to call the binary but there were once upon a time deadlock issues.

**Describe the solution you'd like**
There should be a -very long- timeout on bosh's side to cpi calls (hours? we should probably look into what the most expensive possible calls could be). Having tasks fail with an error message prompting users to look at the cpi logs would help.  It's also likely that if it is flake, users can deploy their way forwards with minimal disruption. 

**Describe alternatives you've considered**
Timeout's can be implemented on the cpi side - but there's a fan out issues as there are multiple CPI's. It also doesn't help with possible director -> cpi hanging communication issues.

mrosecrance:
Here's an instance https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1573162453274000 where it looks to have hung connecting to the IAAS. It does look like the vsphere cpi at least has some [timeout functionality](https://github.com/cloudfoundry/bosh-vsphere-cpi-release/blob/master/src/vsphere_cpi/lib/nsxt/nsxt_client/configuration.rb#L82)

beyhan:
I think that it is better to have CPIs with properly configured IaaS/HTTP clients, because the CPIs can better decide when to timeout. E.g. for the OpenStack CPI connection options are even configurable via the [bosh manifest](https://bosh.io/jobs/openstack_cpi?source=github.com/cloudfoundry/bosh-openstack-cpi-release#p%3dopenstack.connection_options). You can set http (`excon`) options depending on the OpenStack needs. There are also situation, where the CPIs [retry](https://github.com/cloudfoundry/bosh-aws-cpi-release/issues/102) and that is why it takes too long. A general safeguard on the bosh side, which can terminate the operation if something takes too long is a good idea. It is just difficult to find an optimal timeout, but a "very long" timeout is for sure better than no timeout.

mrosecrance:
[...]

PR #2319, opened 2021-09-07
External cpi log streaming

rkoster:

Currently, the external cpi code uses `IO.capture3` for invoking cpi commands. This works pretty well, however, we have experienced cases where the cpi calls would hang for a few days, and since the logs are only written when the process exits there is no way to debug these cases. 

This PR changes cpi invocation to use `IO.popen3` which allows us to stream cpi error logs while the command is still running. This should help to make debugging infinite cpi loop issues easier.

One public instance: https://github.com/cloudfoundry/bosh-vsphere-cpi-release/issues/276
But also internal issues once a month.

- src/bosh-director/spec/unit/cloud/external_cpi_*
- src/spec/integration/cpi_spec.rb

Added support for streaming cpi logs, to help debug hanging cpi issues.

The order in which err log and stderr end up in the logs has changed.

### DIFF
--- a/src/bosh-director/lib/cloud/external_cpi.rb
+++ b/src/bosh-director/lib/cloud/external_cpi.rb
@@ -96,16 +96,43 @@ module Bosh::Clouds
       logger = ::Bosh::Director::TaggedLogger.new(@logger, request_id)
 
       logger.debug("request: #{redacted_request} with command: #{cpi_exec_path}")
-      cpi_response, stderr, exit_status = Open3.capture3(env, cpi_exec_path, stdin_data: request, unsetenv_others: true)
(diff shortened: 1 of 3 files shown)


beyhan:
@rkoster this is a really useful feature! In https://github.com/cloudfoundry/bosh/issues/2276 we were discussing different options for the hanging issue. A timeout could be helpful here also but it's challenging to find the right value for it.

lnguyen:
We've done manual acceptance test and confirmed logs are being streamed. It didn't work on warden because it pipes all messages to stderr [found here](https://github.com/cloudfoundry/bosh-warden-cpi-release/blob/master/jobs/warden_cpi/templates/cpi.erb#L12), after commenting out, the logs are able to stream to test the changes.

PR #2323, opened 2021-09-28
Feature: Adding support for NATS/Blobstore cert rotation without recreating vms

lnguyen:
Currently rotating nats or blobstore certs, vms need to be recreated so BOSH agent can get the new certs. 

This was proposed https://github.com/cloudfoundry/bosh/discussions/2309 and is the PR for the changes on the director side

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/instance.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/instance.rb
@@ -191,6 +191,15 @@ module Bosh::Director
           'trusted_certs' => Config.trusted_certs,
           'disk_associations' => disk_associations,
         }
+        if blobstore_config_changed?
+          blobstore = App.instance.blobstores.blobstore
+          blobstore.validate!(env['bosh']['blobstores'].first.fetch('options', {}), @stemcell.api_version)
+          if blobstore.can_sign_urls?(@stemcell.api_version)
+            settings['blobstores'] = blobstore.redact_credentials(env['bosh']['blobstores'])
+          else
+            settings['blobstores'] = env['bosh']['blobstores']
+          end
+        end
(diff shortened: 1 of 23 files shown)


bgandon (approved):
I've learnt a lot while reading that diff, thanks. This looks like a very neat piece of changes, I appreciate the effort in submitting readable individual commits. Approving with no remarks.

src/bosh-director/lib/bosh/director/deployment_plan/instance.rb
```diff
@@ -187,8 +187,38 @@ def update_instance_settings
...
+          settings['mbus'] = {
+            'cert' => {
+              'ca' => Config.nats_server_ca,
+              'certificate' => agent_cert_key_result[:cert].to_pem,
+              'private_key' => agent_cert_key_result[:key].to_pem,
+            }
+          }
+        end
+
+        AgentClient.with_agent_id(vm.agent_id, @model.name).update_settings(settings)
```
beyhan, line 216: In this class `agent_client.*` is used only this place doesn't use it anymore. Maybe we should add a comment to make sure why it doesn't use `agent_client.*`.

jpalermo:
> What kind of tests have been executed with this change?

We've run all of the unit tests and we've done manual testing of this change and the bosh agent change. We've tested new director with old agent, old director with new agent, and new director with new agent.

We've also tested a full certificate/credential rotation with new director and new agent to verify the happy path works and no VM recreations are needed to rotate the nats and blobstore certificates and credentials.

We have NOT run all of the integration tests or Bosh Acceptance Tests. We will probably not run those before merging as they are very expensive to setup and run. We'll instead fix any bugs uncovered by failures in those test suites after the fact (but before we release a new version of the bosh director)

PR #2324, opened 2021-09-29
add chruby

ramonskie:
add chruby to the main-ruby-go docker image.

it seems it was removed since we moved from the old-docker files


bgandon (approved):
LGTM

PR #2326, opened 2021-09-30
Improve description for director's job properties

bgandon:

Here we improve the description for some director's job properties related to compiled package cache

The property `provider` is actually unnecessary when `blobstore_path` is provided.

Unit tests for the BOSH Release templates are passing, as executed with `(cd src && bundle exec rspec ../spec)`.

These minor changes should not be necessary to appear in the release notes.

No, this is definitely related to documentation only.

Co-Authored-By: @olivermautschke

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -429,10 +429,15 @@ properties:
 
   # Compiled Package Cache
   compiled_package_cache.provider:
-    description: Provider of the blobstore used for the compiled package cache
+    description: |
+      Provider of the blobstore used for the compiled package cache.
+      Possible values are: 'local', 's3' and 'gcs'.
+      A value of 'local' is implied (and thus optional) when blobstore_path is provided.
     default: 's3'
   compiled_package_cache.options.bucket_name:
-    description: AWS S3 Bucket used for the compiled package cache
+    description: AWS S3 or GCS Bucket used for the compiled package cache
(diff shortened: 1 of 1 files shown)


jobs/director/spec
```diff
@@ -429,10 +429,15 @@ properties:
...
-    description: Provider of the blobstore used for the compiled package cache
+    description: |
+      Provider of the blobstore used for the compiled package cache.
+      Possible values are: 'local', 's3' and 'gcs'.
+      A value of 'local' is implied (and thus optional) when blobstore_path is provided.
     default: 's3'
   compiled_package_cache.options.bucket_name:
-    description: AWS S3 Bucket used for the compiled package cache
+    description: AWS S3 or GCS Bucket used for the compiled package cache
+  compiled_package_cache.options.blobstore_path:
```
rkoster, line 439: Why was this field added? It seems to go beyond improving the description of job properties? Not saying it is wrong just lacking context here.

bgandon: Yes, this property was already existing and supported, but missing here in the spec, and thus not documented.

It has been added here because it has an impact on `compiled_package_cache.provider`. So, it's mentionned at line 435 above: “_A value of 'local' is implied (and thus optional) when blobstore_path is provided._”

This explains why`compiled_package_cache.options.blobstore_path` has been documented here.

PR #2327, opened 2021-09-30
Optional Blobstore agent creds

bgandon:

Here we have the agent credentials for local Blobstore become optional, because implementing [signed URLs](https://bosh.io/docs/director-blobstore-signed-urls/) means they are not required anymore.

When operators opt for [signed URLs](https://bosh.io/docs/director-blobstore-signed-urls/), they should not be forced to specify any username and password for the agent to access the blobstore.

Unit tests for the BOSH Release templates are passing, as executed with `(cd src && bundle exec rspec ../spec)`.

- When using [signed URLs](https://bosh.io/docs/director-blobstore-signed-urls/), the credentials for the agent to access the local Blobstore are not required anymore.

No breaking change is introduced here, but missing agent credentials with a `blobstore.enable_signed_urls` of `false` is a new potential pitfall.

Co-Authored-By: @OliverMautschke


jobs/blobstore/templates/read_users.erb
```diff
@@ -1,9 +1,12 @@
 <%=
 all_users = {
-	p('blobstore.agent.user') => p('blobstore.agent.password'),
-	p('blobstore.director.user') => p('blobstore.director.password'),
+  p('blobstore.director.user') => p('blobstore.director.password'),
 }
 
+if_p('blobstore.agent.user') do
+  all_users[p('blobstore.agent.user')] = p('blobstore.agent.password')
+end
```
beyhan, line 8: I thought that we want to make these properties depend on `blobstore.enable_signed_urls`?

rkoster: If `blobstore.agent.user` is not set and `blobstore.enable_signed_urls` is false it should raise an error. In order to shorten the operator feedback loop in the case of a misconfiguration.

KaiHofstetter: I think we shouldn't configure the `blobstore.agent.user` if `blobstore.enable_signed_urls` is set to true.
The agent doesn't need this credential anymore in this case. And from a security perspective, it's better to not configure it so that you don't need to rotate it.

We currently have the issue that some CPIs still reference the `blobstore.agent.user`, `blobstore.agent.password` even though they don't use it if `blobstore.enable_signed_urls` is set to true.

Till these CPIs have been updated, we could set blobstore.agent.user`, `blobstore.agent.password` in the manifest and still wouldn't need to rotate the password, since it is not configured in the blobstore.

docs/running_tests.md
```diff
@@ -134,6 +135,23 @@ To focus on a given spec file you can use the environment variable `SPEC_PATH`
...
+### Run the ERB unit tests
+
+Install the Gem dependencies.
+```
+(cd src && bundle install)
+```
+
+Run the ERB unit tests.
+```
+scripts/test-unit-erb
```
beyhan, line 152: This should be `../scripts/test-unit-erb` because we end up in the `bosh/src` folder after executing the previous command `cd src && bundle install` or specify that it needs to be executed from the `bosh` folder.

docs/running_tests.md
```diff
@@ -134,6 +135,23 @@ To focus on a given spec file you can use the environment variable `SPEC_PATH`
...
+
+## BOSH Release tests
+
+The BOSH Release
+
+### Run the ERB unit tests
+
+Install the Gem dependencies.
+```
+(cd src && bundle install)
```
beyhan, line 147: Why these commands are in parenthesis?

jobs/blobstore/templates/read_users.erb
rkoster: When mixing old and new stemcells it could in theory be that both signed URLs and non agent blobstore credentials are used at the same time: https://github.com/cloudfoundry/bosh/blob/master/src/bosh-director/lib/bosh/blobstore_client/base.rb#L122-L125

Apparently the ubuntu-xenial/456.x line (which is still supported for VMware customers) does not support singed URLs ([API version 2](https://github.com/cloudfoundry/bosh-linux-stemcell-builder/blob/ubuntu-xenial/456.x/bosh-stemcell/lib/bosh/stemcell/stemcell_packager.rb#L45)). 
Stemcell builder was bumped to API version 3 in this commit: https://github.com/cloudfoundry/bosh-linux-stemcell-builder/commit/f4de1ec001c4dd653f24d7ba52efa6a31f139e19

docs/running_tests.md
bgandon: Well that wasn't the idea, because `(cd src && bundle install)` changes the current working directory for the sub-shell only.

bgandon:
I've pushed some changes to take your remarks into account @beyhan

PR #2330, opened 2021-10-10
Adds test-running and minimum Ruby version docs

kennethlakin:
This commit tells folks the minimum version of Ruby required to run the
tests (and -presumably- run the software under test), as well as how to
run the unit tests.

Because neither the CONTRIBUTING file, nor the README file currently
contain this information, I'm putting it in the README file.

Feel free to move the information to another file if that's more appropriate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,3 +26,17 @@ Cloud Foundry BOSH is an open source tool for release engineering, deployment, l
 ## Contributions
 
 Please read the [contributors' guide](https://github.com/cloudfoundry/bosh/blob/master/CONTRIBUTING.md)
+
+## Minimum software requirements
+
+Ruby 3.0.2 or later
+
+## Running the tests
+
+The code and all tests live in the `src` directory. All commands assume that `src` is your current working directory, and that `bundle install` has been run from that directory to install all required Ruby gems.
+
+### Unit Tests
+
+Either run `bundle exec rake unit:spec` or `bundle exec rake unit:spec:parallel`.
+
+The latter will run all unit tests, and should spread the load across all of the logical CPU cores that you have on your system.


kennethlakin:
I strongly suggest that if you don't mention how to run the tests in the project's README, that you _do_ mention them somewhere in your "How to contribute" file, which -in this case- is referenced in your README file.

For a good example of how one would expect this to be structured, check out the relevant docs for the Erlang/OTP project:

The readme here: <https://github.com/erlang/otp#contributing> links to the CONTRIBUTING file, which contains this lovely section: <https://github.com/erlang/otp/blob/master/CONTRIBUTING.md#before-you-submit-your-pull-request>. That section explicitly links to the docs on how to run the tests, and also mentions all of the other things you'll need to do before submitting a patch. (I've gone through the process for a couple of patches, it's well-explained and pretty easy to do.)

How would you feel about having someone update `bosh`'s CONTRIBUTING file to -at a minimum- point to the test-running docs?

(It should probably also be updated to either remove the reference to the `release:create_dev_release` `rake` task that doesn't work, or be updated to explain from _where_ one should run that task.)

beyhan:
I also agree that from the CONTRIBUTING guide should be clear for the contributors which tests they should ran before creating a pull request. At the moment this is unclear und not easy to find. @rkoster What do you think?

beyhan:
Unfortunately, I couldn't use this pr to propose add my suggestion. In https://github.com/cloudfoundry/bosh/pull/2335 I implemented the proposal in my previous comment.

rkoster:
@kennethlakin can this PR be closed now that #2335 has been merged? Since it contains most of the changes proposed here?

PR #2336, opened 2021-11-02
Use IO#readline_nonblock instead of IO#readline when reading output from CPI sub-processes.

yatzek:
IO.#readline, being a blocking method, can cause a dealock when the cpi
sends output without a line break. We had such issues with azure cpi.

### DIFF
--- a/src/bosh-director/lib/cloud/external_cpi.rb
+++ b/src/bosh-director/lib/cloud/external_cpi.rb
@@ -142,6 +142,8 @@ module Bosh::Clouds
             end
           rescue EOFError
             files.delete f
+          rescue IO::WaitReadable
+            # do nothing
           end
         end
         exit_status = wait_thr.value.exitstatus
--- a/src/bosh-director/lib/cloud/external_cpi.rb
+++ b/src/bosh-director/lib/cloud/external_cpi.rb
@@ -1,6 +1,28 @@
 require 'membrane'
 require 'open3'
 
+# Robbed from:
+# https://github.com/Homebrew/brew/blob/master/Library/Homebrew/extend/io.rb
+class IO
+  def readline_nonblock(sep = $INPUT_RECORD_SEPARATOR)
+    line = +""
+    buffer = +""
+
+    loop do
+      break if buffer == sep
+
+      read_nonblock(1, buffer)
+      line.concat(buffer)
+    end
+
+    line.freeze
+  rescue IO::WaitReadable, EOFError => e
+    raise e if line.empty?
+
+    line.freeze
+  end
+end
+
 module Bosh::Clouds
   class ExternalCpi
     # Raised when the external CPI executable returns an error unknown to director
@@ -110,7 +132,7 @@ module Bosh::Clouds
 
           readable = ready[0]
           readable.each do |f|
-            line = f.readline
+            line = f.readline_nonblock
             case f.fileno
             when stdout.fileno
               cpi_response.write(line)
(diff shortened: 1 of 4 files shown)


src/bosh-director/spec/unit/cloud/external_cpi_spec.rb
```diff
@@ -644,4 +644,20 @@ def self.it_raises_an_error_with_ok_to_retry(error_class)
   describe '#info' do
     it_calls_cpi_method(:info)
   end
+
+  context 'Fix for a dealock scenario when a cpi sub-process sends an inclomplete line (missing "\n") to STDOUT' do
```
rkoster, line 648: This test still passes for me when I use `readline` instead of `readline_nonblock`. I would have expected it to fail.

rkoster (changes requested):
Thanks for the PR, seems like a great improvement. However, when testing test change locally I found that changing back to the original implementation did not make the test fail.

src/bosh-director/spec/unit/cloud/external_cpi_spec.rb
yatzek: Well, the test just hangs because the whole thing deadlocks.

yatzek: Hm, interesting, I tried this on my machine yesterday and it deadlocked straight away.

yatzek: How about a timeout? It should throw an exception after 20 seconds and fail the test.

src/bosh-director/lib/cloud/external_cpi.rb
```diff
@@ -1,6 +1,28 @@
...
+
+    loop do
+      break if buffer == sep
+
+      read_nonblock(1, buffer)
+      line.concat(buffer)
+    end
+
+    line.freeze
+  rescue IO::WaitReadable, EOFError => e
```
beyhan, line 19: Should we `rescue` also `IO::WaitReadable` in https://github.com/cloudfoundry/bosh/blob/f4ae87f7e2ff696ad9048d2a58bdcd6b9a2eee8e/src/bosh-director/lib/cloud/external_cpi.rb#L121 or did we leave it out on purpose?

beyhan (changes requested):
My summary:
- all unit tests ran successful on my MacOS. 
- leaving the new `does not deadlock` unit test and reverting to the old code failed the test

@rkoster  is the failure with the unit test reproducible on your machine?

src/bosh-director/lib/cloud/external_cpi.rb
yatzek: @rkoster `bundle exec rspec spec/unit/cloud/external_cpi_spec.rb:648` from `src/bosh-director` folder worked for me.

yatzek: @beyhan good catch. Looks like we should rescue IO::WaitReadable and do nothing with it so that IO.select can retry the read, as described here: https://github.com/ruby/ruby/blob/98cc6d158b501a4c30aa578239b35085a4612b94/io.rb#L36

rkoster: running from `src/bosh-director` did the trick, was able to confirm the unit test indeed fails when reverting the nonblocking change (and pass when applied).

rkoster (approved):

beyhan (approved):

rkoster:
Thanks @yatzek!

Issue #1478, opened 2016-10-14
Failure when compiling packages named the same, with the same package/blob checksums, from two releases


geofffranks:
Tried deploying these two releases together in the same deployment today on my bosh-lite:

```
releases:
- name: cf-rabbitmq
  version: 215.9.0
- name: rabbitmq-metrics
  version: 1.29.0
```

They both have a `golang` package, and each package's `packaging` and `golang/go1.6.1.linux-amd64.tar.gz` files have the same checksums:

```
2614edf1b19b46a86272be883eb1d1bffb3c1c76  golang-rmq-metrics/packaging
2614edf1b19b46a86272be883eb1d1bffb3c1c76  golang-rmq/packaging
c73492474dcc7a233772d27e948b74468d9f1849  golang-rmq-metrics/golang/go1.6.1.linux-amd64.tar.gz
c73492474dcc7a233772d27e948b74468d9f1849  golang-rmq/golang/go1.6.1.linux-amd64.tar.gz
```

 When I deployed, I got the following error from BOSH:

```
E, [2016-10-14 17:58:29 #7597] [task:535] ERROR -- DirectorJobRunner: object id golang-1a4729b73e51c23d864b5af2b73c406ce857c7c7 is already in use
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/blobstore_client-1.3232.2.0/lib/blobstore_client/local_client.rb:20:in `create_file'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/blobstore_client-1.3232.2.0/lib/blobstore_client/base.rb:27:in `create'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/blob_util.rb:44:in `block (2 levels) in save_to_global_cache'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/blob_util.rb:43:in `open'
/var/vcap/packages/director/gem_home/ruby/2.1.0/gems/bosh-director-1.3232.2.0/lib/bosh/director/blob_util.rb:43:in `block in save_to_global_cache'
/var/vcap/packages/ruby/lib/ruby/2.1.0/tmpdir.rb:88:in `mktmpdir'
[...]

Issue #1300, opened 2016-06-14
use compiled package cache as a fallback and ignore fetching errors


sklevenz:
Hi,

We make use of the [compiled package cache](https://bosh.io/jobs/director?source=github.com/cloudfoundry/bosh#p=compiled_package_cache) feature and found out that bosh fails in any case of problem with this cache. Problems are that there is a S3 change in policy setting, user credentials or network issues. In the worst case a change in AWS can break all deployments.

Because of the cache is only an optimization I would like to suggest to handle issues with the cache as minor or warning issues and start compiling instead of break the deployment. 

WDYT?

Regards,
Stephan

cppforlife:
@sklevenz sounds like a good suggestion.

mfine30:
Given some roadmap items and the overall shift to moving towards compiled releases, this is not a big priority for us. I'm going to close it, but if you'd like to continue the conversation, feel free to reopen.

Issue #353, opened 2013-07-08
Publish light stemcells for Openstack via CI


rkoster:
A while back work was started on creating light stemcells for Openstack https://github.com/cloudfoundry-community/bosh-bootstrap/issues/200. 
The last step is for CI to publish these light-stemcells.

Currently these stemcells can be created with the following steps:
1. download the latest full stemcell
2. untar it, and then untag the image file _You should see a root.img file_
3. upload the root.img file to whatever remote location (i.e. S3)
4. delete the image file, and touch image _this will create a 0 bytes image size_
5. modify the stemcell.MF and add under cloud_properties a image_location option with the url to the uploaded root.img
6. tar it all up again and you are done

Credits for above instruction go to @frodenas

PR #2338, opened 2021-11-17
removed global package cache as it was causing issues

nouseforaname:

[GH Fixes: #1478, #1300, #353]

[finishes #3494307298](https://www.pivotaltracker.com/story/show/176753258)

```
cd src/
bundle install 
bundle exec rake spec:unit
```

Remove obsolete & problematic code path. Obsolete config value `compiled_package_cache`

It makes a property on the director spec obsolete and removes support for it


src/bosh-director/spec/unit/deployment_plan/stages/package_compile_stage_spec.rb
```diff
@@ -906,16 +905,12 @@ def compile_package_with_url_encrypt_stub(args)
       before do
```
beyhan, line 905: It looks like that we should remove the `describe 'the global blobstore'` completely here. The tests doesn't make much sense or do I miss something here?

beyhan (changes requested):

src/bosh-director/spec/unit/deployment_plan/stages/package_compile_stage_spec.rb
nouseforaname: you're right. removed it.

beyhan (approved):

rkoster (approved):
LGTM

rkoster:
Thanks @nouseforaname

PR #2339, opened 2021-11-17
Aborts errand run if instance(s) are stopped

lnguyen:
A customer discovered a problem when running an errand on a VM that
was stopped with `bosh stop`. It is our understanding that Bosh will
never `stop` a VM automatically; this is a manual process that an
operator must perform.

Because we can't know why an operator stopped an instance that might
be used to run an errand, it's safer to just halt the planning process
and raise an error that informs the operator that manual intervention
is required.

We're not sure if this behavior is correct, thoughts?

### DIFF
--- a/src/bosh-director/lib/bosh/director/errand/errand_provider.rb
+++ b/src/bosh-director/lib/bosh/director/errand/errand_provider.rb
@@ -24,6 +24,18 @@ module Bosh::Director
       matcher = Errand::InstanceMatcher.new(requested_instances)
       instances, unmatched_filters = matcher.match(instances_from_db)
 
+      # Check if any of the instances are stopped because we don't know why an operator stopped an instance,
+      # maybe it might be colocated with other jobs to save time on rebooting vms but forgot it was stopped.
+      # Since we don't know why an instance is stopped, it is safer to error out.
+      stopped_instances = []
+      instances.each do |instance|
+        stopped_instances << "'#{instance.name}'" if instance.stopped?
+      end
+      unless stopped_instances.empty?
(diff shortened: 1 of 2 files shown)


klakin-pivotal:
So, a few things:
* You might be asking "When would you ever have multiple instances on which to run an errand?". We're not sure, but `instances` is an array, and writing the code to handle more than one instance was not notably more difficult than writing the code to handle one. We tested the relevant code by pre-populating the array with a value, and then stopping an instance.
* The nasty use of `unless` and the single line `if` statement are in there because `rubocop` wouldn't let us commit unless we used that style. IMO, it would be nice if we removed those `rubocop` rules, but this is just me complaining.

beyhan:
This look to me more related to the use case where an errand is running on an existing instance group (colocated). Have you tried this with an errand with a dedicated instance group? Here the VMs are usually stopped when the errand is finished.

lnguyen:
@beyhan yeah we tested it with the following
```
bosh run-errand hello --keep-alive
bosh stop hello
bosh run-errand hello # new code errors out
```

This was a bug reported by customer, the follow steps to reproduce
```
bosh run-errand hello --keep-alive
bosh stop hello
bosh run-errand hello # errand succeed and vm is deleted
bosh run-errand hello # vm is created but errand fails to run
```

rkoster (changes requested):
Would it be possible to add at least a unit test for this?

jpalermo (approved):

jpalermo (approved):

beyhan (approved):

rkoster (approved):

PR #2340, opened 2021-11-17
Fix config rollback behavior.

camilalonart:
- Currently, if you add another version of an existing config (same name
  and type), then delete that config, when deploying, bosh won't find
  the prior config and will remove it from the deployment entirely. This
  is surprising behavior because bosh configs correctly lists the prior
  version. This commit fixes that behavior so that bosh will use the
  prior version configuration in this scenario.
[#179020742]

Signed-off-by: Rajan Agaskar <ragaskar@vmware.com>

### DIFF
--- a/src/bosh-director/lib/bosh/director/models/config.rb
+++ b/src/bosh-director/lib/bosh/director/models/config.rb
@@ -9,7 +9,7 @@ module Bosh
         end
 
         def self.latest_set(type)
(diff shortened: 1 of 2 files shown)


lnguyen (approved):

beyhan (approved):

rkoster:
Thanks! @camilalonart

Issue #2331, opened 2021-10-22
No_Proxy settings getting ignored by Health Monitor


benjaminguttmann-avtq:
**Describe the bug**
Since the update to BOSH version v271.9.0 (we are using the commit 36f5c668b57bc2ebe53fc3404cc72b74bb0beb9e of bosh-deployment) we do not get any metrics from the graphite health_monitor plugin anymore. After checking the logs we can see that the health_monitor gets blocked by the proxy when trying to access localhost:25555. This is strange as localhost as well as 127.0.0.1 are part of the no_proxy config. When switching back to v271.8.0 the issue disappeared again. We also tried the last version at this point v2.13.0 which did not solve the issue. Funny enough as it seems that the proxy setting itself gets applied but not the no_proxy one.

**To Reproduce**
Steps to reproduce the behavior (example):
1. Deploy a bosh director on openstack with a http_proxy configured, the proxy needs to block access to localhost. no_proxy settings need to include localhost and 127.0.0.1
2. Deploy any deployment so that health_monitor has something to check
3. Check the logs of health_monitor on the director VM, it should have an error page from proxy that states that access to https://localhost:25555/info has been blocked.

**Expected behavior**
I would expect that access localhost is possible because no_proxy setting is getting applied.

**Versions (please complete the following information):**
 - Infrastructure: openstack
 - BOSH version 271.9.0 - 271.13.0
 - BOSH CLI version 6.4.5-197a25ae-2021-08-19T18:47:22Z
 - Stemcell version bosh-openstack-kvm-ubuntu-bionic-go_agent-raw 1.1
 - uaa: 75.5.0
 - credhub: 2.9.0
 - bosh-openstack-cpi: 45 
 - bpm: 1.1.12
 - os-conf: 22.1.1

**Deployment info:**
bosh-deployment: 36f5c668b57bc2ebe53fc3404cc72b74bb0beb9e

rkoster:
I'm assuming this ops-file is being used: https://github.com/cloudfoundry/bosh-deployment/blob/master/misc/proxy.yml
```
- type: replace
  path: /instance_groups/name=bosh/properties/env?
  value: &env
    http_proxy: ((http_proxy))
    https_proxy: ((https_proxy))
    no_proxy: ((no_proxy))

- type: replace
  path: /cloud_provider/properties/env?
  value: *env
```

This uses global properties which is why the proxy settings are picked up by the health monitor component.
For this case, I would recommend using job properties like for example so:
```
[...]

PR #2342, opened 2021-11-26
Add no_proxy handling

nouseforaname:

EM HttpRequest does not adhere to the no_proxy env var.
added code that checks no_proxy and matches request target to no proxy entries.
if a match is found, it skips setting :proxy on the request  object to avoid EM HttpRequest limitations

fixes: https://github.com/cloudfoundry/bosh/issues/2331

```
cd src/bosh-monitor
bundle exec rspec
```

updated logic around EM::HttpRequest to only set request[:proxy] if the request target does not match an entry in no_proxy

It should not


src/bosh-monitor/lib/bosh/monitor/director.rb
```diff
@@ -68,7 +68,15 @@ def perform_request(method, uri, options = {})
       headers = {}
       headers['authorization'] = auth_provider.auth_header unless options.fetch(:no_login, false)
 
-      http = EM::HttpRequest.new(target_uri).send(method.to_sym, head: headers)
+      unless use_proxy?(target_uri, ENV['no_proxy'] || ENV['NO_PROXY'] || '')
+        conoptions = HttpConnectionOptions.new(target_uri, proxy: { host: '127.0.0.1', port: 25555, type: :http })
+        conoptions.https = false
+      end
+
+      http_con = EM::HttpConnection.new
```
klakin-pivotal, line 76: I intended to add some comments to this place in the code, but this is in your personal Github "organization" and I don't have privs to push, so I leave a comment here.

I would add comments that say two things:
1) This code is ugly, but it exists to add in handling for the `NO_PROXY` environment variable. The intention is for the code to check if the host is in the `NO_PROXY` list, and -if it is- create a special `HttpConnectionOptions` object that is configured so that `em-http-request` will treat the host as a plain HTTP "proxy", and just talk HTTP(S) directly to it, and replace the HttpConnectionOptions object that `em-http-request` makes for the `HttpConnection` with our specially-constructed one.
2) To future programmers: If you can find a reasonable way to call `HttpRequest.new` and add our special `HttpConnectionOptions` to the connection that it returns, rather than having code that duplicates what the `HttpRequest` initializer does, then do that. 

I guess if _you_ can verify that you can do thing #2 now, then doing that would be better than leaving a comment encouraging others to do it in the future.

bgandon:
Could we explicit why we've closed this PR?

Is there any way to contribute `NO_PROXY` support to EM HttpRequest?
Has the shim code here been submitted anywhere else, like in cloudfoundry/bosh-utils?

beyhan:
It looks like this has been addressed in https://github.com/cloudfoundry/bosh/pull/2344

bgandon:
Much cleaner code in #2344 indeed, thanks for the answer Beyhan 👍

PR #2344, opened 2021-12-08
Update request helper with no_proxy parsing

nouseforaname:
What is this change about?

EM HttpRequest does not adhere to the no_proxy env var.
added code that checks no_proxy and matches request target to no proxy entries.
if a match is found, it skips setting :proxy on the request object to avoid EM HttpRequest limitations
Please provide contextual information.

fixes: #2331
What tests have you run against this PR?

cd src/bosh-monitor
bundle exec rspec

updated logic around EM::HttpRequest to only set request[:proxy] if the request target does not match an entry in no_proxy

It should not


jpalermo (approved):
👍

lnguyen (approved):
👍

PR #2347, opened 2022-01-05
Speed up templating rendering

lnguyen:
Speed up template rendering by caching the properties of a given instance_group/job, in some cases you have alot of variables in credhub and rendering of properties templates is done for each instance in an instance_group and goes to credhub (which is cache by variable) so in the case of having 1k certs it takes 2~ seconds to render by the number of instances, in large production environments this causes a really slow template rendering.

This was tested with 200 instances in one instance group with 1k ca certs, this took 7:02 mins to render templates, with changes it goes down to 47 seconds. This will greatly help huge production environments with tons of credhub variables. 

Signed-off-by: Kenneth Lakin <klakin@vmware.com>

### DIFF
--- a/src/bosh-director/lib/bosh/director/config_server/variables_interpolator.rb
+++ b/src/bosh-director/lib/bosh/director/config_server/variables_interpolator.rb
@@ -4,13 +4,14 @@ module Bosh::Director::ConfigServer
     def initialize
       @logger = Bosh::Director::Config.logger
       @config_server_client = ClientFactory.create(@logger).create_client
+      @cache_by_job_name = {}
     end
 
     # @param [Hash] template_spec_properties Hash to be interpolated
     # @param [Hash] deployment_name The deployment context in-which the interpolation will occur
     # @param [VariableSet] variable_set The variable set which the interpolation will use.
     # @return [Hash] A Deep copy of the interpolated template_spec_properties
-    def interpolate_template_spec_properties(template_spec_properties, deployment_name, variable_set)
+    def interpolate_template_spec_properties(template_spec_properties, deployment_name, instance_name, variable_set)
       if template_spec_properties.nil?
         return template_spec_properties
       end
@@ -24,7 +25,13 @@ module Bosh::Director::ConfigServer
 
       template_spec_properties.each do |job_name, job_properties|
         begin
-          interpolated_hash = @config_server_client.interpolate_with_versioning(job_properties, variable_set)
+          key = job_name + "-" + instance_name
+          if @cache_by_job_name.has_key?(key)
+            interpolated_hash = @cache_by_job_name[key]
+          else
+            interpolated_hash = @config_server_client.interpolate_with_versioning(job_properties, variable_set)
+            @cache_by_job_name[key] = interpolated_hash
+          end
           result[job_name] = interpolated_hash
         rescue Exception => e
           @logger.debug("Unable to render templates for job '#{job_name}'. Received error: #{e.inspect}")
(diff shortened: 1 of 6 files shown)


src/bosh-director/lib/bosh/director/config_server/variables_interpolator.rb
```diff
@@ -4,13 +4,14 @@ class VariablesInterpolator
     def initialize
       @logger = Bosh::Director::Config.logger
       @config_server_client = ClientFactory.create(@logger).create_client
+      @cache_by_job_name = {}
```
beyhan, line 7: Every BOSH task will have its own cache, which is used during the execution of that task and the cache is shared between all instances of that task.  The cache (the class `VariablesInterpolator`) gets garbage collected after the task is finished. So we don't have any long living cache in memory, which can cause creds rotation issues. Is my understanding correct?

src/bosh-director/lib/bosh/director/config_server/variables_interpolator.rb
```diff
@@ -24,7 +25,13 @@ def interpolate_template_spec_properties(template_spec_properties, deployment_na
 
       template_spec_properties.each do |job_name, job_properties|
         begin
-          interpolated_hash = @config_server_client.interpolate_with_versioning(job_properties, variable_set)
+          key = job_name + "-" + instance_name
+          if @cache_by_job_name.has_key?(key)
+            interpolated_hash = @cache_by_job_name[key]
+          else
+            interpolated_hash = @config_server_client.interpolate_with_versioning(job_properties, variable_set)
+            @cache_by_job_name[key] = interpolated_hash
```
beyhan, line 33: We don't need to care about synchronisation here because the worst what can happen is to call the config server multiple times. Is this assumption correct?

src/bosh-director/spec/unit/config_server/variables_interpolator_spec.rb
```diff
@@ -6,6 +6,7 @@
   let(:client_factory) { double(Bosh::Director::ConfigServer::ClientFactory) }
   let(:config_server_client) { instance_double(Bosh::Director::ConfigServer::ConfigServerClient) }
   let(:deployment_name) {'my_deployment_name'}
+  let(:instance_name) {'my_instance_name'}
```
beyhan, line 9: I would expect to see a new test case here which verifies the use of the cache?

beyhan:
@lnguyen @kennethlakin This change can help a lot. Thank you for looking into this! I added some comments to it.

kennethlakin:
(Tagging  @klakin-pivotal , my Corporate Github Account)

src/bosh-director/lib/bosh/director/config_server/variables_interpolator.rb
lnguyen: Yup on first call, we get properties from config server and cache just the properties block for each job in an instance group which shouldnt change

src/bosh-director/spec/unit/config_server/variables_interpolator_spec.rb
lnguyen: So we wrestled with this cause a lot of the calls are just mocked, This could be added if you think we would get a lot of value out of this

beyhan: If you looked into this and didn't see much value in adding the tests is fine with me. I just wanted to make sure that we don't forget them.

beyhan (approved):

rkoster:
@bgandon could you review this PR, or unassing yourself, so someone else will review it?

bgandon:
Yeah, I've just unassigned myself @rkoster

rkoster (approved):
LGTM

rkoster:
Thanks @lnguyen

PR #2348, opened 2022-01-11
add netcat for the use of ssh via proxy

ramonskie:
when using `BOSH_ALL_PROXY` we need a netcat version with `-x` option to ssh via proxy


rkoster (approved):

rkoster:
Thanks @ramonskie

PR #2334, opened 2021-11-01
Leverage the new BPM `shutdown_signal` feature in Postgres jobs

bgandon:
### What is this change about?

Now that BPM v1.1.14 is shipped and used in `bosh-deployment`, we can use the new `shutdown_signal` feature.

### Please provide contextual information.

- Using `SIGINT` to shutdown Postgres was recently implemented in: #2320
- The new BPM that ships with the `shutdown_signal` feature is out: https://github.com/cloudfoundry/bpm-release/releases/tag/v1.1.14
- This version has been adopted in `bosh-deployment`: https://github.com/cloudfoundry/bosh-deployment/blob/master/bosh.yml#L147-L150

### What tests have you run against this PR?

```
./scripts/test-unit-erb
```

### How should this change be described in bosh release notes?

- Refactored the Postgres jobs to leverage the new `shutdown_signal` feature provided by BPM v1.1.14+

### Does this PR introduce a breaking change?

No.

### Tag your pair, your PM, and/or team!

Solo-ed on this.

PR #2352, opened 2022-01-25
Cleanup of Postgres pre-start code

bgandon:
Hello dear Bosh fellows,

Turns out the Postgres `pre-start` scripts were executed twice: first by Bosh Agent as `root` on the VM with no restrictions, second inside the BPM context. This issue comes from commit 4ab31932f in the first place, maybe due to lack of review.

When fixing this unnecessary duplicate execution of `pre-start` scripts, then mounting the old persistent store directory with BPM is no more necessary. This is useful to prevent an empty `postgres-10` directory to re-appear at every startup, only because it is declared in BPM for purposes of migrating from v10 to v13.

I've also fixed some bugs related to typos, as the actual directory for Postgres 9.4 is `postgres-9.4` and not `postgres-9`. This was preventing Postgres 10 `pre-start` script from was properly removing the Postgres 9.4 old persistent data after successful upgrade.

Also, after we've merged #2334, some data repair code in migration from Postgres v9.4 to v10 (due to improper signal being previously used to stop Postgres) is no more necessary, so I've removed it.

Less importantly, I've reviewed the `pre-start` scripts for the 3 Postgres jobs. I've noticed that the word 'data dir' was used for persistent data, that go to the ' store' dir in Bosh parlance, whereas 'data' dir usually designates the ephemeral data. So, I'm proposing some improved wording in that regards, for cleaner code.

Best,
Benjamin

### DIFF
--- a/jobs/postgres-10/templates/pre-start.erb
+++ b/jobs/postgres-10/templates/pre-start.erb
@@ -5,10 +5,10 @@ set -eu
 PACKAGE_DIR=/var/vcap/packages/postgres-10
 PACKAGE_DIR_OLD=/var/vcap/packages/postgres-9.4
 
-STORE_DIR=/var/vcap/store
+PERSISTENT_DISK_DIR=/var/vcap/store
 
-DATA_DIR=$STORE_DIR/postgres-10
-DATA_DIR_OLD=$STORE_DIR/postgres-9.4
+STORE_DIR=${PERSISTENT_DISK_DIR}/postgres-10
+STORE_DIR_OLD=${PERSISTENT_DISK_DIR}/postgres-9.4
 
 USER='<%= p("postgres.user") %>'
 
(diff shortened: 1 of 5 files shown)


jobs/postgres/templates/pre-start.erb
```diff
@@ -21,47 +21,47 @@ fi
...
     echo "ERROR: Unable to Initialize Postgres DB"
     exit 1
   fi
 
-  echo "host all $USER 0.0.0.0/0 md5" >> $DATA_DIR/pg_hba.conf
+  echo "host all ${USER} 0.0.0.0/0 md5" >> "${STORE_DIR}/pg_hba.conf"
 
-  mkdir -p $DATA_DIR/pg_log
-  chown vcap:vcap $DATA_DIR/pg_log
+  mkdir -p ${STORE_DIR}/pg_log
```
rkoster, line 40: Missing quotes

jobs/postgres-10/templates/pre-start.erb
```diff
@@ -26,70 +26,47 @@ fi
...
-  touch /var/vcap/store/postgres-10/fresh
+  touch "${STORE_DIR}/fresh"
 
   if [ $? != 0 ]; then
     echo "ERROR: Unable to Initialize Postgres DB"
     exit 1
   fi
 
-  echo "host all $USER 0.0.0.0/0 md5" >> $DATA_DIR/pg_hba.conf
+  echo "host all $USER 0.0.0.0/0 md5" >> ${STORE_DIR}/pg_hba.conf
```
rkoster, line 43: Missing quotes

rkoster (changes requested):
Looks good, just a few missing quotes here an there

bgandon:
Thank you Ruben for the good catches!
I've extended the scope of the PR quite a lot today, which will require some more reviewing from you unfortunately.

rkoster (approved):
LTGM

lnguyen (approved):
LGTM

PR #2353, opened 2022-01-28
Exported Releases only contain first level job dependent packages

jpalermo:
Current if a package depends on another package, that is also exported.

Jobs never use these packages, and they don't even end up getting deployed to VMs so they are just making compiled release tarballs larger.


lnguyen:
LGTM

PR #2356, opened 2022-02-04
Update bundler-audit to latest version

nouseforaname:
our very old bundler-audit version does not play well with the
newer psych version. Trying to run bundler-audit currently produces:

```shell
/usr/local/lib/ruby/3.1.0/psych/class_loader.rb:99:in `find': Tried to load unspecified class: Date (Psych::DisallowedClass)

        from /usr/local/lib/ruby/3.1.0/psych/class_loader.rb:28:in `load'

        from (eval):2:in `date'
```


rkoster:
Thanks! @nouseforaname

PR #2362, opened 2022-03-04
Increase Key Size for generated test certs

nouseforaname:
on linux when you try running integration tests,
almost every test fails

with:

```sh
 Failure/Error: current_sandbox.reset

RuntimeError:
  Failed: 'echo '
  '\ \ \ \ \ \ \ \ \ \ SELECT\ pg_terminate_backend\(pg_stat_activity.pid\)'
  '\ \ \ \ \ \ \ \ \ \ FROM\ pg_stat_activity'
  '\ \ \ \ \ \ \ \ \ \ WHERE\ pg_stat_activity.datname\ \=\ \'dc294b9eec144ff4b22c52a3eaf3b9ff\'\ AND\ pid\ \<\>\ pg_backend_pid\(\)\;'
  '\ \ \ \ \ \ \ \  | psql postgres://postgres:@127.0.0.1:5432/dc294b9eec144ff4b22c52a3eaf3b9ff > /dev/null' from /home/kkiess/workspace/bosh/src, with exit status 512
```

but that is a red herring. what actually happens is:

```sh
cat tmp/integration-tests-workspace/pid-25162/sandbox/logs/error.log
2022/03/04 16:38:23 [warn] 25184#0: the "ssl" directive is deprecated, use the "listen ... ssl" directive instead in /home/kkiess/workspace/bosh/src/tmp/integration-tests-workspace/pid-25162/sandbox/nginx.conf:27
2022/03/04 16:38:23 [emerg] 25184#0: SSL_CTX_use_certificate("/home/kkiess/workspace/bosh/src/bosh-dev/assets/sandbox/ca/certs/server.crt") failed (SSL: error:140AB18F:SSL routines:SSL_CTX_use_certificate:ee key too small)
```

the nginx is not starting because the key size is too small.

this updates the script to generate 2048 byte keys and also
regenerates the current certs.


cunnie:
Looks good to me.

I searched through our other repos making sure these certs aren't being used elsewhere, and it seems that they aren't. Good work!

PR #2367, opened 2022-03-23
Cleanup postgres 9.4

ramonskie:
as postgres 9.4 is now almost 2 years EOL
and as it is and should not be used anymore
it seemed time to remove this from our codebase as wel.


rkoster:
Context: https://www.postgresql.org/support/versioning/

jpalermo (approved):
Added a couple of commits. 

One to fully remove the postgres 9 to postgres 10 migration logic from the postgres 10 job since it won't work without postgres 9.

And a second one to update the ci pipeline to add tests against postgres 13.

ci/pipeline.yml
```diff
@@ -1111,6 +1120,14 @@ resources:
       username: ((docker.username))
       password: ((docker.password))
 
+  - name: main-postgres-10-image
```
ramonskie, line 1123: this should also be renamed to 13

lnguyen (approved):
LGTM

PR #2370, opened 2022-04-07
IPv6 ambiguity issues

FlorianNachtigall:
### tldr
I believe currently only IPv6 addresses larger than `::ffff:ffff` are supported - all others will result in issues when using them together with IPv4 addresses. I added some tests to illustrate those issues and to serve as a starting point for a discussion how to fix it.

### What is this issue about?

IPv4 and IPv6 addresses have overlapping integer ranges. Since bosh stores all IP addresses in their integer representation in the database without storing any information about their IP version, this leads to ambiguity when retrieving the IP addresses and converting them back to their human-readable string representation.

More specifically, the current implementation considers all integers less than `2**32` as IPv4 addresses by default. Consequently, IPv6 addresses from `::` to `::ffff:ffff` are treated as or converted to IPv4 addresses for instance when [storing](https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb#L164) and [retrieving](https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb#L14) them from the database, making [network reservations](https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/network_reservation.rb#L33) or [looping over them with IpUtil's each_ip function](https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/ip_util.rb#L54-L57).

Illustration with vendored netaddr gem:
```
3.1.0 > NetAddr::CIDR.create('192.168.1.5').to_i
 => 3232235781
3.1.0 > NetAddr::CIDR.create('::c0a8:105').to_i
 => 3232235781
3.1.0 > NetAddr::CIDR.create(3232235781).ip
 => "192.168.1.5"
```
Resulting problems include:
* accidental deletion or blocking of IPv4/IPv6 addresses with same integer representation respectively
* all IPv6 addresses smaller than `::ffff:ffff` are essentially converted to IPv4 addresses

Please refer to the added unit tests for more detailed examples of what specific problems are caused by this ambiguity.

This PR introduces unit tests to ensure that this ambiguity is being handled in a selection of important cases, for instance IPv4 addresses are not accidentally deleted when IPv6 addresses should be deleted. However, the test suite is not complete in covering all possible cases of ambiguity. More importantly, the PR should serve as a starting point for further discussions.

### Example of allocating and releasing IP addresses based on their integer representation
https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb#L162-L164
https://github.com/cloudfoundry/bosh/blob/acf041f931f808057c043e7f02162a1fd021e996/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb#L11-L14

### Ideas how to fix / mitigate it

* storing IP version in database as additional auxiliary attribute
* storing string representation of IP addresses in database
* updating docs and announcing that only IPv6 addresses larger than `::ffff:ffff` are supported (when using them together with IPv4 addresses)

To a large extent, we leave it to the vendored netaddr gem to distinguish between IPv6 and IPv4 addresses and manage them under the hood. Using the `:Version` parameter of `NetAddr::CIDR` would allow us to explicitly distinguish the version.
```
3.1.0 > NetAddr::CIDR.create(3232235781, Version: 6).ip
 => "0000:0000:0000:0000:0000:0000:c0a8:0105"
3.1.0 > NetAddr::CIDR.create(3232235781).ip
 => "192.168.1.5"
```

* Newly introduced bosh-director unit tests fail as expected.

* No, only unit tests have been changed.

@friegger @cloudfoundry/cf-bosh-europe

### DIFF
--- a/src/bosh-director/spec/unit/deployment_plan/ip_provider/database_ip_repo_ipv6_spec.rb
+++ b/src/bosh-director/spec/unit/deployment_plan/ip_provider/database_ip_repo_ipv6_spec.rb
@@ -378,18 +378,27 @@ module Bosh::Director::DeploymentPlan
 
     describe :delete do
       before do
-        network_spec['subnets'].first['static'] = ['fdab:d85c:118d:8a46::5']
+        network_spec['subnets'].first['static'] = ['0000:0000:0000:0000:0000:0000:c0a8:0105']
 
-        reservation = BD::DesiredNetworkReservation.new_static(instance_model, network, 'fdab:d85c:118d:8a46::5')
+        reservation = BD::DesiredNetworkReservation.new_static(instance_model, network, '0000:0000:0000:0000:0000:0000:c0a8:0105')
         ip_repo.add(reservation)
       end
 
       it 'deletes IP address' do
         expect {
-          ip_repo.delete('fdab:d85c:118d:8a46::5')
+          ip_repo.delete('0000:0000:0000:0000:0000:0000:c0a8:0105')
         }.to change {
             Bosh::Director::Models::IpAddress.all.size
           }.by(-1)
+      end
+
+      it 'does not accidentally delete a IPv6 address when a IPv4 address is being deleted' do
+        expect(logger).to receive(:debug).with("[network-configuration] Skipping releasing ip '192.168.1.5': not reserved")
+        expect {
+          ip_repo.delete('192.168.1.5')
+        }.to change {
+            Bosh::Director::Models::IpAddress.all.size
+          }.by(0)
       end
     end
   end
--- a/src/bosh-director/spec/unit/deployment_plan/ip_provider/ip_repo_spec.rb
+++ b/src/bosh-director/spec/unit/deployment_plan/ip_provider/ip_repo_spec.rb
@@ -511,6 +511,15 @@ module Bosh::Director::DeploymentPlan
             Bosh::Director::Models::IpAddress.all.size
           }.by(-1)
       end
+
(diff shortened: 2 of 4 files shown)


rkoster:
I like the idea of storing IP version in database, where during the up migration all addresses below `::ffff:ffff` would be marked as ipv4 and everything above as ipv6. This should not cause any unexpected behaviour for existing users, but would allow using ipv6 addresses below `::ffff:ffff` in the future.

jpalermo (approved):
👍  I also like the idea of just storing the version.

Storing the string also seems probably fine, and would actually make human inspection of the database more sane perhaps, but it does seem more likely that some situation would come up in the future where we might regret using the string.

FlorianNachtigall:
Great, thanks for the feedback. 

I'd also agree that storing the IP version has the smallest potential for breaking things in the future and is the easiest to rollout and potentially rollback in case of problems.

If there are no objections from your side, I would go ahead and implement the functionality and append it to the PR.

Issue #2349, opened 2022-01-11
Running errands on stopped VMs is a valid scenario


friegger:
**Describe the bug**
The change https://github.com/cloudfoundry/bosh/pull/2339 preventing an errand to run on stopped VMs shipped with https://github.com/cloudfoundry/bosh/releases/tag/v271.17.0 is a breaking change. We've a scenario for a service deployment, where a co-located errand would restore the database after the processes have been stopped with `bosh stop`. We had to downgrade the Director to v271.16.0. The scenario might be less relevant for not co-located errands or hard stopped instances.

**To Reproduce**
Steps to reproduce the behavior (example):
Well this requires a specific release of course, but it should be clear from the description.

**Expected behavior**
Don't prevent to run errands against a stopped instance.

**Logs**
-
**Versions (please complete the following information):**
 - Infrastructure: all
 - BOSH version 217.17.0

mrosecrance:
The issue we were intending to fix is this:
1) bosh stop the vm
2) forgot about it and deploy more changes -> some jobs are now missing from the stopped vm
3) run an errand
4) see errors since the errand is expecting missing jobs on the older stopped vm

Can you share more details on why bosh stop is used as a regular flow? It may not be specifically relevant to this conversation but I think there's a pattern of usage here that I'm not familiar with.

This could potentially be added as a flag on run-errand like `--include-stopped` going forward

friegger:
@anoopjp Can you elaborate a little why it is implemented like it is?

@mrosecrance How much effort would it be to bring in such a flag? I still believe that running an errand on stopped jobs is a valid scenario.

mrosecrance:
It wouldn't be much effort but would mean we need a new bosh-cli. Is this something SAP has bandwidth to PR? 

We did this because we got a few reports of broken deployments and unexpected behavior but since this is a case you all use, I think it makes sense to support it (I would still argue with a flag rather than as default behavior though)

friegger:
@mrosecrance Yes, we would be able to provide this at some point in time. The flag would need to be represented in the [API](https://github.com/cloudfoundry/bosh/blob/b21cafff8aa969645bc771a76f6a51cddd7ba108/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb#L503) as well and handed to the job wouldn't it?

I first would like to understand from @AnoopJP whether this would be something they could adopt easily.

anoopjb:
[...]

PR #2374, opened 2022-05-06
Support running errands on stopped instances

FlorianNachtigall:
### PR following up discussion in #2349

Introduce a new director property to support running errands on stopped instances. The property defaults to false. If set to true, the property brings back old director behavior prior to #2339 which is still needed by some customers.

Please note that we didn't touch the bosh-dev sandbox because we believe integration tests are not needed for this property. It acts as a feature flag, so if not set it defaults to the current behavior.

Please let us know, if changes to the sandbox are still required.

Pair: @ShilpaChandrashekara
PM: @friegger 
Team: @cloudfoundry/cf-bosh-europe

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -179,6 +179,9 @@ properties:
   director.parallel_problem_resolution:
     description: When true, problems (e.g. resurrection, disk reattaching) are resolved in parallel
     default: true
+  director.allow_errands_on_stopped_instances:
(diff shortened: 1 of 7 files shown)


lnguyen (approved):
LGTM, I believe that the tests are really just looking at instance_model if I remember correctly

beyhan (approved):

PR #2380, opened 2022-05-23
Reset unresponsive VM count for deleted deployment

ystros:
We received report of a bug where the Prometheus metrics server
continued to report unresponsive VMs within a specific BOSH deployment
after that deployment had been deleted. This is because the metrics rely
on the Health Monitor `/unresponsive_agents` API endpoint, which only
reports statistics for current deployments. After the deletion, the
Health Monitor endpoint stopped reporting the deployment, so the
Prometheus updater kept the last reported metrics.

The Prometheus client gem does not provide a way to delete metrics
entirely; therefore, we set the unresponsive agent count to 0 instead.

[#182015765] Prometheus metrics are not cleaned up after a deployment is deleted

Signed-off-by: Brian Upton <bupton@vmware.com>


src/bosh-director/lib/bosh/director/metrics_collector.rb
```diff
@@ -116,9 +116,21 @@ def populate_vm_metrics
...
+          # data returned from values looks like:
+          # { { name: "deployment_a"} => 10, { name: "deployment_b "} => 0, ... }
+          key[:name]
+        end
+
+        unresponsive_agent_counts.each do |deployment, count|
           @unresponsive_agents.set(count, labels: { name: deployment })
         end
+
+        removed_deployments = existing_deployment_names - unresponsive_agent_counts.keys
```
danielfor, line 130: I like that we're zeroing just the ones that we need. If the removed deployments are still needed  in `@unresponsive_agents` with a zero, then IMHO this is a good solution.

danielfor (approved):
LGTM

PR #2381, opened 2022-05-25
Use director proxy_timeout also for metrics server

ionphractal:

Use proxy_timeout also for metrics server nginx configuration.

Depending on the size of the environment, the amount of metrics produced also increases and metrics gathering may take more than the default 60s proxy_read_timeout after which nginx responds with "504 Gateway Time-out".

After applying the change to a director the described symptom (504 Gateway Time-out)...
1. curl the /metrics endpoint
2. curl the /api_metrics endpoint

Use director proxy_timeout also for metrics server

No.

@benjaminguttmann-avtq


rkoster (approved):
LGTM

PR #2382, opened 2022-05-27
Make our ruby requirement just plain 3.1

jpalermo:
This will make it easier to auto bump the ruby package as we won't need
to also update these files each time.

I'm simply removing the ruby_version.rb, that doesn't seem to provide
much value these days over what is being done in the .ruby-version and
Gemfile files.


lnguyen:
LGTM

PR #2383, opened 2022-06-06
Do not fail if you have old postgres data dir if you've already upgraded

jpalermo:
If you have both postgres 9 and 10 data directories it currently fails. In the new
postgres 13 job we've handled this appropriately so you won't get failures.

We could `rm -rf` the postgres 9 folder also, but the postgres 13 job will do that
when users eventually move to that.


lnguyen (approved):
LGTM

Issue #2385, opened 2022-06-13
Bosh deployment fails with "Unknown alias: xyz"


mvach:
**Describe the bug**
Some of our deployments are facing an issue with yaml anchors. That fails the deployment at the very beginning while loading the manifest yaml with:

```I, [2022-06-06T01:38:32.086738 #5838] [task:300244]  INFO -- DirectorJobRunner: Reading deployment manifest
D, [2022-06-06T01:38:32.096112 #5838] [task:300244] DEBUG -- DirectorJobRunner: (0.000564s) (conn: 4320) BEGIN
D, [2022-06-06T01:38:32.097385 #5838] [task:300244] DEBUG -- DirectorJobRunner: (0.000708s) (conn: 4320) UPDATE "tasks" SET "event_output" = ("event_output" || '{"time":1654479512,"error":{"code":100,"message":"Unknown alias: ref_0"}}') WHERE ("id" = 300244)
```

**To Reproduce**
We are still not able to reproduce it but expect that the issue was introduces with [v272.1.0](https://github.com/cloudfoundry/bosh/releases/tag/v272.1.0)

**Additional context**
To mitigate the problem we rolled back to our last known good bosh version [v271.16.0](https://github.com/cloudfoundry/bosh/releases/tag/v271.16.0)

beyhan:
Great! Changed the status accordingly.

mvach:
I proposed a PR to fix the issue: https://github.com/cloudfoundry/bosh/pull/2386

mvach:
Since the issue is fixed I do close the issue

PR #2386, opened 2022-06-21
Fix regression: Deployment manifests can contain yaml anchors

mvach:
Manifest do support yaml anchors again. That support got lost as regression of bosh version v272.1.0.

With Ruby 3.1 Psych was updated to version 4 which comes with an incompatible change. The incompatible change is that YAML.load is not an alias to unsave_load anymore but to save_load instead.

The Issue for that PR is: https://github.com/cloudfoundry/bosh/issues/2385

I have executed the director unittests and deployed it onto one of our landscapes. Then I deployed with that fixed BOSH version a manifest that uses aliases.

Deployment manifests do support yaml anchors again.

Actually it fixes a breaking change

@fmoehler

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
@@ -4,7 +4,7 @@ module Bosh::Director
   module Api::Controllers
     class PackagesController < BaseController
       post '/matches', :consumes => :yaml do
-        manifest = YAML.load(request.body.read)
+        manifest = YAML.load(request.body.read, aliases: true)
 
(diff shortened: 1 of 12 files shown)


rkoster (changes requested):
Please add unit tests for this

mvach:
Hi @rkoster,
I added a unit test to update_deployment_spec.rb. This is of course not covering all the touched YAML.loads. Do you want me to test them one by one or is that single one already enough from your perspective? Testing them now one by one could become relatively expensive.

ramonskie (approved):
LGTM

rkoster (approved):

rkoster:
Thanks! @mvach

PR #2387, opened 2022-06-24
Make it clear that defaults apply to PG databases

nouseforaname:
The description for the connection options should mention that the
default settings are for postgres. Currently when someone deploys
the director to use a mysql db but does not specify any connection
options, the director job would pick up the defaults and pass the 
connection options to `Sequel` which then passes them to the mysql2
gem. The equivalent connection options for mysql have different names
and the default values wouldn't actually be usable for the underlying 
connection.

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -306,7 +306,9 @@ properties:
     description: Name of the director database
     default: bosh
   director.db.connection_options:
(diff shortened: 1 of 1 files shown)


ramonskie (approved):
lgtm

rkoster (approved):

PR #2391, opened 2022-07-18
Upgrading NATS to a supported version

danielfor:
- Updated NATS to a supported version: 2.8.3

- NATS 2.8.3 does not support wildcards in the users authorization
  section in the `nats.cfg` file. Wildcards were used previously to
  allow any bosh agents to publish/subscribe to NATS.
  Because of the new wildcards restrictions we need to add all the
  users explicitly in the configuration and that needs to be done
  continuously, as more VMs/Agents are added.
  To execute this continuous synchronization, we created the ruby
  gem 'bosh-nats-sync' which keeps the list of active agents up to
  date.

- Added the monit infrastructure to start the bosh-nats-sync gem.

The BOSH Director unit tests.

This is an update of NATS. This component used by the director was in an unsupported version.  

No, this won't affect users using BOSH with a standard NATS installation.
Warning: For BOSH installations with custom NATS configurations (e.g custom NATS server running outside of the director VM) the `bosh-nats-sync` component needs to be configured and run according to the customizations.

Co-authored-by: Brian Upton <bupton@vmware.com>
Co-authored-by: Seth Boyles <sboyles@vmware.com>
Co-authored-by: Brian Cunnie <bcunnie@vmware.com>
Co-authored-by: Manuel Alba <albam@vmware.com>
Co-authored-by: Long Nguyen <nguyenlo@vmware.com>
Co-authored-by: Daniel Felipe Ochoa <danielfelipo@vmware.com>


lnguyen:
@beyhan @jpalermo @rkoster could ya'll take a look at this at some point?

jpalermo (changes requested):
Did we do load testing on this yet? Feels like we'd want to make sure this scales fine up to ~2000 VMs or so.

jobs/nats/spec
```diff
@@ -40,6 +47,14 @@ properties:
...
   nats.tls.ca:
     description: CA cert used by NATS server to verify clients certificates (For Mutual TLS Connections)
+  nats.tls.health_monitor.certificate:
+    description: |
+      Certificate for establishing mutual TLS with NATS server. The Common-Name for the certificate
+      should be "default.hm.bosh-internal"
+  nats.tls.director.certificate:
+    description: |
+      Certificate for NATs mutual TLS client (Director client). The Common-Name for this certificate
+      should be "default.director.bosh-internal"
```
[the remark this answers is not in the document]
jpalermo: > It would be good if we could provide/consume these via links. That would prevent operators from having to specify the values twice in the manifest or use global properties.
> 
> Would also reduce errors by ensuring the values match.

Never mind, terrible idea. It looks like the bosh cli does NOT support links at all, so this would not work when doing a `bosh create-env` if we changed it to links.

jobs/nats/monit
```diff
@@ -3,3 +3,13 @@ check process nats
...
   stop program "/var/vcap/jobs/bpm/bin/bpm stop nats"
   group vcap
+
+check process bosh_nats_sync
+  with pidfile /var/vcap/sys/run/bpm/nats/bosh_nats_sync.pid
+  start program "/var/vcap/jobs/bpm/bin/bpm start nats -p bosh_nats_sync"
+  stop program "/var/vcap/jobs/bpm/bin/bpm stop nats -p bosh_nats_sync"
+  group vcap
+
+check file nats_auth_conf
```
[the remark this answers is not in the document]
danielfor: This is hard to do in any other way because of the usage of BPM. We cannot restart NATS from the bosh_nats_sync process because both of them are sandboxed. Once we replace Monit with something else we need to use that tool to restart NATS as well.

jpalermo:
@lnguyen could you note the performance tests you did here? How many "VMs" and how long it was taking to render?

jpalermo (approved):

jobs/nats/spec
```diff
@@ -49,3 +64,32 @@ properties:
...
+  #
+  # Bosh Director options (to get the list of managed VMs)
+  #
+
+  director.address:
+    description: Address of the Bosh Director to connect to
+  director.port:
+    description: Port of the Bosh Director to connect to
+    default: 25555
+  nats.director_account.user:
```
rkoster, line 77: Is there also a PR for bosh-deployment which adds all these new properties? Asking because we probably want to make sure some coördination happens around making sure these things get released at the same time.

jobs/nats/templates/nats.cfg.erb
```diff
@@ -8,52 +8,39 @@ logtime: true
...
-      "hm.agent.shutdown.*",
-      "hm.director.alert"
-    ]
-  }
-
-  certificate_clients: [
-    {client_name: director.bosh-internal, permissions: $DIRECTOR_PERMISSIONS},
-    {client_name: agent.bosh-internal, permissions: $AGENT_PERMISSIONS},
-    {client_name: hm.bosh-internal, permissions: $HM_PERMISSIONS},
+  users = [
```
rkoster, line 11: Are you sure users can be specified twice (it also exists in the dynamic auth.conf). Also it looks like these users already get added here: https://github.com/cloudfoundry/bosh/pull/2391/files#diff-5cd6181b35dea1b09c5ef32d5ddaaf004157132204bc1436ec2e3293656bce11R47-R48

rkoster (changes requested):
Looks good overall, just a small question with regards to the director and hm users (see comments above).

lnguyen:
> @lnguyen could you note the performance tests you did here? How many "VMs" and how long it was taking to render?

I ran 500 vms oh BOSH-lite and it was taking about 1 min per vm to come up

jobs/nats/spec
danielfor: @rkoster there's no PR yet, however, our development tasks for NATs 2 include one to modify bosh-deployment. As you mention coordination needs to happen between the creation of the BOSH release with NATs 2 and the changes on bosh-deployment.

jobs/nats/templates/nats.cfg.erb
danielfor: During testing we saw that they can be specified twice and NATs will work. 
The ones you see on nats.cfg.erb are some sort of failsafe in the case that the nats-sync app cannot access the BOSH API when it is still initializing, making the list on the dynamic auth.conf being empty. But having them specified twice is not really necessary.

rkoster (approved):

PR #2393, opened 2022-07-28
deleted bosh config leaves dangling hm rule

nouseforaname:
start with no existing bosh configs of type resurrection

create a config of type resurrection via `bosh updatei-config`healthmanager
will create a rule for that config that it tracks internally in the `@parsed_rules`
array.

```
number of bosh configs of type resurrection == 1
parsed rules array length == 1
```

run `bosh delete-config`

```
number of bosh configs of type resurrection == 0
parsed rules array length == 1
```

The problem is in the return early. While it makes sense to return
early if the argument for the method call is nil, it didn't make
sense to return early if the argument was set to an empty array.
Because we actually want the array to be empty if we try to delete
the last resurrection config.

Effectively this issue will cache the contents of the last rule. In a
scenario where the global config is ON and the contents of the last
rule were used to override that value for deployment XYZ, even after
deleting the rule, deployment XYZ would have the resurrection switched
off because the rule in health manager was never deleted even if the
bosh config didn't exists anymore.

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/resurrection_manager.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/resurrection_manager.rb
@@ -16,23 +16,27 @@ module Bosh::Monitor
     end
 
     def update_rules(resurrection_configs)
+      return if resurrection_configs.nil?
+
       new_parsed_rules = []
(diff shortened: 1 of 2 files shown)


ramonskie (approved):
lgtm

PR #2395, opened 2022-08-15
Add postgres package autobumping

nouseforaname:
we want to keep postgres packages up to date for the supported major
version lines. This task does exactly that. It checks the postgres git
repo for REL_* tags matching our used postgres versions and creates bosh
blobs from the src. It then replaces and uploads the packages used in
the release.

Implenents: [#182288132]
Fixes: CVE-2022-1552


lnguyen:
LGTM

PR #2398, opened 2022-09-14
Fix for the race condition related to NATs users

danielfor:
Every time that a new VM is added by BOSH the authentication file for NATs receives a new user so the BOSH Agent of that VM is able to authenticate to NATs. Changes in this file require a NATs reload. This reload was being made by monit, but it was proving to be unreliable specifically at director startup. This change is to remove the responsibility of restarting NATs from monit and move it to the bosh-nats-sync program, which is more flexible since we control its behavior.

NATs reload will happen in the following scenarios:
- NATs is up and running and bosh-nats-sync is starting.
- NATs is up and running and bosh-nats-sync finds a change in the bosh instances that are running.

If for some reason bosh-nats-sync is unable to reload NATs, it will raise an error and monit will restart it.

For the bosh-nats-sync to be able to reload NATs, privileges were given to its container, so it's able to send signals to the NATs server process.

Signed-off-by: Daniel Felipe Ochoa <danielfelipo@vmware.com>
Signed-off-by: Brian Upton <bupton@vmware.com>

### DIFF
--- a/src/bosh-nats-sync/lib/nats_sync/users_sync.rb
+++ b/src/bosh-nats-sync/lib/nats_sync/users_sync.rb
@@ -1,21 +1,34 @@
 require 'rest-client'
 require 'base64'
 require 'nats_sync/nats_auth_config'
+require 'open3'
 
 module NATSSync
   class UsersSync
-    def initialize(nats_config_file_path, bosh_config)
+    def initialize(nats_config_file_path, bosh_config, nats_server_executable, nats_server_pid_file)
       @nats_config_file_path = nats_config_file_path
       @bosh_config = bosh_config
+      @nats_server_executable = nats_server_executable
+      @nats_server_pid_file = nats_server_pid_file
     end
 
     def execute_users_sync
       NATSSync.logger.info 'Executing NATS Users Synchronization'
       vms_uuids = query_all_running_vms
+      current_file_hash = nats_file_hash
       write_nats_config_file(vms_uuids, read_subject_file(@bosh_config['director_subject_file']),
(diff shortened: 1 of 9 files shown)


src/bosh-nats-sync/lib/nats_sync/users_sync.rb
```diff
@@ -1,23 +1,36 @@
...
+      current_file_hash = nats_file_hash
       write_nats_config_file(vms_uuids, read_subject_file(@bosh_config['director_subject_file']),
                              read_subject_file(@bosh_config['hm_subject_file']))
+      new_file_hash = nats_file_hash
+      UsersSync.restart_nats_server(@nats_server_executable, @nats_server_pid_file) unless current_file_hash == new_file_hash
       NATSSync.logger.info 'Finishing NATS Users Synchronization'
       vms_uuids
     end
 
+    def self.restart_nats_server(nats_server_executable, nats_server_pid_file)
```
jpalermo, line 27: Everything looks good to me but it feels like we use the word `restart` a lot here and the nats server isn't being restarted (unless I'm understanding incorrectly). It's just getting a signal to reload the config. I'd recommend changing the wording so it's more clear.

jpalermo (approved):

lnguyen (approved):

src/bosh-nats-sync/lib/nats_sync/users_sync.rb
danielfor: Thanks, new name is `reload_nats_server_config` .

beyhan (approved):

PR #2409, opened 2022-11-09
Cache /info on nats sync

lnguyen:
The nats sync calls info endpoint on every call it does to /vms This is not ideal.
This is not ideal. This caches the info call as it doesn't change that often


nouseforaname (approved):
lgtm

danielfor (approved):
LGTM

PR #2413, opened 2022-11-28
make task state db calls more resilient against flaky pg. Add cleanup for orphaned tasks without a delayed job entry.

nouseforaname:
in some environments postgres may be overloaded or not working properly. This can lead to connections being unexpectedly closed or failing. In a scenario where a `delayed job` execuction finished (and the job was updated by the `delayed job` gem in the `delayed job table` the update for the task itself in the `tasks` table may fail. The worker will not pick up the task again (as the data in `delayed jobs` table indicates it finished, but the tasks table (which bosh uses to display `bosh tasks`) leaves the task in the `queued` or another state.

Fixes:
`Bosh::Director::Jobs::DBJob (id=897735) (queue=urgent) FAILED (0 prior attempts) with Sequel::DatabaseDisconnectError: PG::UnableToSend: server closed the connection unexpectedly`

Currently some of the task update logic was already wrapped in a repeatable transaction but the errors we've seen in the user-report indicated that updating the task fail with the above error.

While we cannot ensure that the task state will be updated in 100% of the cases, this PR  minimizes the chance that the task state will not be updated when the connection to postgres is flaky.

Additionally this adds discovery of such oprhaned tasks in the tasks cleanup jobs. Only tasks that still have a backing delayed job can possibly get their state updated. If we find no delayed job containing a handler yaml (which contains the task id for the delayed job) with a corresponding task_id, that indicates the task is orphaned and won't be updated anymore. It should be safe to mark these tasks as `error`ed. 

This behaviour can be reproduced by executing the below sql statement against the PG instance backing the director

```
SELECT
 pg_terminate_backend(pid)
FROM
 pg_stat_activity
WHERE
 -- don't kill my own connection!
 pid <> pg_backend_pid()
 -- don't kill the connections to other databases
 AND datname = 'bosh';
pg_terminate_backend
```

while running arbitrary bosh tasks.

### DIFF
--- a/src/bosh-director/lib/bosh/director/job_queue.rb
+++ b/src/bosh-director/lib/bosh/director/job_queue.rb
@@ -8,6 +8,7 @@ module Bosh::Director
       task = create_task(username, job_class.job_type, description, deployment, context_id)
 
       Delayed::Worker.backend = :sequel
+
       db_job = Bosh::Director::Jobs::DBJob.new(job_class, task.id, params)
       Delayed::Job.enqueue db_job
       task
--- a/src/bosh-director/lib/bosh/director/job_queue.rb
+++ b/src/bosh-director/lib/bosh/director/job_queue.rb
@@ -42,13 +42,7 @@ module Bosh::Director
       end
 
       task.output = log_dir
-      Config.db.transaction(
-        wait: 1.seconds,
-        attempts: 5,
-        retry_on: [Sequel::DatabaseConnectionError, Sequel::DatabaseDisconnectError],
-      ) do
-        task.save
-      end
+      task.save
 
       task
     end
--- a/src/bosh-director/lib/bosh/director/job_queue.rb
+++ b/src/bosh-director/lib/bosh/director/job_queue.rb
@@ -4,7 +4,7 @@ module Bosh::Director
   # Abstracts the delayed jobs system.
 
   class JobQueue
-    def enqueue(username, job_class, description, params, deployment = nil, context_id = '')
+    def enqueue(username, job_class, description, params, deployment = nil, context_id = '') # rubocop:disable Metrics/ParameterLists
       task = create_task(username, job_class.job_type, description, deployment, context_id)
 
       Delayed::Worker.backend = :sequel
@@ -16,17 +16,20 @@ module Bosh::Director
     private
 
     def create_task(username, type, description, deployment, context_id)
-      task = Models::Task.create_with_teams(username:,
-                                            type:,
-                                            description:,
-                                            state: :queued,
-                                            deployment_name: deployment ? deployment.name : nil,
-                                            timestamp: Time.now,
-                                            teams: deployment ? deployment.teams : nil,
-                                            checkpoint_time: Time.now,
-                                            context_id:,
-                                            result_output: '',
-                                            event_output: '')
+      task = Models::Task.create_with_teams(
+        username:,
+        type:,
+        description:,
+        state: :queued,
+        deployment_name: deployment ? deployment.name : nil,
+        timestamp: Time.now,
+        teams: deployment ? deployment.teams : nil,
+        checkpoint_time: Time.now,
+        context_id:,
+        result_output: '',
+        event_output: '',
+      )
+
       log_dir = File.join(Config.base_dir, 'tasks', task.id.to_s)
       FileUtils.rm_rf(log_dir)
       task_status_file = File.join(log_dir, 'debug')
@@ -38,7 +41,14 @@ module Bosh::Director
       end
 
       task.output = log_dir
-      task.save
+      Config.db.transaction(
+        wait: 1.seconds,
+        attempts: 5,
+        retry_on: [Sequel::DatabaseConnectionError, Sequel::DatabaseDisconnectError],
+      ) do
+        task.save
+      end
+
       task
     end
 
--- a/src/bosh-director/lib/bosh/director/job_queue.rb
+++ b/src/bosh-director/lib/bosh/director/job_queue.rb
@@ -1,7 +1,6 @@
 require 'bosh/director/api/task_remover'
 
 module Bosh::Director
-
   # Abstracts the delayed jobs system.
 
   class JobQueue
@@ -17,17 +16,17 @@ module Bosh::Director
     private
 
     def create_task(username, type, description, deployment, context_id)
-      task = Models::Task.create_with_teams(:username => username,
-        :type => type,
-        :description => description,
-        :state => :queued,
-        :deployment_name => deployment ? deployment.name : nil,
-        :timestamp => Time.now,
-        :teams => deployment ? deployment.teams : nil,
-        :checkpoint_time => Time.now,
-        :context_id => context_id,
-        :result_output => "",
-        :event_output => "")
+      task = Models::Task.create_with_teams(username:,
+                                            type:,
+                                            description:,
+                                            state: :queued,
+                                            deployment_name: deployment ? deployment.name : nil,
+                                            timestamp: Time.now,
+                                            teams: deployment ? deployment.teams : nil,
+                                            checkpoint_time: Time.now,
+                                            context_id:,
+                                            result_output: '',
+                                            event_output: '')
       log_dir = File.join(Config.base_dir, 'tasks', task.id.to_s)
       FileUtils.rm_rf(log_dir)
       task_status_file = File.join(log_dir, 'debug')
--- a/src/bosh-director/lib/bosh/director/jobs/db_job.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/db_job.rb
@@ -34,7 +34,6 @@ module Bosh::Director
 
           @job_class.perform(@task_id, @worker_name, *perform_args)
         end
-
         return unless process_status.signaled?
 
         Config.logger.debug("Task #{@task_id} was terminated, marking as failed")
@@ -54,7 +53,8 @@ module Bosh::Director
       private
 
       def update_task_state
-        Config.db.transaction(retry_on: [Sequel::DatabaseConnectionError]) do
+        Config.db.transaction(wait: 1.seconds, attempts: 5,
+                              retry_on: [Sequel::DatabaseDisconnectError, Sequel::DatabaseConnectionError]) do
           task = Models::Task.where(id: @task_id).first
           raise DirectorError, "Task #{@task_id} not found in queue" unless task
 
@@ -66,15 +66,21 @@ module Bosh::Director
           when 'queued'
             task.state = 'processing'
           else
-            task.save
+            raise DirectorError, "Cannot update task state in DB for #{@task_id}" if task.save.nil?
+
             raise DirectorError, "Cannot perform job for task #{@task_id} (not in 'queued' state)"
+
           end
-          task.save
+
+          raise DirectorError, "Cannot update task state in DB for #{@task_id}" if task.save.nil?
         end
       end
 
       def fail_task
-        Models::Task.first(id: @task_id).update(state: 'error')
+        Config.db.transaction(wait: 1.seconds, attempts: 5,
+                              retry_on: [Sequel::DatabaseConnectionError, Sequel::DatabaseDisconnectError]) do
+          Models::Task.first(id: @task_id).update(state: 'error')
+        end
       end
 
       def encode(object)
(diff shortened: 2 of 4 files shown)


klakin-pivotal:
Would it be a big trouble to separate out the parts of the commit that keep RuboCop from muggging us in broad daylight from the parts of the commit that make the change mentioned in the PR? If it wouldn't, would you be so kind to do that?

Maybe it's just me, but I find commits that have a ton of unrelated formatting changes difficult to understand what -exactly- they're trying doing.

nouseforaname:
> Would it be a big trouble to separate out the parts of the commit that keep RuboCop from muggging us in broad daylight from the parts of the commit that make the change mentioned in the PR? If it wouldn't, would you be so kind to do that?
> 
> Maybe it's just me, but I find commits that have a ton of unrelated formatting changes difficult to understand what -exactly- they're trying doing.

I separated them out.

jpalermo:
Retries on database connection errors tend to have race conditions that lead to bad database states.

For example:
Client => Start Transaction
Server => Started
Client => Do Stuff
Server => Recording Stuff
Client => Commit
Server => Done!
<-- Terrible Connection Thing Happens Here and Server Response Never Makes It Back To Client 

Now if the client goes to retry this, because it never heard back about the "Commit", the database is already in the desired state. So you can end up creating a duplicate task, or maybe you do an "upsert" instead, in which case if somebody else already started modifying the task, you might revert it back to the initial state.

You can work around this, but you have to wrap all of your database actions in a way that when retrying you have to validate that none of the failed state actually made it through. This is often a lot of extra work to validate this correctly.

Everything I've said above is nonsense if the DB library we're using is smart enough NOT to retry when if it tried to send the "Commit". If that is true, I'd feel a lot better about this.

nouseforaname:
> Retries on database connection errors tend to have race conditions that lead to bad database states.
> 
> For example: Client => Start Transaction Server => Started Client => Do Stuff Server => Recording Stuff Client => Commit Server => Done! <-- Terrible Connection Thing Happens Here and Server Response Never Makes It Back To Client
> 
> Now if the client goes to retry this, because it never heard back about the "Commit", the database is already in the desired state. So you can end up creating a duplicate task, or maybe you do an "upsert" instead, in which case if somebody else already started modifying the task, you might revert it back to the initial state.
> 
> You can work around this, but you have to wrap all of your database actions in a way that when retrying you have to validate that none of the failed state actually made it through. This is often a lot of extra work to validate this correctly.
> 
> Everything I've said above is nonsense if the DB library we're using is smart enough NOT to retry when if it tried to send the "Commit". If that is true, I'd feel a lot better about this.

Thanks for the explanation. To be honest, I though that making this "safe" was the point of transactions. I tried looking into this and what I found is:

Sequel by default uses autocommit mode for all adapters [see](https://github.com/jeremyevans/sequel/blob/master/doc/transactions.rdoc). (PG)[https://www.postgresql.org/docs/13/ecpg-sql-set-autocommit.html] and (MySQL)[https://dev.mysql.com/doc/refman/5.7/en/innodb-autocommit-commit-rollback.html] both support autocommits. But that doesn't answer your question. I wasn't able to find the information on `what happens if just the commit fails` but everything else succeeds between sequel // the db.

`So you can end up creating a duplicate task`. 

I guess this would only apply here:
https://github.com/cloudfoundry/bosh/pull/2413/files#diff-755d6f30fed127a397969460ea4ad113534dc672578cc569037174fb96813fd5R44

But I'm not sure how retrying saving that task, would lead to a duplicate. Looking at the tasks table and the model it has an ID

https://github.com/cloudfoundry/bosh/pull/2413/files#diff-0628a0099bcf26e0203cbe9d841fc5e72389d8926054b88d96f1112ee111ad17L4

that wouldn't change between retries and I do not see how this could lead to a duplicate tasks as we do have the task id set as the primary key on the table
```
bosh=> SELECT a.attname, format_type(a.atttypid, a.atttypmod) AS data_type
FROM   pg_index i
JOIN   pg_attribute a ON a.attrelid = i.indrelid
                     AND a.attnum = ANY(i.indkey)
WHERE  i.indrelid = 'tasks'::regclass
AND    i.indisprimary;
 attname | data_type
---------+-----------
 id      | integer
(1 row)
```

Statement from (postgres-docs)[https://wiki.postgresql.org/wiki/Retrieve_primary_key_columns]

So I think we should be safe from duplicate tasks. Since they are not created by the workers (but by the single instance director) and even if the commit failed and the whole transaction is retried, we shouldn't be ending up with a dupe as the task_id itself wouldn't change.

The other situation I see in your example is that technically, two workers could pick up the task and make updates to it. Which is true but also the case now. Since this PR makes no changes to the handling of the delayed jobs table that actually tracks the execution of the tasks as a delayed job by the delayed job workers. 

The changes above only attempt to make the writes to the bosh tasks table more resilient (which is not what the delayed job workers use to track their execution state). We call `update_task_state` from within `perform` which is what a delayed job worker will execute:
https://github.com/collectiveidea/delayed_job#custom-jobs

src/bosh-director/lib/bosh/director/job_queue.rb
```diff
@@ -39,7 +42,14 @@ def create_task(username, type, description, deployment, context_id)
       end
 
       task.output = log_dir
-      task.save
+      Config.db.transaction(
+        wait: 1.seconds,
+        attempts: 5,
+        retry_on: [Sequel::DatabaseConnectionError, Sequel::DatabaseDisconnectError],
+      ) do
+        task.save
```
jpalermo, line 50: The transaction block here is not idempotent. When "save" is called on sequel models, the library checks to see if the model is `@new`, and if so, [performs an insert and sets `@new` to `false`](https://github.com/jeremyevans/sequel/blob/master/lib/sequel/model/base.rb#L1787-L1793), if the model is NOT new, it performs an update.

If the transaction here fails after the save has happened, during the transaction commit, the `task` model will already have cleared the `@new` attribute, so trying to call `task.save` a second time during the retry will try to `update` the record in the database, but it doesn't currently exist due to the first transaction being rolled back and will throw an error.

src/bosh-director/lib/bosh/director/jobs/db_job.rb
```diff
@@ -51,26 +53,34 @@ def queue_name
...
+          when 'queued'
             task.state = 'processing'
           else
-            task.save
+            raise DirectorError, "Cannot update task state in DB for #{@task_id}" if task.save.nil?
+
             raise DirectorError, "Cannot perform job for task #{@task_id} (not in 'queued' state)"
+
           end
-          task.save
```
jpalermo, line 68: Looking at the Sequel code, the database error can be thrown after the transaction `commit` has been issued, so the database successfully updates the record, but you still end up in the `retry` state here.

In that situation the `state` for the task after you find it from the database again will be `cancelling` or `processing` which will lead to the `else` case and a very strange error in the logs.

src/bosh-director/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb
```diff
@@ -15,6 +15,27 @@ def initialize(_param = {})
...
+        end
+
+        errored_tasks = []
+        Bosh::Director::Models::Task.select.where('state': %w[queued processing]).each do |task|
+          next if actual_delayed_job_scheduled_task_ids.include?(task.id)
+
+          errored_tasks << task.id
+          task.state = 'error'
+          task.save
+        end
```
jpalermo, line 32: This is tricky, and I'm not sure how to solve it.

You select the delayed jobs first from the database, then select the tasks. If a task doesn't exist in the delayed job records, it gets errored.

But what if a task is created AFTER you pull the delayed jobs records from the database? Then a new task is going to get marked as errored incorrectly.

I'm not sure you can reverse the order of the finds either, you have the same problem where either record could be modified. Doing this work in a transaction and setting the isolation level to `Repeatable read` might work, but it is a bit risky and I've never done this before. 

https://www.postgresql.org/docs/current/transaction-iso.html

You'd have to ensure the isolation level gets reset at the end, not sure if Sequel does that or not, it might keep it for the connection if you don't turn it back down. And then you have the problem of the database raising errors if there is a serialization failure where the Task record is updated in a different transaction at the same time. This might be fine if the error only happens here, we could rescue it and retry the function. But if the error can happen on a normal transaction elsewhere that wouldn't be great.

Maybe there is some other way of doing this where you collect up all the potential errored tasks, then double check at the end that there are no delayed jobs for them, and then when you go to update them to `error`, you'd need to make sure they still exist in the database? Not 100% sure that will catch all the edge cases, but it might...

rkoster:
ping @nouseforaname

src/bosh-director/lib/bosh/director/job_queue.rb
nouseforaname: going over this again, I think I was a bit overzealous adding the retry here as well. Since this are of the code wasn't even involved in the problem that this tries to solve, I just removed this part of the change.

src/bosh-director/lib/bosh/director/jobs/db_job.rb
nouseforaname: I think I addressed that concern.

src/bosh-director/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb
nouseforaname: So I think it should be safe because:

A task gets created by something calling JobQueue.enqueue

JobQueue.enqueue first creates the task by calling 
create_task. Within create_task we call task.save. At this point the task exists in the bosh tasks db, finally create_task returns the task object

then we create the db_job (using, among other things the value of task.id ) 
and use the db_job object and pass it into DelayedJob.enqueue

At this point the delayed job exists in the delayed job db.. 

So we cannot have `but what if a task is created AFTER you pull the delayed jobs records from the database? Then a new task is going to get marked as errored incorrectly.` being the case

nouseforaname: how about using the timestamp? 

If the task got updated / created within the last day, the cleanup is skipped

jpalermo: Probably just a couple of minutes is fine actually. But yeah, that seems like a good solution. `then` isn't normally used in ruby, and there is an extra `)` in there too. I would also expect the ORM to return the `timestamp` as a DateTime object already so the `parse` isn't probably needed.

I was kind of surprised that `-` on a `DateTime` object just flows up to the `Date` class and gives you the number of days difference. If you want smaller offset you can `to_time` the object first in which case `-` will give you a seconds offset.

nouseforaname: Thanks. I updated it to five minutes.. Also updated the test to check for that scenario

src/bosh-director/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb
```diff
@@ -15,6 +17,34 @@ def initialize(_param = {})
...
+          # Newly created tasks are first created in the tasks db, then in the delayed jobs db.
+          # So if a task is being created while we get the list of all jobs in the delayed jobs table,
+          # we may end up missing an item.
+          # To avoid cleaning up tasks where the delayed job item doesn't exist yet, we only clean tasks
+          # if the timestamp older than 5 minutes.
+          next unless (DateTime.now.to_time.to_i - task.timestamp.to_i) >= 300
+
+          errored_tasks << task.id
+          task.state = 'error'
+          task.save
```
beyhan, line 40: I was wondering whether we should do something here in case the `task.save` fails like we do [here](https://github.com/cloudfoundry/bosh/pull/2413/files#diff-0628a0099bcf26e0203cbe9d841fc5e72389d8926054b88d96f1112ee111ad17R71). Failing the operation doesn't make much sense because the task is anyway removed and we will leave also the rest of the tasks orphaned. Maybe we can log this for debug reasons.

beyhan (commented):
The change looks good now I just have a minor comment. Thanks for looking into tihs!

src/bosh-director/lib/bosh/director/jobs/scheduled_tasks_cleanup.rb
nouseforaname: done

jpalermo (approved):
👍

beyhan (approved):

rkoster:
Thanks! @nouseforaname

Issue #1846, opened 2017-11-30
Duplicate `scan and fix` tasks for the same issue 


beyhan:
**Symptoms**
Many `scan and fix` tasks for the same deployment in the BOSH director queue.

**Analysis**
The BOSH director is busy with other tasks and is not able to process the `scan and fix` tasks. In this case the Health Monitor will generate alerts depending on the `analyze_*` interval and this will trigger `scan and fix` tasks generated by the resurrector. The `scan and fix` tasks get faster generated by the Health Monitor than they get executed by the BOSH director.

**Mitigations**
Reconfigure the `analyze_*` interval.

**Identified Issues**
Director is under load and the duplicated `scan and fix` tasks add additional load. 

**Possible Solution**
Director is able to identify duplicate tasks and doesn't add them to the queue. This could be just a solution for `scan and fix` tasks or a general solution for all tasks.

jfmyers9:
I'm going to close this issue and reopen it to generate a tracker story.

This does not appear to be an immediate priority for the BOSH team to fix, but it is something that we would be open to accepting contributions for.

voelzmo:
@jfmyers9 Are you saying the way that @beyhan proposed above would be the one you'd like to see a contribution for?
> Possible Solution: Director is able to identify duplicate tasks and doesn't add them to the queue. This could be just a solution for scan and fix tasks or a general solution for all tasks.

If so: should this be a generic mechanism for all kinds of tasks or just scoped towards `scan_and_fix`?

h4xnoodle:
Ideally any 'problems' for a deployment are collapsed into one scan & fix, given we lock the entire deployment, and we can potentially fix all problems at once in a deployment. 

Would another mitigation here have been to increase the director resources? It's unclear to me whether the director was under an abnormal amount of stress for the amount of deployments compared with VM allocation, or normal load.

beyhan:
May wish was to fix/reduce the duplicate `san and fix` tasks. Nothing general for director. I'm actually OK with the resurrection improvements https://github.com/cloudfoundry/bosh/pull/2144, which will reduce the duplicate `san and fix` tasks a lot. I'm going to close this and will reopened it in case we still observe issue with the new approach.

PR #2414, opened 2022-11-28
do not add more than one scan_and_fix task to the director queue

nouseforaname:
When a director manages a large number of single instance deployments in one AZ

And that AZ goes down

Then queuing additional scan_and_fix tasks for a deployment can create a very large queue for the director containing duplicate scan_and_fix tasks for the same deployments.

If we already have a scan and fix task queued or being processed for a particular deployment, there is no additional benefit in queuing more tasks.

This PR adds a check and some tests around that scenario to avoid overloading a director that is currently dealing with a large number of deployments in a bad state.

[ #183759912 ]

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/plugins/http_request_helper.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/plugins/http_request_helper.rb
@@ -4,7 +4,6 @@ module Bosh::Monitor::Plugins
   module HttpRequestHelper
     def send_http_put_request(uri, request)
       logger.debug("sending HTTP PUT to: #{uri}")
-
       env_proxy = URI.parse(uri.to_s).find_proxy
       request[:proxy] = { host: env_proxy.host, port: env_proxy.port } unless env_proxy.nil?
       name = self.class.name
@@ -36,13 +35,16 @@ module Bosh::Monitor::Plugins
       end
     end
 
-    def send_http_get_request(uri)
+    def send_http_get_request(uri, headers = nil)
       # we are interested in response, so send sync request
       logger.debug("Sending GET request to #{uri}")
       cli = sync_client(OpenSSL::SSL::VERIFY_NONE)
       env_proxy = URI.parse(uri.to_s).find_proxy
       cli.proxy = env_proxy unless env_proxy.nil?
-      cli.get(uri)
+
+      return cli.get(uri) if headers.nil?
+
+      cli.get(uri, nil, headers)
     end
 
     def send_http_post_sync_request(uri, request)
--- a/src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
@@ -105,13 +105,23 @@ module Bosh::Monitor
 
       def scan_and_fix_already_queued_or_processing?(deployment_name)
         url = @uri.dup
+        url.path = '/tasks'
+        if auth_provider(director_info).auth_header.is_a?(Array)
+          auth_header_value = "Basic #{Base64.strict_encode64("#{auth_provider(director_info).auth_header[0]}:#{auth_provider(director_info).auth_header[1]}")}"
+        else
+          auth_header_value = auth_provider(director_info).auth_header
+        end
+        headers = {
+          'Authorization' => auth_header_value,
+          'Content-Type' => 'application/json',
+        }
+        url.query = URI.encode_www_form({ deployment: deployment_name, state: 'queued,processing', verbose: 2 })
+        response = send_http_get_request(url.to_s, headers)
+        return true unless JSON.parse(response.body).select do |item|
+                             item['description'] == 'scan and fix'
+                           end.empty? && response.status_code == 200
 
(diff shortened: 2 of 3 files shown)


beyhan:
FYI and for Context: https://github.com/cloudfoundry/bosh/issues/1846 and https://github.com/cloudfoundry/bosh/pull/2144

src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
```diff
@@ -98,6 +103,17 @@ def process(alert)
 
       private
 
+      def scan_and_fix_already_queued_or_processing?(deployment_name)
+        url = @uri.dup
+
+        url.path = "/tasks"
+        url.query = URI.encode_www_form({deployment: deployment_name, state: "queued,processing", verbose: 2}) 
+        response = send_http_get_request(url.to_s)
```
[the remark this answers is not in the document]
nouseforaname: should be adressed now. Thanks for pointing this out.

src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
```diff
@@ -98,6 +103,27 @@ def process(alert)
 
       private
 
+      def scan_and_fix_already_queued_or_processing?(deployment_name)
+        url = @uri.dup
+        url.path = '/tasks'
+        if auth_provider(director_info).auth_header.is_a?(Array)
```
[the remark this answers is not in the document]
lnguyen: Talked to @nouseforaname I get the reasoning with em:httpclient vs htttpclient

lnguyen (approved):

src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
```diff
@@ -98,6 +103,27 @@ def process(alert)
...
+        end
+        headers = {
+          'Authorization' => auth_header_value,
+          'Content-Type' => 'application/json',
+        }
+        url.query = URI.encode_www_form({ deployment: deployment_name, state: 'queued,processing', verbose: 2 })
+        response = send_http_get_request(url.to_s, headers)
+        return true unless JSON.parse(response.body).select do |item|
+                             item['description'] == 'scan and fix'
+                           end.empty? && response.status_code == 200
```
beyhan, line 122: This statement here is a little bit hard to understand but I'm not sure whether we should return `true` when the `status_code` isn't 200. This is the case at the moment here. Additionally, we don't have a test case which specify what should be the response in case of `status_code` different than 200.

beyhan (changes requested):

src/bosh-monitor/lib/bosh/monitor/plugins/http_request_helper.rb
```diff
@@ -36,13 +35,16 @@ def send_http_post_request(uri, request)
...
       # we are interested in response, so send sync request
       logger.debug("Sending GET request to #{uri}")
       cli = sync_client(OpenSSL::SSL::VERIFY_NONE)
       env_proxy = URI.parse(uri.to_s).find_proxy
       cli.proxy = env_proxy unless env_proxy.nil?
-      cli.get(uri)
+
+      cli.get(uri) if headers.nil?
+
+      cli.get(uri, nil, headers)
```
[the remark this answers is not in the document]
nouseforaname: Thanks! line 45 is supposed to `return`.

src/bosh-monitor/lib/bosh/monitor/plugins/resurrector.rb
nouseforaname: Thanks. Yeah you're right, it is hard to read. 

The idea was to not add an additional task if we currently cannot query the director for tasks. 

The scenario for this PR is one where the director is potentially dealing with 100s of missing vms. So the general load may be high and the tasks endpoint maybe affected. Scheduling additional tasks in that scenario would worsen the situation. The above code would eventually schedule a scan and fix when it was able to check current tasks ( status_code == 200) and it's sure there is no `scan and fix` ( empty? == true ).

I could rewrite it to:

```+        tasks_response = send_http_get_request(url.to_s, headers)
+
+        logger.debug(tasks_response.body)
+        queued_scan_and_fix = JSON.parse(tasks_response.body).select do | task |
+          task['description'] == 'scan and fix'
+        end
+
+        return false if tasks_response.status_code == 200 && queued_scan_and_fix.empty?```

which probably makes it easier to understand but I'm not sure we should schedule a task if the status code isn't 200


in the case where the response isn't 200, the resurrector would skip sending a `scan and fix` for this iteration but still would do it in the next iteration (given that we can properly acccess the tasks endpoint).

beyhan: Ok, this makes sense. It will be great if you can implement the suggestion you provided and also add a comment why we don't want to schedule new tasks when status codes different than 200 or a debug log that we are skipping the scheduling because of that.

nouseforaname: Done.

PR #2415, opened 2022-11-29
Configure Blubraket to ignore non-inclusive language in vendor dirs

rkoster:
LFX is scanning CFF repos for non-inclusive language, however a lot of the findings are for vendored dependencies which we don't control. Scans are performed using Blubraket which support igore files:

https://docs.blubracket.com/how-to/cli/using-cli-with-git/#blubracket-ignore

This PR sets up such a file to try and reduce the noice generated by the LFX scans.

The resuling dashboard can be found here:

https://security.lfx.linuxfoundation.org/#/a092M00001IbM3gQAF/non-inclusive-language


klakin-pivotal:
As an aside, something seems wrong with that Dashboard:

I created a new account by logging in via my `vmware.com` Google account.
Screen Shot 2022-11-29 at 12 09 24

No matter how long I let the page sit, those spinners never go away.

_Should_ I be able to access that dashboard? For reference, the username I selected is `klakinpivotal`.

rkoster:
@klakin-pivotal with regards to access. This is being worked on by the Linux Foundation. I just wanted us to get ahead of the curve and adress some of these false positives before these dashboards are being used more broadly.

rkoster:
@cunnie given that you wholeheartedly support ignoring non-inclusive language in vendor dirs, could you maybe approve this PR so we can go ahead and merge this?

beyhan (approved):

bgandon (approved):
OK for this one-liner 👌

PR #2417, opened 2023-01-05
feat: Short Lived NATS credentials

danielfor:
When bootstraping a new VM, BOSH stores credentials in the metadata services of the different IAASes. This includes the certificates the VM uses to access NATS. This is a security concern for some environments where a malicious user that can access the metadata service gets the NATS credentials. He will be able to connect to NATS and listen to the traffic between the the VMs and the director.

This feature removes that risk by making the credentials stored in the metadata service ephemeral/short lived, so if a user gets access to them, they won't be useful after the VM bootstrap. These credentials will be used only while starting a new VM, and after that, they will be replaced by permanent credentials that are not stored in the metadata service.

This feature is disabled by default. To enable it a new parameter called `enable_short_lived_nats_credentials` should be added with true as its value in the director manifest during the create-env. This will affect only newly created/recreated VMs.

When an operator enables this feature its a MUST to ensure that the stemcells used in the director are compatible with it; stemcells that are incompatible will result in unresponsive VMs.
The oldest compatible stemcell versions are:
Windows 2019 - 2019.41
Ubuntu Xenial - 621.171
Ubuntu Bionic - 1.36
Ubuntu Jammy - All versions are suported.

Story:[ [#183316690] [Deadline: Before End of FYQ1] [AHA] Short-lived NATS bootstrap credential](https://www.pivotaltracker.com/story/show/183316690)

The BOSH unit tests.
Manual integration tests:
- Create environment and deployment with the feature disabled.
- Create environment and deployment with the feature disabled.
- Update an environment with running VMs with the feature disabled and enabled.
- Having a deployment without the feature, recreating it with the feature enabled.

To increase security, this feature will mitigate the risk of plain text NATS credentials, obtained through the metadata service, being used with malicious intent.
The feature will rotate the NATS credential directly after VM creation, and also revoking access of the old credential.

No


src/bosh-director/lib/bosh/director/vm_creator.rb
```diff
@@ -86,6 +86,9 @@ def get_agenda_for_instance_plan(instance_plan, disks, tags, ip_provider, total,
         @dns_encoder,
         @link_provider_intents,
       )
+      if Config.enable_short_lived_nats_credentials
+        agenda.steps << DeploymentPlan::Steps::PermanentNatsCredentialsStep.new
```
rkoster, line 90: This feels counter-intuitive, to add a permanent nats credential only when using short-lived creds.

rkoster (changes requested):
It looks good overall, but I would like to discuss naming.
For example, if instead of creating a tracking the use of a permanent credential we would track the use of a bootstrap nats credential.
This would add an additional step to add a rotate the bootstrap credential and when using short lived creds generate an initial cert with a `bootstrap-` prefix.

jobs/director/spec
```diff
@@ -144,6 +144,9 @@ properties:
   director.enable_nats_delivered_templates:
     description: When true, rendered templates will be sent over NATs
     default: false
+  director.enable_short_lived_nats_credentials:
```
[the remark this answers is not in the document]
danielfor: Yes, I think adding the `bootstrap` word clarifies what is the intention of the flag.

src/bosh-director/lib/bosh/director/vm_creator.rb
rkoster: Yes, reads much more intuitively 👍

src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
```diff
@@ -595,7 +595,7 @@ def create_vms_response(vms_instances_hash)
         results = []
         vms_instances_hash.each_pair do |instance, vms|
           vms.each do |vm|
-            results << create_vm_response(instance, vm).merge('active' => vm.active)
+            results << create_vm_response(instance, vm).merge('active' => vm.active).merge('permanent_nats_credentials' => vm.permanent_nats_credentials)
```
jpalermo, line 598: This will make the `permanent_nats_credentials` attribute come back as part of a `/deployments/NAME/vms` call. If it it was instead added to the `create_vm_response` method directly, rather than being merged in here, it would also come back as part of `/deployments/NAME/instances` api calls.

I'm not really sure it needs to be in both, but it doesn't appear like it would be harmful to include in both.

Did you have a better understanding of this than me and only include it in `vms` for a particular reason?

src/bosh-director/lib/bosh/director/deployment_plan/steps/rotate_nats_bootstrap_credentials_step.rb
```diff
@@ -0,0 +1,21 @@
...
+          agent_permanent_creds = NatsClientCertGenerator.new(@logger).generate_nats_client_certificate "#{report.vm.agent_id}.agent.bosh-internal"
+          settings = {}
+          settings['mbus'] = {
+            'cert' => {
+              'ca' => Config.nats_server_ca,
+              'certificate' => agent_permanent_creds[:cert].to_pem,
+              'private_key' => agent_permanent_creds[:key].to_pem,
+            },
+          }
+          AgentClient.with_agent_id(report.vm.agent_id, report.vm.instance.name).update_settings(settings)
```
jpalermo, line 15: I'm not sure it's safe to perform a partial `update_settings` to the agent here. The agent looks like it expects all the data for `update_settings` and if you don't sent the Trusted Certs, it will blank them out again: https://github.com/cloudfoundry/bosh-agent/blob/6afc0689eeabe7673de54ab0f293df2c746030f1/agent/action/update_settings.go#L63

I think we ran into a similar problem when we were doing the live NATS cert rotation, where `update_settings` was getting called from multiple places, and we had to send all the data every time. At that point we pulled the `update_settings` call into a central place to make it easier to keep them all in sync: https://github.com/cloudfoundry/bosh/blob/3b031a55e4eabbdb6aa7389d9fef444f52aab219/src/bosh-director/lib/bosh/director/deployment_plan/instance.rb#L181

You'll probably need to somehow use that existing tooling (it will likely need to be modified), or you'd have to duplicate the logic and pull the other parts needed into this new Step so it can send the full `update_settings` data too.

jobs/director/spec
```diff
@@ -144,6 +144,9 @@ properties:
   director.enable_nats_delivered_templates:
     description: When true, rendered templates will be sent over NATs
     default: false
+  director.enable_short_lived_nats_bootstrap_credentials:
+    description: When true, NATS bootstrap credentials will be short lived on new VMs
+    default: false
```
jpalermo, line 149: @beyhan & @rkoster, what do you think about flipping this so it's enabled by default?

We've seen very little cost to turning this on, and secure by default seems like the correct security posture.

src/bosh-director/lib/bosh/director/deployment_plan/steps/rotate_nats_bootstrap_credentials_step.rb
jpalermo: In fact, if you could eliminate the new Step entirely and push the logic down into the existing `UpdateInstanceSettingsStep`, it might make things easier and cleaner.

jobs/director/spec
rkoster: How much time does it add to a typical vm_creation (absolute value and percentage of total)?

rkoster: If it adds a few seconds it makes sense to enable it by default, If it's more than say half a minute I'm a bit hesitant.

rkoster: On the other hand, they can always disable it, we just need to make sure to properly communicate about it in the release notes.

beyhan: Having this enabled by default will help us to get faster feedback for the feature and I think our strategy is to provide a secure configuration by default. In case it has significant performance implications we we should at least document this so that the operators can decide.

danielfor: @rkoster a deployment with the feature deactivated for a couple of vms take 1:01 minutes to deploy. If the features is activated then it takes 1:52 minutes.
The VM creation for a single vm goes from:
Task 335 | 21:55:28 | Creating missing vms: jammy-group/e3a8c572-fc51-4055-abce-09828ceecf03 (0) (**00:00:45**)
Task 333 | 21:51:26 | Creating missing vms: jammy-group/efd9ee27-82d6-424a-87c3-a52da210385d (0) (**00:01:35**)
However,  this gets minimized if the deployment creates the VMs in parallel, deploying around 50 VMs takes nearly 03:49 minutes to deploy when the feature is deactivated, and around 3:55 minutes when the feature is activated. 
So the parallelism reduces the total time to deploy many VMs. (At least in GCP, which uses light stemcells.)

src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
danielfor: @jpalermo the rationale to add it in the vms endpoint because the parameter was added to the vms table in the database, and is not part of the instances table. Also in the vms table we have another parameter that is the nats_config_sha so it made sense to leave the nats stuff on the vms call. Also nats-sync uses the vms endpoint.

src/bosh-director/lib/bosh/director/deployment_plan/steps/rotate_nats_bootstrap_credentials_step.rb
danielfor: @jpalermo thanks for noticing the problem, we're trying to replicate and fix it. About moving things to `UpdateInstanceSettingsStep` , it will be difficult to track where the rotation will be triggered, because `instance.update_instance_settings` gets called in other places, so adding more logic there could bring unforeseen behaviors.
This function gets called in the following places according to the IDE:
`recreate_handler.rb:39`
`update_procedure.rb:153`
Do you think it's safe to rotate NATs creds everytime these functions get executed?

jpalermo: `UpdateInstanceSettingsStep` is only used during create, so you can assume you want to regenerate the NATS creds if that step is called (and the feature is enable).

`instance.update_instance_settings`, the underlying function, does get called in a couple of places, you'd probably want to change the method signature to take an additional argument telling it you want to force new NATS creds. Might be other ways to accomplish it too.

Adding more logic and branching to `instance.update_instance_settings` would make it more complicated and add risk. But all that logic would at least be in a single place rather than having multiple spots in the codebase where UpdateSettings could be happening.

src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
jpalermo: If you look at `create_vm_response` about half the response data already comes from the VMs table, so the `/instances` call is already polluted quite a bit. But I can't think of any real reason to add it to both, and the cost of doing that in the future if we do end up needing it there is pretty low, so it's probably fine to leave it just on `vms`

lnguyen:
LGTM

jobs/director/spec
rkoster: From 03:49 to 03:55 on the 50 VM deploy seems acceptable to me. So yeah let's enable it by default.

src/bosh-director/lib/bosh/director/deployment_plan/steps/rotate_nats_bootstrap_credentials_step.rb
danielfor: Hi @jpalermo the changes were completed and now the rotation is happening in `UpdateInstanceSettingsStep`.

jobs/director/spec
danielfor: Cool, the feature is now enabled by default. All the force pushes were made to squash all the commits to tune the code, so please, don't mind them.

jpalermo (approved):
Looks great. Huge feature for such a minimal number of changes.

PR #2418, opened 2023-01-09
Adding Ruby 3.2 support.

lnguyen:
Removed openssl from gemspec as it is build into ruby now

Taint has been removed from ruby 3.2

Updated gems to newer versions to support ruby 3.2


ramonskie (approved):
lgtm

jpalermo (approved):
👍

PR #2419, opened 2023-01-10
Adding support for Postgres 15

lnguyen:
[#182826281] Upgrade Director to use PG15

### DIFF
--- a/jobs/postgres/templates/create-database.erb
+++ b/jobs/postgres/templates/create-database.erb
@@ -1,6 +1,6 @@
 #!/bin/bash -e
 
-PACKAGE_DIR=/var/vcap/packages/postgres-13
+PACKAGE_DIR=/var/vcap/packages/postgres-15
 
 HOST=<%= p("postgres.listen_address") %>
 PORT=<%= p("postgres.port") %>
@@ -13,7 +13,7 @@ export LD_LIBRARY_PATH="$PACKAGE_DIR/lib:$LD_LIBRARY_PATH"
 
 # wait for database to be running
 for i in $( seq 0 30 ); do
-  if /var/vcap/packages/postgres-13/bin/oid2name -H ${HOST} -U ${USER} -d postgres -q ; then
+  if /var/vcap/packages/postgres-15/bin/oid2name -H ${HOST} -U ${USER} -d postgres -q ; then
     break
   fi
 
@@ -34,4 +34,7 @@ for DBNAME in "${DBNAMES[@]}"; do
 
   echo "Trying to alter user (to change password)..."
   $PACKAGE_DIR/bin/psql -d $DBNAME -h $HOST  -p $PORT -U vcap -c "alter role \"$USER\" with password '$PASSWORD'"
+
+  echo "Grant permission to user"
+  $PACKAGE_DIR/bin/psql -d $DBNAME -h $HOST  -p $PORT -U vcap -c  "grant all on schema public to \"$USER\";"
 done
--- a/jobs/postgres/templates/pre-start.erb
+++ b/jobs/postgres/templates/pre-start.erb
@@ -2,14 +2,14 @@
 
 set -eu
 
-PACKAGE_DIR=/var/vcap/packages/postgres-13
-PACKAGE_DIR_OLD=/var/vcap/packages/postgres-10
+PACKAGE_DIR=/var/vcap/packages/postgres-15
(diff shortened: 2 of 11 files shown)


jobs/postgres/templates/pre-start.erb
```diff
@@ -2,14 +2,14 @@
...
-PACKAGE_DIR=/var/vcap/packages/postgres-13
-PACKAGE_DIR_OLD=/var/vcap/packages/postgres-10
+PACKAGE_DIR=/var/vcap/packages/postgres-15
+PACKAGE_DIR_OLD=/var/vcap/packages/postgres-13
 
 PERSISTENT_DISK_DIR=/var/vcap/store
 
-STORE_DIR=${PERSISTENT_DISK_DIR}/postgres-13
-STORE_DIR_OLD=${PERSISTENT_DISK_DIR}/postgres-10
-STORE_DIR_OBSOLETE=${PERSISTENT_DISK_DIR}/postgres-9.4
```
jpalermo, line 12: We are fully dropping detection of postgres-9.4 here.

The last version that shipped postgres 9 was 271.20, which was cut February of 2022. That version also had postgres 10, but the postgres 9 job was still there.

If somebody does upgrade from 271.20 or before with a postgres 9, it won't delete their data, but it will deploy successfully and come up with a totally empty director, and then probably take a bit of manual work to get themselves back into a happy state.

Not saying we must include a check for postgres 9 still, mostly just looking to get everybody's opinion on it.

beyhan: What about failing in the case of postgres 9 with an error message which suggest an update to an older version first and then to the ones which support postgres 15?

lnguyen: @beyhan @jpalermo ive added changes to detect both 9.4 and 10 as obsolete versions and fail out

packages/postgres-15/README.md
```diff
@@ -0,0 +1,5 @@
+postgres-13 package
+============
+Postgres! 
+
+The blob contains source code, we probably got it [here?](https://www.postgresql.org/ftp/source/v13.2/)
```
beyhan, line 5: This still states postgres-13 and most probably should be postgres-15 and the link also.

jobs/postgres/templates/create-database.erb
```diff
@@ -34,4 +34,7 @@ for DBNAME in "${DBNAMES[@]}"; do
 
   echo "Trying to alter user (to change password)..."
   $PACKAGE_DIR/bin/psql -d $DBNAME -h $HOST  -p $PORT -U vcap -c "alter role \"$USER\" with password '$PASSWORD'"
+
+  echo "Grant permission to user"
+  $PACKAGE_DIR/bin/psql -d $DBNAME -h $HOST  -p $PORT -U vcap -c  "grant all on schema public to \"$USER\";"
```
beyhan, line 39: I was wandering why we need this now?

beyhan (changes requested):
In general it looks good. I have just a minor change request and a question.

jobs/postgres/templates/create-database.erb
lnguyen: I don't know if some permissions changed on postgres 13 -> 15 but the user wasnt able to migrate db for BOSH on new instances.

jobs/postgres/templates/pre-start.erb
```diff
@@ -2,30 +2,32 @@
...
+  if [ -d ${PERSISTENT_DISK_DIR}/postgres-${version} ]; then
+    # uh-oh, we have years-old BOSH Director
+    if [ ! -d $STORE_DIR_OLD ] && [ ! -d $STORE_DIR ]; then
+      # we never upgraded this BOSH Director from PostgreSQL 10 or below to 15
+      echo "Please use a previous bosh release version (277.x or lower) to migrate data from postgres-${version} to postgres-13."
+      exit 1
+    fi
+    # delete the obsolete directory to free up space
+    echo "Deleting obsolete $STORE_DIR_OBSOLETE directory."
+    rm -rf "${STORE_DIR_OBSOLETE}"
```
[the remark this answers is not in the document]
lnguyen: Fixed

jpalermo (approved):
👍

rkoster (approved):

beyhan (approved):

Issue #2424, opened 2023-02-02
dd-agent failure 277.0 on '=~'


blyles:
**Describe the bug**
We hit a problem where our deployment started presenting with stack traces like the below..it has 2 separate examples.

```
Task 913881

Task 913881 | 13:35:18 | Preparing deployment: Preparing deployment (00:00:01)
Task 913881 | 13:35:19 | Preparing deployment: Rendering templates (00:00:00)
                       L Error: Unable to render instance groups for deployment. Errors are:
  - Unable to render jobs for instance group 'hazelcast'. Errors are:
    Error filling in template 'monit' (line 7: undefined method `=~' for false:FalseClass)
Task 913881 | 13:35:19 | Error: Unable to render instance groups for deployment. Errors are:
  - Unable to render jobs for instance group 'hazelcast'. Errors are:
    Error filling in template 'monit' (line 7: undefined method `=~' for false:FalseClass)
```

It's similar in all our zookeeper, storm and hazelcast deployments and in multiple environments.

A further log we pulled from debug shows - 

```
D, [2023-01-31T13:29:10.386790 #970603] [task:1570265] DEBUG -- DirectorJobRunner: Error filling in template 'monit' (line 15: undefined method `=~' for false:FalseClass)
dd-agent/monit:15:in `get_binding'
/var/vcap/data/packages/director-ruby-3.2/7dc1b64d4a62f07f0ed58f53b5aa0805a3e6cbac/lib/ruby/3.2.0/erb.rb:429:in `eval'
/var/vcap/data/packages/director-ruby-3.2/7dc1b64d4a62f07f0ed58f53b5aa0805a3e6cbac/lib/ruby/3.2.0/erb.rb:429:in `result'
/var/vcap/data/packages/director/fb2ee62009b2434c44ecd9925493389491fecd33/gem_home/ruby/3.2.0/gems/bosh-director-core-0.0.0/lib/bosh/director/core/templates/source_erb.rb:20:in `block in render'
/var/vcap/data/packages/director/fb2ee62009b2434c44ecd9925493389491fecd33/gem_home/ruby/3.2.0/gems/bosh-director-core-0.0.0/lib/bosh/director/core/templates/source_erb.rb:19:in `synchronize'
[...]

PR #2425, opened 2023-02-02
Add option to enable pre ruby 3.2 =~ beahvior

jpalermo:
Addresses #2424

This is disabled by default but is needed if a release template is incorrectly using `=~`
`=~` was removed from Kernel in ruby 3.2 and those templates will now break without this.

Ruby change context: https://bugs.ruby-lang.org/issues/15231

### DIFF
--- a/jobs/director/spec
+++ b/jobs/director/spec
@@ -144,6 +144,9 @@ properties:
   director.enable_nats_delivered_templates:
     description: When true, rendered templates will be sent over NATs
     default: false
+  director.enable_pre_ruby_3_2_equal_tilde_behavior:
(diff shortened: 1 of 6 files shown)


beyhan (commented):
In general the change looks good. I'm just wandering whether we should continue allowing incorrect use of `=~` by patching the Ruby Kernel or rather force the bosh releases to fix their usage. WDYT?

rkoster:
Forcing release authors to update is the right thing to do. However sometimes it is not possible for operators to make this happen. We don't want to block operators from upgrading their director version which is why it is important to have this patch available.

beyhan (approved):

beyhan:
After discussing this in the FI WG meeting let's go this way.

lnguyen (approved):
LGTM

PR #2427, opened 2023-02-15
add pg-13 job

nouseforaname:
some scenarios may require users to keep using pg-13 for various reasons (compliance, testing, etc). These users are currently locked out from receiving release updates since they have to update to pg15 to consume newer release versions. This adds a pg13 job based on what we've had used previously.

[ #184442896 ]

### DIFF
--- a/jobs/postgres-13/templates/pre-start.erb
+++ b/jobs/postgres-13/templates/pre-start.erb
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+set -eu
+
+PACKAGE_DIR=/var/vcap/packages/postgres-13
+PACKAGE_DIR_OLD=/var/vcap/packages/postgres-10
+
+PERSISTENT_DISK_DIR=/var/vcap/store
+
+STORE_DIR=${PERSISTENT_DISK_DIR}/postgres-13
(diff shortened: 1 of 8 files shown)


ramonskie (approved):
looks good to me.

jobs/postgres-13/templates/pre-start.erb
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+set -eu
+
+PACKAGE_DIR=/var/vcap/packages/postgres-13
+PACKAGE_DIR_OLD=/var/vcap/packages/postgres-10
+
+PERSISTENT_DISK_DIR=/var/vcap/store
+
+STORE_DIR=${PERSISTENT_DISK_DIR}/postgres-13
```
ystros, line 10: I confirmed that when the `postgres` job was configured to use Postgres 13, it also used the same `${PERSISTENT_DISK_DIR}/postgres-13` store directory. This should ensure that folks can seamlessly upgrade from that Postgres 13 version of the `postgres` job to this new `postgres-13` specific job without any data migration.

See: https://github.com/cloudfoundry/bosh/blob/2fb569760b5ed669f225e15ce6586a3c4312d9c2/jobs/postgres/templates/pre-start.erb

lnguyen (approved):
LGTM

PR #2428, opened 2023-02-24
Do not attempt to "fix" packages in other releases

ystros:
When to different releases contain a package with the same fingerprint but different shasums, running `bosh upload-release --fix` on one of those releases will update the blobs for the original release. However, it does not update the shasum for that package to reflect the new contents of the blob. This will result in compilation failures like:

> Checking downloaded blob ''0a282ed0-9c46-4a13-bbbe-6ed1578480d5'': Expected stream to have digest ''sha256:ff5a3c4c82953ddc10ee19fa776ff4603216b58d673b3fd90fa0ee80cf750340'' but was ''sha256:279bfb24ccede8477400bb76d636a41da973d9b33fa811898b3d43bb4d3bb4a5''

The intention of `--fix` is to re-upload a release that has become corrupted in some way. As such, we have changed the behavior to not attempt to re-use packages from unrelated releases when `--fix` is present.

[#184562287] `bosh upload-release --fix` should not tamper with existing packages in other releases

Signed-off-by: Joseph Palermo <jpalermo@vmware.com>

### DIFF
--- a/src/bosh-director/lib/bosh/director/jobs/update_release.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/update_release.rb
@@ -214,6 +214,7 @@ module Bosh::Director
           @version,
           manifest_packages,
           logger,
+          @fix,
         )
 
         PackagePersister.persist(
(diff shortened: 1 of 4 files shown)


ramonskie (approved):
lgtm

lnguyen (approved):
lgtm

PR #2430, opened 2023-03-02
Rebuilds Postgres indexes on glibc version change

klakin-pivotal:

This commit:
* Adds early fails to the Postgres 10 and Postgres 13 jobs if the Postgres 15 data directory is present. This omission seemed to be a clear oversight.
* Adds code to detect glibc version changes and rebuild indexes on all Postgres databases if the running glibc is newer than 2.27, and has changed since the last time we rebuild the indexes.

glibc version 2.28 introduced a major change to how it performs collation. Whenever any changes to collations are introduced, indexes on Postgres tables must be rebuilt in order to avoid problems which include successful insertion of rows that violate uniqueness constraints on some keys but not others.

Postgres 15 and later includes code that detects and warns about this situation, but leaves correcting the issue up to the operator... so the updates in this commit are still required.

 See: <https://wiki.postgresql.org/wiki/Locale_data_changes>
      <https://www.crunchydata.com/blog/glibc-collations-and-data-corruption>

Manual testing for script correctness.

No.

### DIFF
--- a/jobs/postgres-10/templates/pre-start.erb
+++ b/jobs/postgres-10/templates/pre-start.erb
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -eu
+set -o pipefail
 
 PACKAGE_DIR=/var/vcap/packages/postgres-10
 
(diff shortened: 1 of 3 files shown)


lnguyen (approved):
lgtm

jpalermo (approved):
👍

PR #2438, opened 2023-04-27
Properly handle package fingerprint collisions

ystros:
When a release is uploaded, the BOSH Director attempts to match those packages in the release to "similar" packages that have been previously uploaded. It does this currently by solely looking at the package fingerprint.

Recently, the golang-release was updated to template its various packages (e.g. golang-1-linux, golang-1.20-linux) and also support prefixes when vendoring its packages into another release. The packaging scripts then replace a string literal with the actual version/prefix in the final compiled package. Because of this templating, the fingerprint for each package is identical, but the compiled bits are different.

Since the Director matches only on fingerprint, this has led to situations where two releases vendor different packages from the golang-release, those packages have matching fingerprints, and thus when deploying VMs, one will receive its expected package bits while the other will receive the other release's version of the package.

For example, if the golang-1.20-linux package is compiled first, then a release that uses the golang-1-linux package will receive the compiled bits for the golang-1.20-linux package instead. This can be identified by `/var/vcap/packages/golang-1-linux/bosh/compile.env` referencing golang-1.20-linux instead of golang-1-linux:
```
  export GOROOT=$(readlink -nf /var/vcap/packages/golang-1.20-linux)
```

Theoretically, this could also happen to any unrelated packages that happened to have the same fingerprint. It would also occur if two releases were vendoring the same package but with different prefixes.

To solve this, "similar" packages are now discovered using the combination of the package name and the package fingerprint.

[#185044288] BOSH Director should only re-use compiled packages if both fingerprint and package name match

### DIFF
--- a/src/bosh-director/spec/unit/jobs/update_release/package_persister_spec.rb
+++ b/src/bosh-director/spec/unit/jobs/update_release/package_persister_spec.rb
@@ -128,6 +128,111 @@ module Bosh::Director
             expect(compiled_packages.length).to eq(2)
           end
 
+          context 'when there is a collision between package fingerprints' do
+            let(:manifest_compiled_packages) do
+              [
+                {
+                  'sha1' => 'fakesha2',
+                  'fingerprint' => 'same-fingerprint',
+                  'name' => package_name_2,
+                  'version' => 'fake-version-2',
+                  'compiled_package_sha1' => 'compiled-sha1-2',
+                  'stemcell' => 'foo/bar',
+                  'dependencies' => [],
+                },
+              ]
+            end
+
+            before do
+              other_release = Models::Release.make(name: 'other-release')
+              existing_package = Models::Package.make(
+                release: other_release,
+                sha1: 'sha1-1',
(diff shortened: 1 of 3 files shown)


src/bosh-director/spec/unit/jobs/update_release/package_persister_spec.rb
```diff
@@ -128,6 +128,114 @@ def persist_packages(manifest_packages, compiled_flag)
...
+                  'version' => 'fake-version-2',
+                  'compiled_package_sha1' => 'compiled-sha1-2',
+                  'stemcell' => 'foo/bar',
+                  'dependencies' => [],
+                },
+              ]
+            end
+
+            before do
+              # Models::ReleaseVersion.make(version: )
```
selzoc, line 147: Extraneous comment?

ystros: Thanks! Removed + force pushed.

lnguyen (approved):

PR #2439, opened 2023-04-28
Enable coverage report generation when running unit tests on ci

rkoster:

This PR restores simplecov code coverage reporting when running unit tests in ci.
In the past, this must have worked because the required configuration is [already in place](https://github.com/cloudfoundry/bosh/blob/main/src/spec/shared/spec_helper.rb#L3-L18), however, the simpelcov gem itself was no longer included in the Gemfile.

At VMware we have an internal ask to report test coverage for components we ship, also from an OSS point of view knowing our test coverage is a best practise. 

Have tested locally using:
```
docker run \
       --mount type=bind,source="$(pwd)",target=/workspace/bosh-src \
       --workdir /workspace \
       --env DB=sqlite \
       --env RUBY_VERSION=3.2.0 \
       --env COVERAGE=true \
       bosh/integration bosh-src/ci/tasks/test-unit.sh
```
Which matches the [test-unit.yml ](https://github.com/cloudfoundry/bosh/blob/main/ci/tasks/test-unit.yml)task invocation

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/src/vendor/cache/** filter=lfs diff=lfs merge=lfs -text
--- a/ci/tasks/test-unit.yml
+++ b/ci/tasks/test-unit.yml
@@ -16,3 +16,4 @@ params:
   RUBY_VERSION: 3.2.0
   DB:           replace-me
   DB_VERSION:   replace-me
+  COVERAGE:     true
--- a/src/Gemfile
+++ b/src/Gemfile
@@ -48,6 +48,8 @@ group :development, :test do
   gem 'sinatra-contrib', '>= 2.2.0'
   gem 'webmock'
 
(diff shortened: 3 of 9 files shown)


aramprice (approved):
✅ lgtm

aramprice:
🤔 is anyone consuming this coverage output? 

If not is it worth adding this code, configuration, and run-time back in?

ragaskar:
Didn't clone + test but LGTM.

ragaskar (approved):

jpalermo:
Adding some new gems, maybe now is a good time to enable git lfs for the vendor folder

rkoster:
> Adding some new gems, maybe now is a good time to enable git lfs for the vendor folder

Done: https://github.com/cloudfoundry/bosh/pull/2439/commits/dac1f51d92d0015bc6e23c6c06c50b251e09513c

PR #2442, opened 2023-05-09
Maintenance/update java build integration image

nouseforaname:
Update the JDK in use to a newer build. The one currently in use is several years old


nouseforaname:
cryo work, there's a task that needs the ecosystem tooling to run (among other things) credhub..

ramonskie (approved):
Lgtm

PR #2444, opened 2023-05-16
support instance tags and include relevant specs

anshrupani:

The change allows to configure instance group specific tags, by allowing the user to include tags under <instance_group> -> env -> bosh -> tags in the manifest.

These tags can be configured as follows:
```
instance_groups:
- name: <instance_group>
  env:
    bosh:
     .
     .
      tags:
        tag1: value1
        tag2: value2
```
The deployment tags and the tags from the runtime config are also merged.

These tags can be useful under many circumstances, for example, in a large organisation, it makes identifying teams responsible for different instances easier. Also, This feature removes the necessity of including tags for the entire deployment when such tags aren't needed for all the involved instance groups. 

bosh-director unit tests, dev-release testing on development environments

Support instance group specific tags.

No breaking change, since this PR just supports additional functionality of supporting instance tags.

@mvach @a-hassanin

### DIFF
--- a/src/bosh-director/lib/bosh/director/vm_creator.rb
+++ b/src/bosh-director/lib/bosh/director/vm_creator.rb
@@ -54,6 +54,12 @@ module Bosh::Director
       instance = instance_plan.instance
       already_had_active_vm = instance.vm_created?
 
+      variables_interpolator = Bosh::Director::ConfigServer::VariablesInterpolator.new
+      env = variables_interpolator.interpolate_with_versioning(instance.env, instance.desired_variable_set)
+      env['bosh'] ||= {}
+      env['bosh'] = Config.agent_env.merge(env['bosh'])
+      tags = combined_tags(tags, env['bosh']['tags'])
+
       agenda.steps = [
         DeploymentPlan::Steps::CreateVmStep.new(
           instance_plan,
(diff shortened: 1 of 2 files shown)


beyhan:
@ansh-SAP could you please sign the CLI and also add information about why this change is useful.

anshrupani:
> @anshrupani one of the commits is with @ansh-SAP which didn't sign the CLA. Could you fix this and also add information about why this change is useful.

Thanks for pointing out!

ramonskie:
could you squash these commits?

rkoster (changes requested):
The feature makes sense but why was added under `bosh.env` instead of added as a instance-level key:
```yaml
instance_groups:
- name: foo
  tags:
    bar: foo
```

Furthermore I think it would be better to add this logic somewhere in the deployment_plan, for example in: [src/bosh-director/lib/bosh/director/deployment_plan/instance_plan_factory.rb](https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/lib/bosh/director/deployment_plan/instance_plan_factory.rb) (but maybe you find a better place, but this looked fairly close).

Last but not least I would expect some tests which exercise the new code paths.

rkoster:
Closing in favour of https://github.com/cloudfoundry/bosh/pull/2448

PR #2446, opened 2023-05-25
Fix for "Extracting compiled package archive failed" bug when uploading release with compiled packages

jpalermo:
This situation exists if the bosh director already has a compiled package with a fingerprint and the cli is trying to upload a release that has two different copies of that same package with the same fingerprint using two different names. The current package matching logic only uses fingerprints to match, so the cli ends up removing both packages from the release before uploading it, but the director doesn't actually have compiled versions of both packages.

So instead, we now do not send the fingerprint in the matching response if there is a package in the release manifest that includes that fingerprint but was not successfully matched.

Reproduction steps:
Two compiled releases that share a common package name and fingerprint.
The second release contains a second copy of the package with a different name
Upload the first release
Upload the second release (with two packages)

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
@@ -67,7 +67,12 @@ module Bosh::Director
             filtered_packages << package
           end
         end
-        filtered_packages
+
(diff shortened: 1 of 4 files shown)


lnguyen (approved):
LGTM

PR #2448, opened 2023-05-31
Support instance group tags

a-hassanin:

The change allows to configure instance group specific tags, by allowing the user to include tags under <instance_group> -> tags in the manifest.

These tags can be configured as follows:

```
instance_groups:
- name: <instance_group>
  tags:
    tag1: value1
    tag2: value2
```

These tags are merged with the instance plan tags during VM creation and have higher priority when there are keys which already exist. These tags can be useful under many circumstances, for example, in a large organisation, it makes identifying teams responsible for different instances easier. Also, This feature removes the necessity of including tags for the entire deployment when such tags aren't needed for all the involved instance groups.

This PR builds on the feedback given in: https://github.com/cloudfoundry/bosh/pull/2444 

bosh-director unit tests, dev-release testing on development environments

Support instance group specific tags.

No breaking change, since this PR just supports additional functionality of supporting instance tags.

@anshrupani @mvach

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/compilation_instance_pool.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/compilation_instance_pool.rb
@@ -277,7 +277,7 @@ module Bosh::Director
 
     class CompilationInstanceGroup
       attr_reader :vm_type, :vm_resources, :vm_extensions, :stemcell, :env, :name
-      attr_reader :instance_plans
+      attr_reader :instance_plans, :tags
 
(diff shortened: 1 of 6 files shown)


a-hassanin:
@rkoster Thursday is a public holiday in Germany, can you please have a look at this enhanced PR and share your feedback?
Thanks,

Cheers, 
Ahmed

rkoster (approved):

rkoster:
@a-hassanin would you be so kind as to also create a PR to update: https://github.com/cloudfoundry/docs-bosh/edit/master/content/manifest-v2.md

beyhan (approved):

PR #2450, opened 2023-06-21
Fixes Postgres 10 -> 15 upgrade with 9.4 db dir

klakin-pivotal:
This commit
* Fixes the upgrade case where a customer upgraded from Postgres 9.4 to Postgres 10, and then chose to upgrade to Postgres 15. The previous logic expected that when a PG 9.4 directory was present, then a PG 10 AND a PG 13 directory also HAD to be present, or else it would fail the pre-start script early. This commit corrects that logic.
* Also refactors the logic that checks for older (and unsupported) PG versions, and also the logic that upgrades from supported PG versions to the latest version. Should we ever find ourselves needing to obsolete a version of Postgres, or add a new one, hopefully this rewrite will make modifying this script easier.
* Fixes a bug introduced in 94522f80ef3bccd5dc72f9e81060d628dd700721

[#185335638]

NOTE: The refactored logic has been "mocked" out and tested, so we're confident in its correctness.
@ystros

### DIFF
--- a/jobs/postgres/templates/pre-start.erb
+++ b/jobs/postgres/templates/pre-start.erb
@@ -3,36 +3,63 @@
 set -eu
 set -o pipefail
 
-PACKAGE_DIR=/var/vcap/packages/postgres-15
-PACKAGE_DIR_OLD=/var/vcap/packages/postgres-13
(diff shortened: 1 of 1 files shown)


lnguyen (approved):
LGTM

PR #2452, opened 2023-07-05
Allow overriding of `max_in_flight` values when resolving deployment problems

selzoc:

This will allow the CLI (or other clients) to override the `max_in_flight` values per instance group when resolving deployment problems.  Previously, the `max_in_flight` values in the manifest were used, and the only way to change them was to re-deploy.  However, in disaster recovery scenarios, often a re-deploy is quite difficult, but you want to resolve problems as fast as possible.

In support of the bosh recover work ([story](https://www.pivotaltracker.com/story/show/185483613)), which asks users for desired problem resolutions on a per-instance-group basis, and asks if they'd like to override the `max_in_flight` values.

[bosh create-recovery-plan PR](https://github.com/cloudfoundry/bosh-cli/pull/622)

Units

Resolve problem endpoint now allows `max_in_flight_overrides`.

No

### DIFF
--- a/src/bosh-director/lib/bosh/director.rb
+++ b/src/bosh-director/lib/bosh/director.rb
@@ -65,6 +65,7 @@ require 'bosh/director/download_helper'
 require 'bosh/director/formatter_helper'
 require 'bosh/director/tagged_logger'
 require 'bosh/director/duplicate_detector'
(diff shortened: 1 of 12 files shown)


cunnie (approved):
Looks good to me!

PR #2454, opened 2023-07-11
Attempt to fix "Extracting compiled package archive failed" error

ystros:
Previously, there were cases where a package existed the name and fingerprint, but did not match stemcell, was mistakenly excluded from upload. This results in errors like the following:

```
Task 1939132 | 20:03:34 | Creating new compiled packages: routing-golang-1-linux/8c04109541f4d504f5be559da433998bd459b0f45cd3654557cc3642cc4d2f60 for ubuntu-jammy/1.147 (00:00:00)
                        L Error: Extracting compiled package archive failed. Check task debug log for details.
```

Now, when deciding which packages do not need to be uploaded, we consider name, fingerprint, dependency key, and stemcell os/version.

We ended up refactoring quite a bit in an effort to make the code more understandable (and understand it ourselves...). The most relevant logic change is in the `compiled_package_fingerprints_not_to_be_uploaded` method.

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
@@ -3,50 +3,67 @@ require 'bosh/director/api/controllers/base_controller'
 module Bosh::Director
   module Api::Controllers
     class PackagesController < BaseController
-      post '/matches', :consumes => :yaml do
-        manifest = YAML.load(request.body.read, aliases: true)
+      post '/matches', consumes: :yaml do
+        manifest_hash = YAML.load(request.body.read, aliases: true)
 
-        unless manifest.is_a?(Hash) && manifest['packages'].is_a?(Array)
+        unless manifest_hash.is_a?(Hash) && manifest_hash['packages'].is_a?(Array)
           raise BadManifest, "Manifest doesn't have a usable packages section"
         end
 
-        fingerprint_list = []
+        fingerprints =
+          if existing_release_version_dirty?(manifest_hash)
+            []
+          else
+            fingerprint_list = manifest_hash['packages'].map { |package| package['fingerprint'] }.compact
+            Models::Package.where(fingerprint: fingerprint_list)
+              .where(Sequel.~(sha1: nil))
+              .where(Sequel.~(blobstore_id: nil)).all
+              .map(&:fingerprint).compact.uniq
+          end
 
-        manifest['packages'].each do |package|
-          fingerprint_list << package['fingerprint'] if package['fingerprint']
-        end
-
-        matching_packages = []
-
-        unless existing_release_version_dirty?(manifest)
-          matching_packages = Models::Package.where(fingerprint: fingerprint_list)
-                                             .where(Sequel.~(sha1: nil))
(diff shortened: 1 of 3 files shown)


jpalermo (approved):
👍

src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
```diff
@@ -3,81 +3,101 @@
...
         end
 
-        fingerprint_list = []
+        fingerprints =
+          if existing_release_version_dirty?(manifest_hash)
+            []
+          else
+            fingerprint_list = manifest_hash['packages'].map { |package| package['fingerprint'] }.compact
+            Models::Package.where(fingerprint: fingerprint_list)
+              .where(Sequel.~(sha1: nil))
```
ramonskie, line 19: could these result in issues where ppl are using the sha256 hack for releases.
where they provide `sha1: sha256:HASH256` ?

aramprice: Possibly - this code no [moved, and not new](https://github.com/cloudfoundry/bosh/pull/2454/files#diff-cec27a0c53f8e3d25c796492e829047ac038ce2f52a56b126d081e64146679f3L23) so if this is an issue we should probably address it in a separate change.

ystros: The column in the database is always called `sha1`, regardless of whether an actual sha1 or a sha256 is provided. `Sequel.~(sha1: nil)` should translate to SQL `sha1 IS NOT NULL`, so it should cover both sha1 and sha256 cases.

Regardless, this query is not something we changed, simply reordered things to make the structure of both the `/matches` and `/matches_compiled` controller actions similar. The diff is messy, but you can see the query in the previously existing code: https://github.com/cloudfoundry/bosh/blob/5a882fddf3cf0e04ee7be0e74d1fb503b4c59b16/src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb#L22-L24

src/bosh-director/lib/bosh/director/api/controllers/packages_controller.rb
```diff
@@ -3,81 +3,101 @@
...
-                                             .where(fingerprint: fingerprint_list).all
+        package_hashes_that_are_already_uploaded = existing_package_hashes & compiled_package_hashes
 
-          matching_packages = filter_matching_packages(matching_packages, manifest)
-        end
+        fingerprints_that_need_upload = package_hashes_that_need_upload.map { |h| h[:fingerprint] }.compact.uniq
+        fingerprints_that_are_already_uploaded = package_hashes_that_are_already_uploaded.map { |h| h[:fingerprint] }.compact.uniq
 
-        json_encode(matching_packages.map(&:fingerprint).compact.uniq)
+        fingerprints_that_are_already_uploaded - fingerprints_that_need_upload
```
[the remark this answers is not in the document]
nouseforaname: Nevermind, just realized that the method is called 
`compiled_package_fingerprints_not_to_be_uploaded`

nouseforaname (approved):
lgtm

PR #2460, opened 2023-08-09
Introduce on-stemcell-change variable strategy

ystros:
This commit introduces a new `update` block under variable declarations, as well as a `strategy` key underneath it to control when BOSH retrieves the latest version of that variable. This feature allows better control over when new variable versions are rolled out to VMs during a deploy.

Currently, when a manifest is deployed, the BOSH Director will query CredHub for the latest version of each variable in the manifest. This ensures that deployments always use the current configuration stored in CredHub.

This is a problem for being able to automatically rotate certificates - if a new version of a certificate is in CredHub, the deploy will use it. Normally this would be desirable, but since the rotation is done without users taking a direct action, they may not be aware that there are updated certs and thus will be surprised by a longer-than-normal deploy if they have no other changes. Predicting when there's going to be a longer-than-normal deploy is important for strict change / maintenance windows.

The new key on variables has two potential values:
* on-deploy: BOSH will use the current behavior of looking up the latest variable value on each deploy (default)
* on-stemcell-change: BOSH will only look up the latest variable value when a deploy is occuring that updates all stemcells for the deployment. If one or more stemcells is not being updated, then BOSH will continue to use the previously deployed value for the variable.

The on-stemcell-change behavior can be overridden by passing the `force_latest_variables` parameter to the deploy API endpoint. This will ensure that the latest variable value is used, regardless of the variable's declared strategy.

Corresponding docs PR: https://github.com/cloudfoundry/docs-bosh/pull/799

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/deployments_controller.rb
@@ -416,6 +416,7 @@ module Bosh::Director
         options['canaries'] = params[:canaries] if params['canaries']
         options['max_in_flight'] = params[:max_in_flight] if params['max_in_flight']
         options['scopes'] = token_scopes
+        options['force_latest_variables'] = true if params['force_latest_variables'] == 'true'
 
         # since authorizer does not look at manifest payload for deployment name
         @deployment = Models::Deployment[name: deployment_name]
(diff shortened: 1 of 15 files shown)


selzoc (approved):

lnguyen (approved):
LGTM

PR #2474, opened 2023-10-30
Re-allow job spec manifests to use yaml anchors

jpalermo:
This was previously allowed, but the functionality was removed as part of a previous upgrade of Ruby.

Report of the problem in [CFF Slack](https://cloudfoundry.slack.com/archives/C02FL4A2K/p1698317333685859)


nouseforaname:
Lgtm. Thanks!

nouseforaname (approved):

PR #2477, opened 2023-11-06
[WIP] refactor pipeline so it can be used in the bosh foundation ci

ramonskie:
we want to move the pipeline over to bosh.ci.cloudfoundry.org

and for this we need to run the acceptance tests on the gcp/aws foundation accounts.


ramonskie:
this can be closed as it superseded by #2483

PR #2478, opened 2023-11-27
remove unused ruby 3.1 package

ramonskie:
as we migrated to ruby 3.2 a while ago i think its time to remove the unused ruby 3.1 package


jpalermo (approved):
👍

PR #2479, opened 2023-11-28
ERB template rendering code review: rename old bosh v1 names to v2

bgandon:

Here, I’ve reviewed the ERB template rendering code in order to document it in the cloudfoundry/docs-bosh#813 PR.

In that part of the Director’s code, I’ve noticed _many_ misleading use of the term “template”, used instead of “job”. And this makes the code very confusing because “template” could also refer to “ERB template”.

So, I’ve started chasing these names that had not been migrated back in 2017 (when Bosh v1→v2 refactoring effort had been conducted), and I’ve looked for more suitable names. Don’t hesitate to challenge these, by the way.

Then I’ve updated the unit tests. Here also I’ve tried to clarify some ambiguous namings.

In that process, I’ve noticed that when rendering ERB templates, the job name was read from `job.MF`, freshly extracted from the job blob, whereas the name already persisted in database. I’ve removed the unnecessary unmarshalling of `job.MF` and added a name verification at release-upload time.

The modifications in this PR are quite large overall, but I’ve paid great attention in separating this into smaller chunks of work in to the individual commits.

This work was done in order to gather information for authoring cloudfoundry/docs-bosh#813.

Successful run of these test suites:

```shell
bundle exec rake spec:unit:parallel

DB=sqlite     bundle exec rake spec:unit:director
DB=postgresql bundle exec rake spec:unit:director
DB=mysql      bundle exec rake spec:unit:director
```

This change brings no new feature, and only more clarity in the ERB rendering code.

Definitely not.

I’m a [poor lonesome cowboy](https://www.pipelinecomics.com/wp-content/uploads/2020/10/lucky_luke_75_the_end.jpg?ezimgfmt=ngcb1/notWebP) 😄

### DIFF
--- a/src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
+++ b/src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
@@ -3,12 +3,39 @@ require 'bosh/director/core/templates/rendered_job_instance'
 require 'bosh/director/formatter_helper'
 
 module Bosh::Director::Core::Templates
+
+  # @param [Array<DeploymentPlan::Job>] instance_jobs
+  # @param [JobTemplateLoader] job_template_loader
   class JobInstanceRenderer
     def initialize(templates, job_template_loader)
       @templates = templates
       @job_template_loader = job_template_loader
     end
 
+    # Render all templates for a Bosh instance.
+    #
+    # From a list of instance jobs (typically comming from a single instance
+    # plan, so they cover all templates of some instance) this method is
+    # responsible for orchestrating several tasks.
+    #
+    # Lazily-delegated work to a 'JobTemplateLoader' object:
+    #   - Load all templates of the release job that the instance job is
+    #     referring to
+    #   - Convert each of these to a 'JobTemplateRenderer' object
+    #
+    # Work done here on top of this:
+    #   - Render each template with the necessary bindings (comming from
+    #     deployment manifest properties) for building the special 'spec'
+    #     object that the ERB rendring code can use.
+    #
+    # The actual rendering of each template is delegated to its related
+    # 'JobTemplateRenderer' object, as created in the first place by the
+    # 'JobTemplateLoader' object.
+    #
+    # @param [Hash] spec_object A hash of properties that will finally result
+    #                           in the `spec` object exposed to ERB templates
+    # @return [RenderedJobInstance] An object containing the rendering results
+    #                               (when successful)
     def render(spec)
       errors = []
 
--- a/src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
+++ b/src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
@@ -7,8 +7,8 @@ module Bosh::Director::Core::Templates
   # @param [Array<DeploymentPlan::Job>] instance_jobs
   # @param [JobTemplateLoader] job_template_loader
   class JobInstanceRenderer
-    def initialize(templates, job_template_loader)
-      @templates = templates
+    def initialize(instance_jobs, job_template_loader)
+      @instance_jobs = instance_jobs
       @job_template_loader = job_template_loader
     end
 
@@ -36,14 +36,14 @@ module Bosh::Director::Core::Templates
     #                           in the `spec` object exposed to ERB templates
     # @return [RenderedJobInstance] An object containing the rendering results
     #                               (when successful)
-    def render(spec)
+    def render(spec_object)
       errors = []
 
-      rendered_templates = @templates.map do |template|
-        job_template_renderer = job_template_renderers[template.name]
+      rendered_templates = @instance_jobs.map do |instance_job|
+        job_template_renderer = job_template_renderers[instance_job.name]
 
         begin
-          job_template_renderer.render(spec)
+          job_template_renderer.render(spec_object)
         rescue Exception => e
           errors.push e
         end
@@ -51,7 +51,7 @@ module Bosh::Director::Core::Templates
 
       if errors.length > 0
         combined_errors = errors.map{|error| error.message.strip }.join("\n")
-        header = "- Unable to render jobs for instance group '#{spec['job']['name']}'. Errors are:"
+        header = "- Unable to render jobs for instance group '#{spec_object['name']}'. Errors are:"
         message = Bosh::Director::FormatterHelper.new.prepend_header_and_indent_body(header, combined_errors.strip, {:indent_by => 2})
         raise message
       end
@@ -62,9 +62,9 @@ module Bosh::Director::Core::Templates
     private
 
     def job_template_renderers
-      @job_template_renderers ||= @templates.reduce({}) do |hash, template|
-        hash[template.name] = @job_template_loader.process(template)
-        hash
+      @job_template_renderers ||= @instance_jobs.reduce({}) do |renderers_hash, instance_job|
+        renderers_hash[instance_job.name] = @job_template_loader.process(instance_job)
+        renderers_hash
       end
     end
   end
--- a/src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
+++ b/src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
@@ -95,6 +95,10 @@ module Bosh::Director::Core::Templates
           release: release
         )
 
+        job_model = double('Bosh::Director::Models::Template')
+        expect(job).to receive(:model).and_return(job_model)
+        expect(job_model).to receive(:spec).and_return({ "templates" => { "test" => "test_dst" } })
+
         monit_erb = instance_double(SourceErb)
         job_template_erb = instance_double(SourceErb)
         fake_renderer = instance_double(JobTemplateRenderer)
@@ -136,6 +140,10 @@ module Bosh::Director::Core::Templates
           blobstore_id: 'blob-id',
           release: release,
         )
+
+        job_model = double('Bosh::Director::Models::Template')
+        expect(job).to receive(:model).and_return(job_model)
+        expect(job_model).to receive(:spec).and_return({ "templates" => {} })
 
         monit_erb = instance_double(SourceErb)
         fake_renderer = instance_double(JobTemplateRenderer)
@@ -181,6 +189,10 @@ module Bosh::Director::Core::Templates
             release: release,
             model: double('Bosh::Director::Models::Template', provides: [])
           )
+
+          job_model = double('Bosh::Director::Models::Template')
+          expect(job).to receive(:model).and_return(job_model)
+          expect(job_model).to receive(:spec).and_return({ "templates" => {} })
 
           job_template_renderer = job_template_loader.process(job)
           expect(job_template_renderer.source_erbs).to eq([])
--- a/src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
+++ b/src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
@@ -79,7 +79,7 @@ module Bosh::Director::Core::Templates
 
       it 'returns the jobs template erb objects' do
         tarball_path = create_job_tarball(
-          'release-job-name',
+          'fake-job-name-1',
           'monit file erb contents',
           tmp_file,
           { 'test' => {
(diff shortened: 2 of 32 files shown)


src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
```diff
@@ -3,28 +3,55 @@
...
+          job_template_renderer.render(spec_object)
         rescue Exception => e
           errors.push e
         end
       end
 
       if errors.length > 0
         combined_errors = errors.map{|error| error.message.strip }.join("\n")
-        header = "- Unable to render jobs for instance group '#{spec['job']['name']}'. Errors are:"
+        header = "- Unable to render jobs for instance group '#{spec_object['name']}'. Errors are:"
```
aramprice, line 54: Should this be `spec_object['job']['name']` instead of `spec_object['name']`?

src/bosh-director/lib/bosh/director/deployment_plan/instance_group.rb
```diff
@@ -409,10 +409,19 @@ def create_desired_instances(count, deployment)
 
       private
 
+      # Returns the agregated list of package names, considering all jobs
```
aramprice, line 412: agregated => aggregated

src/bosh-director/lib/bosh/director/deployment_plan/instance_group.rb
```diff
@@ -409,10 +409,19 @@ def create_desired_instances(count, deployment)
...
+      # desired on this instance group, with no duplicate.
+      #
+      # @return [Array<String>] The list of package names
       def run_time_dependencies
         run_time_packages.map(&:name)
       end
 
+      # The agregated list of models for all packages to install on instances
```
aramprice, line 420: agregated => aggregated

src/bosh-director/lib/bosh/director/jobs/update_release.rb
```diff
@@ -304,18 +305,18 @@ def use_existing_jobs(jobs, release_dir)
         return false if jobs.empty?
 
         single_step_stage("Processing #{jobs.size} existing job#{'s' if jobs.size > 1}") do
-          jobs.each do |template, job_meta|
-            job_desc = "#{template.name}/#{template.version}"
+          jobs.each do |job_model, manifest_job|
+            job_desc = "#{job_model.name}/#{job_model.version}"
```
aramprice, line 309: At some point maybe the "#{name}/#{version}" moniker should be extracted into a method rather than having this pattern of string interpolation.

I don't think this is necessary for this PR but I thought it was worth including a comment in case there are more refactors coming.

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
```diff
@@ -79,7 +79,7 @@ module Bosh::Director::Core::Templates
 
       it 'returns the jobs template erb objects' do
         tarball_path = create_job_tarball(
-          'release-job-name',
+          'plop-job-name',
```
aramprice, line 82: Is there a reason for the change from `release` => `plop`? If this is a meaningless identifier maybe use something like `FAKE_XYZ...`

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
```diff
@@ -103,19 +107,18 @@ module Bosh::Director::Core::Templates
...
-          'plan-job-name',
+          'plop-job-name',
         ).and_return(monit_erb)
 
         expect(SourceErb).to receive(:new).with(
           'test',
           'test_dst',
           'test contents',
-          'plan-job-name'
+          'plop-job-name'
```
aramprice, line 117: Is there a reason for the change from `plan` => `plop`? If this is a meaningless identifier maybe use something like `FAKE_XYZ...` or `FAKE_123...`

src/bosh-template/lib/bosh/template/evaluation_context.rb
```diff
@@ -40,11 +40,7 @@ def initialize(spec, dns_encoder)
                     "Hash expected, #{spec.class} given"
         end
 
-        if spec['job'].is_a?(Hash)
-          @name = spec['job']['name']
-        else
-          @name = nil
-        end
+        @name = spec['name']
```
aramprice, line 43: Curious about this change, is this always safe, even if a v1 manifest is being consumed?

rkoster:
@bgandon did you get a change yet to look at the changes requested by Aram?

src/bosh-director-core/lib/bosh/director/core/templates/job_instance_renderer.rb
bgandon: I’ve deliberately switched to using `spec.name` instead of `spec.job.name`, which holds the same information anyway.

For the reason why, first see: https://bosh.io/docs/jobs/#deprecated-properties-accessors

Indeed, I’ve documented there the `spec.job.*` properties (where “job” is the Bosh v1 name for “intance group” and “ template” the old Bosh v1 name for “job”, so that’s old). To my understanding, these have accidentally leaked in the `spec` object, as there is no reason for `spec.job.*` to be exposed to release authors because the data there are not useful for rendering ERB templates.

That’s why I’ve started favoring `spec.name` over `spec.job.name`. For preparing future removal of `spec.job`.

src/bosh-director/lib/bosh/director/jobs/update_release.rb
```diff
@@ -242,29 +243,29 @@ def process_jobs(release_dir)
...
 
             raise(
               ReleaseExistingJobFingerprintMismatch,
-              "job '#{job_meta['name']}' had different fingerprint in previously uploaded release '#{@name}/#{@version}'",
+              "job '#{manifest_job['name']}' had different fingerprint in previously uploaded release '#{@name}/#{@version}'",
             )
           end
 
-          jobs = Models::Template.where(fingerprint: job_meta['fingerprint']).all
+          job_models = Models::Template.where(fingerprint: manifest_job['fingerprint']).all
```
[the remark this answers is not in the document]
bgandon: About `Models::Template` representing a “job”, I’ve updated the comment of the class there:
https://github.com/cloudfoundry/bosh/pull/2479/files#diff-56d25d21526860acc6209368812592a1819f9db61fd20db419dabdca49173433R2-R5

I’ve thought this to be sufficient for the moment. Tell me what you think about this.

For a first refactoring, I’ve also deliberately chosen not to rename `Models::Template` in order to avoid renaming tables, and broader and more risky impacts. But it would be good to do it at some point.

src/bosh-director/lib/bosh/director/jobs/update_release.rb
bgandon: Yep, I agree with this indeed

bgandon:
I’m in the process of reviewing the comments

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
bgandon: Yep, “plop” just conventionally mean “_whatever name_” and is meant to be more fun to use than “fake”. “plip” is quite the same, maning “_whatever name, but different from ‘plop’_”.

All right for favoring “fake” over “plip/plop” 👍

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
```diff
@@ -128,29 +131,32 @@ module Bosh::Director::Core::Templates
       end
 
       it 'includes only monit erb object when no other templates exist' do
-        tarball_path = create_job_tarball('release-job-no-templates', 'monit file erb contents', tmp_file, {})
+        tarball_path = create_job_tarball('plip-job-name', 'monit file erb contents', tmp_file, {})
```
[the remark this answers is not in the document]
bgandon: There has been a reason indeed 🙂. Having different names for `plan-job-name` and `release-job-name` was a defect, that hadn’t show up before because `spec.name` wasn’t used.

There were no reason for distinguishing “plan” or “release” job, because there where designating the very same thing. These two names actually had to be the same. That’s how I started to use “plop“, and then when I needed a different name later on, I’ve used “plip”.

But I get your point. I’ll use `fake-job-name-1` and `fake-job-name-2` instead.

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
bgandon: I’ll use `fake-job-name-1` and `fake-job-name-2` instead of plip/plop.

src/bosh-template/lib/bosh/template/evaluation_context.rb
bgandon: Considering that “job” means “instance group”, if the instance group is not a Hash then what could it be? An array? `nil`? Is this something that can happen?

To my knowledge, from within an ERB file that needs being rendered, the instance group _must_ be a Hash. If it’s not, then something bad is happening and then the code should crash.

Basically I’ve applied the “_let it crash_” principle: no reason to protect against a situation that should never happen, and possibly maintain unnecessary code or hide a bug.

src/bosh-director-core/spec/bosh/director/core/templates/job_template_loader_spec.rb
aramprice: I kinda figured this was the case. I think over time we've found that  "fun" names tend to be less helpful that boring descriptive names. Sadly.

aramprice (approved):
thanks for the improvements!

bgandon:
Due to the large diff, we are waiting for a second review, thanks for anyone that can do it!

jpalermo (approved):
👍

A lot of changes, but functionally seems to be the same. A lot of great cleanup though.

Thanks!

rkoster:
Thanks! @bgandon

aramprice:
This change appears to have exposed a problem with [src/spec/assets/valid_release.tgz](https://github.com/cloudfoundry/bosh/blob/main/src/spec/assets/valid_release.tgz).

Integration tests which attempt to deploy this release using a command similar to:
> `bosh [redacted args] upload-release src/spec/assets/valid_release.tgz`

Fail with the following error:
```
Using environment 'https://127.0.0.1:62404' as client 'test'
[-----------------------------------------------------------------] 100.00%   0s
Task 6

Task 6 | 17:40:22 | Extracting release: Extracting release (00:00:00)
Task 6 | 17:40:22 | Verifying manifest: Verifying manifest (00:00:00)
Task 6 | 17:40:22 | Resolving package dependencies: Resolving package dependencies (00:00:00)
Task 6 | 17:40:22 | Creating new packages: stuff/0.1.17 (00:00:01)
Task 6 | 17:40:23 | Creating new packages: mutator/2.99.7 (00:00:00)
Task 6 | 17:40:23 | Creating new jobs: cacher/1a (00:00:00)
                 L Error: Inconsistent name for job 'cacher'(exptected: 'cacher', got: '')
Task 6 | 17:40:23 | Error: Inconsistent name for job 'cacher'(exptected: 'cacher', got: '')

Task 6 Started  Thu Jan 11 17:40:22 UTC 2024
Task 6 Finished Thu Jan 11 17:40:23 UTC 2024
Task 6 Duration 00:00:01
Task 6 error
```

This appears to be run from [spec/integration/release/export_release_spec.rb:222](https://github.com/cloudfoundry/bosh/blob/main/src/spec/integration/release/export_release_spec.rb#L222)

beyhan:
I didn't have time to get to the bottom of this but it looks like there is a new validation now https://github.com/cloudfoundry/bosh/blob/dcb90acf53b19bee64ac07ed0565e29fb5e1b847/src/bosh-director/lib/bosh/director/jobs/release/release_job.rb#L79-L84

Actually this change shouldn't have any behaviour changes. @bgandon why that check was added?

aramprice:
@beyhan - nice catch! No need to dig in I only figured this out at the end of my yesterday.

It's likely that `src/spec/assets/valid_release.tgz` is making some "bosh v1" assumptions.

aramprice:
Hopefully fixed with https://github.com/cloudfoundry/bosh/commit/e42f7670484516ad0dcfb2ae8375fb051db085b2

beyhan:
@aramprice thanks for the heads-up. I also saw that the last update of that release was way back in 2015 and had an old structure. Is the support for "bosh v1" removed with this refactoring or it was already the case and the validation here just shows it? If it is with this pr I would suggest to have this in the release notes. FYI @rkoster, @jpalermo

jpalermo:
I think v1 manifests should still work, I don't think release job spec file structure changed much between v1 and v2, and I would have thought it was "required" that jobs in v1 also had a name that matched their actual name.

I kinda bet something blows up at some point if this isn't true, so I don't think this is really a change in behavior, probably just surfaces the error sooner. (could be wrong about all of this though)

PR #2482, opened 2023-12-13
enable metrics for deploy config

fmoehler:

We introduced deploy configs with Bosh CLI v7.5.0+. As this could potentially lead to unwanted deploy flags if the config is forgotten/enabled too long, we introduce a metrics that returns 1 in case a deploy config is uploaded.

https://github.com/cloudfoundry/docs-bosh/blob/master/content/deploy-config.md

Create Dev Release and checked the response of the metrics server 

The default metrics now return if a deploy config was uploaded or not to avoid it being forgotten

No

@AttilaAlmasi

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/config_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/config_manager.rb
@@ -16,10 +16,7 @@ module Bosh
           def deploy_config_enabled?
             deploy_config = find(type: 'deploy', limit: 999)
 
-            deploy_config.each do |config|
-              return true
-            end
-            false
+            return !deploy_config.empty?
           end
 
           def find(type: nil, name: nil, limit: 1)
--- a/src/bosh-director/lib/bosh/director/api/config_manager.rb
+++ b/src/bosh-director/lib/bosh/director/api/config_manager.rb
@@ -13,6 +13,15 @@ module Bosh
             config.save
           end
 
+          def deploy_config_enabled?
+            deploy_config = find(type: 'deploy', limit: 999)
+
+            deploy_config.each do |config|
+              return true
+            end
+            false
+          end
+
           def find(type: nil, name: nil, limit: 1)
             dataset = Bosh::Director::Models::Config.where(deleted: false)
             dataset = dataset.where(type: type) if type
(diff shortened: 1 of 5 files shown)


AttilaAlmasi (approved):

src/bosh-director/lib/bosh/director/api/config_manager.rb
```diff
@@ -13,6 +13,15 @@ def create(type, name, config_yaml, team_id = nil)
             config.save
           end
 
+          def deploy_config_enabled?
+            deploy_config = find(type: 'deploy', limit: 999)
+
+            deploy_config.each do |config|
+              return true
+            end
+            false
```
a-hassanin, line 22: May it is better to write it something like this:
return !deploy_config.empty?

src/bosh-director/spec/unit/api/config_manager_spec.rb
```diff
@@ -282,5 +282,19 @@
...
   end
 
+  describe '#deploy_config_enabled?' do
+    context 'when config of type deploy does not exist' do
+      it "returns 'false'" do
+        expect(manager.deploy_config_enabled?).to eq(false)
+      end
+    end
+
+    context 'when config of type deploy does not exist' do
```
a-hassanin, line 292: typo: when config of type deploy exists

a-hassanin (changes requested):
minor suggestions

src/bosh-director/lib/bosh/director/api/config_manager.rb
fmoehler: Thanks! Leftover from an old approach.

a-hassanin (approved):
LGTM

src/bosh-director/lib/bosh/director/api/config_manager.rb
```diff
@@ -13,6 +13,12 @@ def create(type, name, config_yaml, team_id = nil)
             config.save
           end
 
+          def deploy_config_enabled?
+            deploy_config = find(type: 'deploy', limit: 999)
```
ystros, line 17: Is there any significance for choosing a limit of 999 here? It seems like you could do the (default) limit of 1, since you only care about the existence of the data, not the data itself.

fmoehler: Thanks for the review! You are right. I have changed it.

ystros (approved):
lgtm!

beyhan (approved):

a-hassanin (approved):
LGTM

rkoster:
Thanks! @fmoehler

PR #2484, opened 2024-01-03
Populate stemcell os when manifest contains only the name

jpalermo:
Follow up to: https://github.com/cloudfoundry/bosh/pull/2475

The previous PR didn't handle diffs correctly so it showed that runtime configs were getting added/removed even though they weren't.

This commit moves the OS population up into the manifest parsing so it can apply to both the deployment manifest and the diff manifest.

This caused the DeploymentPlan Stemcell to start failing because it had a validation that both Name and OS shouldn't be specified. We moved that validation up to the ManifestValidator which seems like a more appropriate place for it anyway.

We modified the manifest parsing to prefer Name over OS when resolving the stemcell version aliases. This is more consistent with how the DeploymentPlan Stemcell resolves Name and OS. It shouldn't be a behavioral change though since you can't specify both Name and OS.

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/manifest_validator.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/manifest_validator.rb
@@ -28,6 +28,14 @@ module Bosh
                   'resource_pools is no longer supported. You must now define resources in a cloud-config'
           end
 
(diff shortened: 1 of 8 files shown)


aramprice (approved):
lgtm

PR #2487, opened 2024-01-11
Bump openssl gem to 3.2.0 for FIPS compatibility

ystros:
openssl 3.2.0 gem implements features necessary for running OpenSSL-related code in Ruby in a FIPS environment. We've added this as a minimum requirement for each gem that makes up the overall BOSH Director code.

It should be possible to remove this requirement once we upgrade to Ruby 3.3, which ships with openssl 3.2.0 gem as a standard gem.


jpalermo (approved):
👍

PR #2488, opened 2024-01-22
Remove json minor-version-dependency lock

a-hassanin:

The currently used version of json library (json 2.6.3) has not been updated for more than 1 year.

This change updates the gems to the latest version with no known vulnerabilities. `bundle update json `was used to ensure indirect dependencies are updated as well.

unit tests

bump json gem

No

### DIFF
--- a/packages/director/packaging
+++ b/packages/director/packaging
@@ -29,7 +29,7 @@ popd > /dev/null
 
 cat > Gemfile <<EOF
 # Explicitly require vendored version to avoid requiring builtin json gem
-gem 'json', '2.6.3'
+gem 'json', '~>2'
 
 gem 'bosh-director'
 gem 'mysql2'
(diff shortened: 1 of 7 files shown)


beyhan:
@a-hassanin could you please remove the version so that we get the latest. We don't need an exact version for json.

rkoster:
@a-hassanin the ask is to remove the version constraint in the Gemfile. So it can be auto-bumped by the existing automation.

jpalermo:
@a-hassanin , do you have a link for that Rubygems 4 deprecation? I wasn't able to find anything about it, but that sounds like it will have a major impact on quite a few projects.

a-hassanin:
> @a-hassanin , do you have a link for that Rubygems 4 deprecation? I wasn't able to find anything about it, but that sounds like it will have a major impact on quite a few projects.

@jpalermo This warning is raised when I bundle:
https://docs.ruby-lang.org/en/master/Gem/Version.html#method-c-nil_versions_are_discouraged-21

It seems this warning exists for a while already as I see some projects already start to handle it more than 2 years ago.

max-soe:
@rkoster / @jpalermo Can we bring this in?

jpalermo (approved):
👍

PR #2489, opened 2024-01-25
Specify ruby patch version

a-hassanin:

Specifying the ruby patch version: from 3.2 => 3.2.2

In my and some colleagues local environments bundle install/update fail without this value, in addition to some unit tests which seems to depend on this patch version. Please check in your local environments if it is the same for you and merge this accordingly. 

unit tests

Specifying the ruby patch version

No


jpalermo:
Hi @a-hassanin 

If we add the patch, we'd need to make sure it always matches the value of the `director-ruby-3.2` or the release probably fails.

We talked in the FIWG meeting about maybe making this a part of the auto bump of the ruby package; but we've seen problems in the past where introducing patch versions causes problems with CI because the ruby version in the docker images used to run tests often has a bit of time where it's out of sync with the automatically bumped version in the repo.

I don't currently experience problems with running `bundle`. I think I'm using `chruby` and it just picks the latest patch of ruby 3.2 to use each time, which I thought most ruby version managers would do a similar thing.

What ruby version manager are you using that is causing problems?

a-hassanin:
> Hi @a-hassanin
> 
> If we add the patch, we'd need to make sure it always matches the value of the `director-ruby-3.2` or the release probably fails.
> 
> We talked in the FIWG meeting about maybe making this a part of the auto bump of the ruby package; but we've seen problems in the past where introducing patch versions causes problems with CI because the ruby version in the docker images used to run tests often has a bit of time where it's out of sync with the automatically bumped version in the repo.
> 
> I don't currently experience problems with running `bundle`. I think I'm using `chruby` and it just picks the latest patch of ruby 3.2 to use each time, which I thought most ruby version managers would do a similar thing.
> 
> What ruby version manager are you using that is causing problems?

@jpalermo Thank you for the explanation.
I am using [asdf](https://asdf-vm.com/) but as the version is locked to 3.2 it somehow does not seem to work for some of us.

I will live then with changing it manually on my side until there is a better solution

jpalermo:
Yeah, it looks like the asdf ruby plugin doesn't support fuzzy matching: https://github.com/asdf-vm/asdf-ruby/blob/master/README.md?plain=1#L63-L65

a-hassanin:
> Yeah, it looks like the asdf ruby plugin doesn't support fuzzy matching: https://github.com/asdf-vm/asdf-ruby/blob/7a221427fbbec030baeb8151c763fdf205dbb204/README.md?plain=1#L63-L65

Thank you, i have updated the development documentation accordingly: https://github.com/cloudfoundry/bosh/pull/2491

PR #2492, opened 2024-01-30
Enable signed url for bosh azure storage blobstore

Malsourie:

This change make it possible to use signed url if azure storage is used for external blobstore.

`test_unit` and `test_unit_erb`, we have also created a dev_release and tested on our own dev_landscape.

Enable signed url for bosh azure storage blobstore

No.

SAP BOSH team


rkoster (approved):

rkoster:
Thanks! @Malsourie

PR #2493, opened 2024-02-01
docker in docker service fails to start

nouseforaname:
https://github.com/docker/cli/issues/4807

As of 02/01/2024 a change in the "/etc/init.d/docker" script shipped with docker v25 is preventing the cpi image to work.
when start-bosh runs `service docker start` it errors with: "/etc/init.d/docker: 62: ulimit: error setting limit (Invalid argument)"

Workaround recommended in github issue suggests to disable resetting ulimit.

Pre v25 the script contained `ulimit -n 1048576`, with v25 it changed to: `ulimit -Hn 524288`.

the default in our base image is:

ulimit -a | grep file
`open files                          (-n) 1048576`

so it was a noOp before the change shipped with v25 of docker

to confirm I tried running `ulimit -Hn 1048576` and it will succeed..

The issue happens when we want to actually change the ulimit. In our case it would attempt to lower it and it's probably fine not doing that in CI?

### DIFF
--- a/ci/dockerfiles/docker-cpi/install.sh
+++ b/ci/dockerfiles/docker-cpi/install.sh
@@ -34,4 +34,18 @@ apt-get update
 
 apt-get install -y --no-install-recommends docker-ce
 
+# https://github.com/docker/cli/issues/4807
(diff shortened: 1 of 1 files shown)


lnguyen (approved):
LGTM

PR #2494, opened 2024-02-05
Blobstore clients supports PUT headers

mvach:

When putting blobs via signed urls to Azure Blobstore, then the mandatory header 'x-ms-blob-type' has to be set to value 'blockblob'. Therefore we add a mechanism to support PUT headers.

We are doing it just the same why like encryption headers can be set for GCS.

### DIFF
--- a/src/bosh-director/lib/bosh/blobstore_client/azurestoragecli_blobstore_client.rb
+++ b/src/bosh-director/lib/bosh/blobstore_client/azurestoragecli_blobstore_client.rb
@@ -40,6 +40,7 @@ module Bosh::Blobstore
     def redacted_credential_properties_list
       %w[account_key]
     end
+
     def headers
       { 'x-ms-blob-type' => 'blockblob' }
     end


mvach:
Hi @jpalermo,
we did not test this PR right now but just want to double check if the approach is right from you perspective.
So you feedback is very welcome :-)

jpalermo:
I haven't looked at it super closely, but I wonder if we can just combine all the headers in the blobstore client classes and rename `signed_url_encryption_header` to just `signed_url_headers` and let the blobstore object determine which headers should be included.

mvach:
After incorporating the feedback we started validating the PR in our local test environment.

mvach:
We tested this PR successfully on our azure landscape.

mvach:
Hi @jpalermo, hi @aramprice,
a review would be very nice.

fmoehler (approved):

src/bosh-director/lib/bosh/blobstore_client/azurestoragecli_blobstore_client.rb
```diff
@@ -40,11 +40,8 @@ def initialize(options)
     def redacted_credential_properties_list
       %w[account_key]
     end
```
Malsourie, line 42: Empty line between methods

Malsourie (approved):

src/bosh-director/spec/unit/deployment_plan/stages/package_compile_stage_spec.rb
```diff
@@ -353,9 +353,7 @@ def compile_package_with_url_encrypt_stub(args)
       end
     end
 
-    context 'when none of the packages are compiled' do
-      let(:instance_groups_to_compile) { [@j_dea, @j_router] }
-
+    shared_examples 'compiles all packages' do |compilation_method, compilation_method_stub|
```
aramprice, line 356: The changes from here down to line 384 seem unrelated to headers. Is this an intentional change?

mvach:
Hi @aramprice,
if you are fine with the PR, then an approval would be nice.

src/bosh-director/lib/bosh/blobstore_client/azurestoragecli_blobstore_client.rb
mvach: Hi @Malsourie,
thx for the hint. I changed it.

jpalermo (approved):
👍

PR #2497, opened 2024-02-15
Switch metrics server from thin to puma 

ystros:
In an effort to get rid of EventMachine, which has been unmaintained for some time, we are switching from using Thin web server (which is based on EventMachine) to Puma instead.

Since there don't seem to be unit specs for this bin script, we applied our patch to a running director to ensure that the metrics server comes up as expected.

See also #2499, #2500 for more removals of EventMachine.

### DIFF
--- a/jobs/health_monitor/spec
+++ b/jobs/health_monitor/spec
@@ -52,7 +52,7 @@ properties:
     default: info
 
   hm.em_threadpool_size:
-    description: EM thread pool size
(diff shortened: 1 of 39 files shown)


aramprice (approved):

mvach (approved):
Nice

a-hassanin (approved):
LGTM!

PR #2499, opened 2024-02-22
Remove EventMachine from nats package

ystros:
This switches from using the EventMachine scheduler functionality to
using Rufus Scheduler, which is already used by the metrics collector in
the Director.

See also #2497, #2500 for more removals of EventMachine.


aramprice (approved):

lnguyen (approved):
LGTM

PR #2502, opened 2024-03-04
Remove EventMachine from Health Monitor and the Director

ystros:

The eventmachine gem is no longer receiving updates. This commit moves
asynchronous functionality of bosh-monitor / Health Monitor to use the
async gem (and associated ecosystem) instead.

* em-http-request gem is replaced by async-http gem
* EventMachine connections are replaced with async-io gem
* EventMachine's SMTP functionality has been replaced with net-smtp gem
  wrapped in async blocks.
* EventMachine.popen is replaced with Open3 wrapped in async blocks
* The em_threadpool_size config has been removed as there does not
  appear to be an analog in the async gem

Previous PRs removing other instances of EventMachine: #2497, #2499, #2500

Overall tests:
* ✅ Unit and functional specs
* ✅ Manually deployed a BOSH Director using a dev release created from these changes, and observed the Health Monitor logs
* ✅  BRATS (tests JSON plugin)

Plugins:
* ❓ Consul (this appears to be undocumented and perhaps should be removed?)
* ❓ Datadog
* ✅  Email
* ✅  Event logger
* ✅  Graphite (confirmed that the Graphite server still receives events, not familiar enough to go beyond that)
* ✅ JSON (through unit tests, local testing, and BRATS)
* ❓ PagerDuty
* ✅ Resurrector
* ❓ Riemann
* ❓ TSDB

Very notably, several individual Health Monitor plugins have not been verified to work. The biggest risk plugins seem to be Email, Graphite, and TSDB, as they required the most rewrites due to high coupling with EventMachine. Email and Graphite have received some very basic manual tests.

Not necessarily a breaking change, but the `em_threadpool_size` config option for Health Monitor has been removed with no analogous configuration option added. EventMachine seems to have created a fixed-size thread pool to run its event loop whereas the async gem uses Ruby fibers.

This does mean there may be a slight risk in certain places in the code of spawning many async tasks simultaneously that would have been previously limited by the size of the EventMachine thread pool; in those places, I've added a semaphore to limit the number of simultaneous tasks.


src/bosh-monitor/lib/bosh/monitor/plugins/json.rb
```diff
@@ -68,39 +73,80 @@ def restart_process(bin)
...
+      ensure
+        @stdin.close if @stdin
+        @stdout.close if @stdout
+        @stderr.close if @stderr
+      end
     end
 
     def self.open(cmd)
-      EventMachine.popen(cmd, DeferrableChildProcess)
+      new(cmd)
```
aramprice, line 135: Is there a reason to keep `DeferrableChildProcess.open` or could `DeferrableChildProcess.new` be inlined wherever it's called?

aramprice (approved):
lgtm

src/bosh-monitor/lib/bosh/monitor/plugins/json.rb
ystros: It could be inlined; part of the reason I didn't is for a couple iterations this method did more, I wanted to minimize interface changes where possible, and it's mocked in tests (mocking constructors is really no different, but always feels worse to me)

lnguyen (approved):
LGTM

PR #2506, opened 2024-03-18
Unpin `riemann-client` gem and upgrade

aramprice:
Per the project's [Changelog](https://github.com/riemann/riemann-ruby-client/blob/main/CHANGELOG.md) there have been no breaking changes between `0.2.6` and `1.2.1`

Updating the riemann-client gem.

n/a

✅ `rake fly:unit`
✅ `rake fly:integration`

> Gem used by `bosh-minotor` Riemann plugin updated

None known.

### DIFF
--- a/src/Gemfile.lock
+++ b/src/Gemfile.lock
@@ -75,7 +75,7 @@ PATH
       net-smtp
       openssl (>= 3.2.0)
       puma
-      riemann-client (~> 0.2.6)
+      riemann-client
       sinatra (~> 2.2.0)
 
 PATH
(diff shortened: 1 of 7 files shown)


mvach:
I asked our Riemann experts if they are aware of issues when doing this version jump and are waiting for their feedback.

mvach:
Hmmm our Riemann people mentioned that they don't know the ruby client but suggested to test the new version carefully :-)

jpalermo (approved):
From the sound of it, this seems safe enough, but if it breaks anything we can roll it back.

PR #2507, opened 2024-03-18
Use `NetAddr::CIDR.create` instead of `CIDRv{4,6}.new`

aramprice:

The sub-classes of `CIDR` do not define an `#initialize` method so these end up calling `CIDR#initialize` which has the following documentation:

> This method performs absolutely no error checking, and is meant to be
> used only by other internal methods for the sake of the speedier
> creation of CIDR objects. [snip]

See: https://github.com/dspinhirne/netaddr-rb/blob/version-1.5.2/lib/cidr.rb#L205-L216

This commit switches to `CIDR.create(..., Version: {4,6})` so that the address type is still explicitly set.

n/a

✅ `rake fly:unit`
✅ `rake fly:integration`

[nternal change, none needed]

No.


ystros (approved):
Somewhat confused why we are using these CIDR objects in the first place when it seems like we're always passing in singular IP addresses into the constructor. But this change from `v#.new` to `.create` looks fine.

aramprice:
> Somewhat confused why we are using these CIDR objects in the first place ...

That is an excellent question.

Another curiosity is why we have vendor'd the 1.x line of a gem which got a massive 2.x rewrite and has been removed from rubygems.org when we only use a small bit of its total functionality.

beyhan (approved):

PR #2511, opened 2024-04-12
Remove unused enable_post_deploy property

ystros:
The default for the enable_post_deploy property was changed from `false` to `true` in commit 8fbf0457d48fbf6ab1223ef5e9134585cbf43f57. However, as part of that commit, the property was removed from the director.yml config file template. The value also defaulted to `true` in the Config class. With these two factors combined, `enable_post_deploy` has been unconfigurable and always true.

This commit removes the property entirely rather than add it back to the config file template, since it hasn't been configurable in several years without complaint.

[#187358662]


jpalermo (approved):
👍

rkoster (approved):

rkoster:
Thanks! @ystros

PR #2512, opened 2024-04-18
Add shellwords to template rendering

jpalermo:
Shellwords was previously required by eventmachine (we believe), so when that was removed, any release templates that were trying to use shellwords without requiring it started failing.

Errors seen in template rendering:
```
undefined method `shellescape' for "":String
```


lnguyen:
LGTM

PR #2513, opened 2024-04-19
Fix Health Monitor email messages and TLS

ystros:

Previously, the EventMachine SMTP client was formatting the headers and body of emails sent by Health Monitor for us. When this was replaced with Net::SMTP, the headers and body were improperly formatted and thus emails were not delivered.

Additionally, this restores the STARTTLS behavior to how it worked when we used EventMachine; setting hm.email.tls to true forces a STARTTLS session, while setting it to false does not attempt a STARTTLS session even if the server supports it.

* Ran unit tests
* Applied patch to a live director to ensure emails were being sent out in both TLS and non-TLS mode.


jpalermo (approved):
👍

PR #2514, opened 2024-04-24
newer credhub versions require jdk 11 to run

nouseforaname:
see:

https://github.com/cloudfoundry/credhub/commit/4dab29337629ebe9df6f7b9cdd921038f652867e


jpalermo (approved):
Seems like it's probably more correct 🤷‍♂️

rkoster (approved):

PR #2518, opened 2024-05-08
Modern uaa integration tests

selzoc:

Use a modern UAA for integration tests

See commit

integration!


jpalermo:
@selzoc , I believe `ci/dockerfiles/integration/install-uaa.rb` is not used correct?

selzoc:
> @selzoc , I believe `ci/dockerfiles/integration/install-uaa.rb` is not used correct?

Ah yes I believe we moved that functionality into `src/bosh-dev/lib/bosh/dev/sandbox/services/uaa_service.rb`.

Removed in ecce4ec08845119df26bb87ba91b3f49658c69b2

cunnie (approved):
I like how you freshen UUA download URL every time.

PR #2519, opened 2024-05-11
Add shared ci task that can be used by release authors to test template rendering

jpalermo:
This makes it easier for release authors to test their template rendering against the latest bosh director.

We've broken template rendering a few times in the last couple years due to director changes.


aramprice (approved):

rkoster (approved):

selzoc:
FYI it looks like we should specify that this has to run in a privileged container, otherwise you get
```
mount: /tmp/devices-cgroup: permission denied.
```

jpalermo:
Forgot about the privileged requirement. Updated the docs section.

PR #2525, opened 2024-05-28
Exclude dev/test gems with GPL license to simplify license compliance

rkoster:
While looking at BlackDuck scan results I noticed that there are a few dev/test gems that bring (strong) copy left licenses. 
Since these gems are not a runtime dependency, let's try and exclude these from our final releases.

### DIFF
--- a/packages/director/spec
+++ b/packages/director/spec
@@ -15,3 +15,6 @@ files:
 - vendor/cache/*.gem
 - vendor/cache/extensions/**
 - vendor/cache/netaddr-rb-*/**
+
+excluded_files:
+- vendor/cache/{bundle-audit,bundler-audit,coderay}-*.gem # test dependency with GPL license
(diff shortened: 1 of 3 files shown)


klakin-pivotal:
@selzoc shouldn't we -at minimum- correct the comment introduced by the modification?

selzoc:
> @selzoc shouldn't we -at minimum- correct the comment introduced by the modification?

You mean `coderay` having an MIT license?  Yes, agreed.

rkoster:
I have not tested this change myself and was hoping to rely on the pipeline for that. I did create a release with these changes and verified the it resolved some of the license compliance issues Black Duck found, and it did.

rkoster:
I'm also thinking about maybe excluding all gems from test groups. What do you y'all think?

aramprice:
I'm worried about the fragility of hand-coded exclusions. Probably fine for the time being but perhaps there is a `bundle` flag to exclude non-production gems from `vendor/cache/` all together?

aramprice:
Its been ~2 years, should we close this? Or maybe revisit the idea in a more wholistic way for the CF ruby ecosystem?

PR #2529, opened 2024-06-11
Add an endpoint in bosh director for download a package with blobstore id 

nader-ziada:
Add an endpoint in bosh director for download a package with blobstore id 

you can use `bosh curl /packages/blob/<id>` to download your blob


ystros (approved):

selzoc (approved):

gm2552:
> Add an endpoint in bosh director for download a package with blobstore id
> 
> you can use `bosh curl /packages/blob/<id>` to download your blob

Quick correction: the url is template consists of  `blobs` and not `blob`

PR #2533, opened 2024-06-26
When adding a new tombstone record, delete all the older ones

jpalermo:
Noticed that these are never cleaned up currently. There may be a performance hit when the first new tombstone record is added and thousands of old ones are destroyed but it shouldn't be terrible.

PR #2536, opened 2024-07-11
Move local dns tombstone cleanup to scheduled job

jpalermo:
Follow up to #2533

The tombstone creation happens in a larger transaction, so if you both insert and clean tombstone records at the same time you can end up in a deadlock state if that is happening in multiple threads at the same time.

This commit moves the tombstone cleanup into a scheduled job that runs once a day instead.


selzoc (approved):

ystros (approved):

aramprice (approved):

aramprice:
thanks @jpalermo

PR #2537, opened 2024-07-17
Brings up Health Monitor HTTP server faster

klakin-pivotal:
## PR Summary

This commit adds a guard around the '/unresponsive_agents' endpoint so that it will return a "not successful" HTTP status code (in this case, 503) if the initial "query all Deployments and their Instances" action run by the Director has not yet completed.

To support that, it pushes down the guts of the fetch_deployments function into the InstanceManager class.

This commit also moves the start of the Health Monitor HTTP server to nearly the top of the 'Monitor#run' function. This should get the '/healthz' endpoint started as quickly as possible so that we don't get terminated by monit just because querying the state of the Director-managed deployments takes longer than ten seconds.

Why do this?

We've discovered that Monit gives a monitored service ~10 seconds to bring up its health-checking HTTP server before Monit declares the service dead and restarts it. What this means is that Director Health Monitors that are on underprovisioned VMs and/or that have large numbers of Deployments and/or Instances make take longer than 10 seconds to come up, and may find themselves in an unending restart loop.

The guard around the '/unresponsive_agents' endpoint is added to preserve the previous behavior of "Calls to /unresponsive_agents do not succeed until the initial query of Director-managed deployments completes.".

## Things to note

The changes to `notifying_plugins_spec.rb` should be somewhat-carefully reviewed. Moving the location of the HTTP server start causes a bunch of "Health monitor failed to connect to director" messages to get put in the `Bosh::Monitor::Plugins::Dummy` plugin event queue. Given that we can scan through all of the messages in the queue and eventually find the one we expect, it seems pretty clear that the test as written assumed that there would only ever be a single message in the event queue.

I did not bother to find out why the `"when health monitor fails to fetch deployments"` test succeeds. Perhaps it succeeds by coincidence, given the nature of the new failure messages?

I was unable to discover where the Bosh Director (or stub of the same) that this thing contacts was being brought up.

See above.

I have run most of the unit tests.

Improves Health Monitor startup reliability by bringing up the Health Monitor HTTP server as fast as is possible.

I do not believe so. The user-visible behavior change is as follows:

Prior to this change, attempts to access the Health Monitor's `/unresponsive_agents` endpoint would fail with "connection refused" until the first survey of the healthiness of all deployments had completed.
After this change, attempts to access the Health Monitor's `/unresponsive_agents` endpoint will return 503 until the first survey of the healthiness of all deployments had completed.

Both are unsuccessful returns, and both indicate server failure, rather than something that could be corrected client-side, so I believe that this is not a breaking change.

@aramprice

### DIFF
--- a/src/bosh-monitor/lib/bosh/monitor/api_controller.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/api_controller.rb
@@ -34,7 +34,11 @@ module Bosh::Monitor
     end
 
     get '/unresponsive_agents' do
-      JSON.generate(@instance_manager.unresponsive_agents)
+      if @instance_manager.director_initial_deployment_sync_done
+        JSON.generate(@instance_manager.unresponsive_agents)
+      else
+        status(503)
+      end
     end
   end
 end
--- a/src/bosh-monitor/lib/bosh/monitor/instance_manager.rb
+++ b/src/bosh-monitor/lib/bosh/monitor/instance_manager.rb
@@ -3,6 +3,7 @@ module Bosh::Monitor
     attr_reader :heartbeats_received
     attr_reader :alerts_received
     attr_reader :alerts_processed
+    attr_reader :director_initial_deployment_sync_done
 
     attr_accessor :processor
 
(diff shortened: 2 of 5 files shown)


aramprice (approved):

klakin-pivotal:
Yeah, any situation where the Health Monitor is continually getting restarted every ten seconds, and not reporting some sort of failure in its logs is likely related to this.

PR #2538, opened 2024-07-22
Correct PR #2537

klakin-pivotal:

This change corrects the copy-and-paste error introduced in PR #2537 .

Integration tests for both the Director and Agent caught the error.

We ran the files containing the failing tests in the Concourse container running the tests in CI, so we have confidence that the change we made fixes the problem.

@rajathagasthya


selzoc (approved):
Nice catch

aramprice (approved):

PR #2539, opened 2024-07-30
remove the `tag` on the ubuntu docker image

mingxiao:
These tags are outdated, xenial and trusty, are out of support. Implicity use `latest`, the tasks are using common commands like `cp` and `mkdir`

[#187900893]

remove the `tag` on the ubuntu docker image
    
These tags are outdated, xenial and trusty, are out of support. Implicity use `latest`, the tasks are using common commands like `cp` and `mkdir`
    

None, these are pipeline changes.

There is no impact on release notes, these are pipeline changes

No

@rajathagasthya

### DIFF
--- a/ci/tasks/prepare-container-bosh-docker-build-context.yml
+++ b/ci/tasks/prepare-container-bosh-docker-build-context.yml
@@ -5,7 +5,6 @@ image_resource:
   type: docker-image
   source:
     repository: ubuntu
-    tag: "14.04"
 
 inputs:
   - name: bosh-src
--- a/ci/tasks/prepare-container-docker-cpi-build-context.yml
+++ b/ci/tasks/prepare-container-docker-cpi-build-context.yml
@@ -5,7 +5,6 @@ image_resource:
   type: docker-image
   source:
     repository: ubuntu
-    tag: xenial
 
 inputs:
   - name: bosh-src


aramprice (approved):

ci/tasks/prepare-container-bosh-docker-build-context.yml
```diff
@@ -5,7 +5,6 @@ image_resource:
   type: docker-image
   source:
     repository: ubuntu
-    tag: "14.04"
```
xtreme-nitin-ravindran, line 8: This will always pull the latest version of `ubuntu`, `noble` atm. Is that ok?

ci/tasks/prepare-container-docker-cpi-build-context.yml
```diff
@@ -5,7 +5,6 @@ image_resource:
   type: docker-image
   source:
     repository: ubuntu
-    tag: xenial
```
xtreme-nitin-ravindran, line 8: This will always pull the latest version of `ubuntu`, `noble` atm. Is that ok?

ci/tasks/prepare-container-bosh-docker-build-context.yml
mingxiao: yes, the ubuntu docker image used in other task use latest: https://github.com/cloudfoundry/bosh/blob/main/ci/pipeline.yml#L1427

The bash script this file runs only uses `cp` and `mkdir`

ci/tasks/prepare-container-docker-cpi-build-context.yml
mingxiao: yes, the ubuntu docker image used in other task use latest: https://github.com/cloudfoundry/bosh/blob/main/ci/pipeline.yml#L1427

The bash script this file runs only uses `cp` and `mkdir`

ci/tasks/prepare-container-bosh-docker-build-context.yml
mingxiao: also the `14.04` line is `trusty` which is older than the now out of support `xenial`

xtreme-nitin-ravindran (approved):

xtreme-nitin-ravindran (approved):

aramprice:
thinking it's probably fine to float on ubuntu `latest` since apparently that points to LTS and we are expecting to produce a Noble stemcell ... soon™ 
> The ubuntu:latest tag points to the "latest LTS", since that's the version recommended for general use. The ubuntu:rolling tag points to the latest release (regardless of LTS status).
> per: https://hub.docker.com/_/ubuntu/

PR #2540, opened 2024-07-30
Remove CidrRangeCombiner and spec, no uses found

aramprice:

Removing unused code, and spec.

`bundle exec rake spec:unit`

n/a

No.


klakin-pivotal:
I searched everything on my workstation, and the only reference to this class was in its test code. Let's get rid of it.

klakin-pivotal (approved):

ystros (approved):

PR #2543, opened 2024-08-05
Replace netaddr with ipaddr

aramprice:

Replace vendored  `NetAddr` gem dev version and replace it with the ruby standard `IPAddr` gem

- Unit
- Integration: https://bosh.ci.cloudfoundry.org/builds/148593283

n/a

No.


xtreme-nitin-ravindran (approved):

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
```diff
@@ -50,7 +50,7 @@ def allocate_dynamic_ip(reservation, subnet)
         retry
       end
 
-      @logger.debug("Allocated dynamic IP '#{ip_address.ip}' for #{reservation.network.name}")
```
ystros, line 53: Mostly commenting since it wasn't immediately intuitive to me: `IpAddr#to_s` returns only the IP address portion provided, even for CIDRs, and thus is equivalent to the old `NetAddr#ip`.

ystros (approved):

PR #2547, opened 2024-08-09
Configure CI to use ruby-3.3 package

aramprice:
- remove openssl versions, ruby 3.3 ships with a useable version for fips
- update bundler version manually
- update no method error text in specs to macth ruby 3.3

This PR _does not_ update the vendored bosh package to `ruby-3.3`, the change to the `pipeline.yml` should do this.


aramprice:
This should probably be merged after the other outstanding PR's

lnguyen (approved):

ystros (approved):

PR #2554, opened 2024-08-24
Remove unused rspec config tags

aramprice:
This branch is using Github actions for testing and should only be merged after https://github.com/cloudfoundry/bosh/pull/2551


jpalermo:
@aramprice seems good. Want to remove the duplicate github action commit from this PR too?

aramprice:
@jpalermo - rebased and the extra commit is gone 💨

PR #2557, opened 2024-08-31
Replace httpclient with net/http in bosh-monitor

aramprice:
Two methods in `Plugins::HttpRequestHelper` used the [httpclient](https://github.com/nahi/httpclient) gem to make HTTP(S) requests. This PR replaces that gem with Ruby's built-in `net/http` gem.

This change required modifying some of the callers of `HttpRequestHelper` (in bosh-director) since the response object from `net/http` differs slightly from `httpclient`'s, as well as the more mocked portions of the spec for `HttpRequestHelper` which relied on `httpclient` internals.

### DIFF
--- a/src/Gemfile
+++ b/src/Gemfile
@@ -28,7 +28,6 @@ group :production do
 end
 
 group :bat do
-  gem 'httpclient'
   gem 'minitar'
   gem 'net-ssh'
 end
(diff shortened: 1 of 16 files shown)


nader-ziada (approved):
lgtm

xtreme-nitin-ravindran (approved):

PR #2559, opened 2024-09-10
Cloud properties pw use

Sascha222:

We saw now in several situations that we are unable to solve issues fast, if we cannot access a failed or broken vm. We need access to the VM if the agent fails to startup so we can debug. 

We want to have control over vcap password without interfering with other deployments.

https://www.starkandwayne.com/blog/how-to-lock-vcap-password-for-bosh-vms/index.html

- A Unittest in create_vm_step_spec.rb
- We uploaded the code in the local dev landscape and deployed a VM. The password defined in cloud-config was taken by the vm
- Ran all unittest in this repository 

In case of fixed vcap-password in cloud-config the corresponding password is set during deployment of VM’s.

No as long no one configure a vcap password in cloud-config. 

@ansh-SAP @anshrupani  @a-hassanin  @Sascha-Stoj  @cf-bosh

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/steps/create_vm_step.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/steps/create_vm_step.rb
@@ -125,7 +125,10 @@ module Bosh::Director
           end
 
           password = env.fetch('bosh', {}).fetch('password', "")
-          if Config.generate_vm_passwords && password == ""
+          if !cloud_properties.dig('vcap_password').nil?
+            env['bosh'] ||= {}
+            env['bosh']['password'] = cloud_properties['vcap_password']
+          elsif Config.generate_vm_passwords && password == ""
             env['bosh'] ||= {}
             env['bosh']['password'] = sha512_hashed_password
           end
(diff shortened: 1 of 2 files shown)


aramprice:
@Sascha-Stoj - could you say more about why this change is wanted, and what it what problem it is addressing?

Sascha222:
> @Sascha-Stoj - could you say more about why this change is wanted, and what it what problem it is addressing?

Sascha222:
> @Sascha-Stoj - could you say more about why this change is wanted, and what it what problem it is addressing?

@aramprice - Hi commented the questionaire above. I hope it helps to clarify

jpalermo:
Discussion at the 12/9/2024 FIWG meeting notes:

Overall goals for this are being able to configure a password/ssh credentials to a VM where the agent is failing to start, but without having to modify the deployment manifest.

Ideal place for this is probably something new in the runtime configs. This allows placement rules to be used to configure this, as well as all other `env` hash properties.

CloudProperties isn't a great fit for these things because CloudProperties are properties that bosh doesn't know anything about and assumes the CPI knows about them. But the `env` hash is something bosh does know about and is made available to the agent.

aramprice (approved):

beyhan:
We decided to make this draft because the direction this should go is completely different one.

aramprice:
Is work still ongoing, or should we close this?

PR #2562, opened 2024-09-16
Taks name for bosh release template specs follows existing pattern

aramprice:


aramprice:
Merging as the spec failures under MySQL are a known issue not related to the DB starting.

PR #2567, opened 2024-09-23
Pend invalid brats test (external db tls)

aramprice:
These tests have been invalid since their creation. From the beginning they have passed certificate material in the director deployment manifest, however the director itself has (always?) expected to be passed _paths_ to the certificates.

Original tests here:
https://github.com/cloudfoundry/bosh/commit/f97c5565dba9dfa9445a08ee2d7e495b1d5efccd#diff-ab90421945d114536b0c024f3cbaee7f4566c79a592ba9d46b395f0771745b42R293-R306

and

Director config here:
https://github.com/cloudfoundry/bosh/blob/c095b06c91f14ef51b5a8bcea20c3ec3651590b8/src/bosh-director/lib/bosh/director/config.rb#L303-L318

### DIFF
--- a/ci/brats/gcp_mysql.tf
+++ b/ci/brats/gcp_mysql.tf
@@ -1,5 +1,5 @@
 resource "google_sql_database_instance" "mysql-master" {
-  database_version    = "MYSQL_5_7"
+  database_version    = "MYSQL_8_0"
   region              = "us-central1"
   deletion_protection = false
   settings {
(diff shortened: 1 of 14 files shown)


aramprice:
one-off brats run here: https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-director/jobs/pend-invalid-brats-delete/builds/1

selzoc:
This branch seems to have a lot more than just pending the brats test - is that intentional?

aramprice:
Yeah, I suppose that could be split out

The first set of commits were exploratory for understanding the problem (and adding some facilities for debugging this sort of thing in the future).

I am/was hesitant to simply pend the specs without supporting info/evidence.

That being said there are 3-4 commits which are purely cleanup that could likely be pushed separately.

¯\_(ツ)_/¯

PR #2555, opened 2024-08-26
Add CPI method for native disk update

s4heid:
### What is this change about?

Certain cloud providers offer the capability to modify disk attributes (e.g. IOPS and throughput) directly, eliminating the need to create a new disk. To accommodate this, this change introduces a new CPI method called `update_disk`. Previously, adjusting disk properties required creating a new disk and transferring the data, a process that could be time-consuming for larger disks. Should the CPI not support this new method, BOSH will revert to the traditional approach of disk creation.

### Please provide contextual information.

- Feature Request: #2496 
- CPI Change: https://github.com/cloudfoundry/bosh-cpi-ruby/pull/11
- Docs: https://github.com/cloudfoundry/docs-bosh/pull/850

### What tests have you run against this PR?

- Unit tests
- Manual testing with a CPI that implements `update_disk`

I couldn't get the integration tests running on my system (yet).

### How should this change be described in bosh release notes?

Enables direct update of disk properties through the Cloud Provider if supported, significantly speeding up large disk updates.

### Does this PR introduce a breaking change?

No, this change is compatible with older versions.

### Tag your pair, your PM, and/or team!

@MSSedusch

PR #2570, opened 2024-09-27
Revert "Add CPI method for native disk update"

aramprice:
Reverts cloudfoundry/bosh#2555


aramprice:
There is something odd about the spec here https://github.com/cloudfoundry/bosh/blob/main/src/bosh-director/spec/unit/cloud_config/cloud_configs_consolidator_spec.rb#L85-L90

It seems to be relying on `let`'s lazy creation by only referencing `cc_model_1` and `cc_model_2` in the setup but passing in hard-coded Database ID's corresponding to `cc_model_1`, `cc_model_2`, and  `cc_model_3`.

PR #2573, opened 2024-10-04
Gemfile and *.gemspec files are more honest

aramprice:
- move runtime dependencies into respective gemspec files
- remove `:production` and `:bats` gem groups
- ensure Gemfile has required `:development` / `:test` dependencies declared
- move version locking to Gemfile so that it is controlled in one place


aramprice:
Successfully ran `rake fly:integration`
- https://bosh.ci.cloudfoundry.org/builds/162822645

PR #2577, opened 2024-10-11
Collapse internal `bosh-core` gem into `bosh-dev` gem

aramprice:
Note: the pipeline will need to be reconfigured once this is merged to remove the `unit-core` job which is no longer needed (these specs are now part of `unit-dev`).


aramprice:
`rake fly:integration` => https://bosh.ci.cloudfoundry.org/builds/164486992

ystros (approved):

PR #2580, opened 2024-10-23
Upgrade `sequel` from `5.29` => `5.85`

aramprice:
Because sequel now handles migrations ~better~ differently with sqlite some older migrations had to be modified to accommodate the stricter behavior/expectations of sequel + sqlite. As a consquence in the migrations which were not functioning in sqlite we resorted to a backup, recreate, repopulate strategy since `alter_table` and friends were not behaving as expected with sqlite.

Sqlite was added to the Github Actions test matrix for the bosh-director internal gem.

Additionally we made the rspec diff size unlimited.


aramprice:
Closing this in favor of resolving migration issues via creating a roll-up migration https://github.com/cloudfoundry/bosh/pull/2583

PR #2582, opened 2024-11-08
Use blobs in integration specs instead of bucket artifacts

ystros:

We are making an effort to consolidate or eliminate buckets from various pipelines across the ecosystem. Currently the integration specs download several artifacts from S3 buckets. Some of the artifacts being downloaded did not represent the current version used in the Director, such as the NATS server.

Instead of downloading these artifacts from out-of-date buckets, the binaries are extracted from the Director blobs directly. The config server binary must now be provided via environment variables because it is not a blob, and is only used in testing.

One notable change is that since only Linux versions of binaries are included in the current blobs, it is not possible to run the integration specs on a Mac. The previous code appears to have allows running on a Mac, though I personally could not get it to work.

*  ✅  Integration specs via `bundle exec rake fly:integration`


src/spec/integration_support/config_server_service.rb
```diff
@@ -37,10 +36,11 @@ def initialize(port_provider, base_log_path, logger, test_env_number)
     end
 
     def self.install
+      binary_file_path = ENV.fetch('CONFIG_SERVER_BINARY')
```
aramprice, line 39: It might be helpful to pull usages of `ENV` up to `Sandbox` or higher so that it is easier to reason about.

aramprice (approved):
🎉

ystros:
Latest force push rebases against `main` branch again, and rewords the `nats-server` related commit (previously tried to run nats-sync locally but had to abandon that approach, but the commit still referenced running it)

PR #2584, opened 2024-11-13
Remove postgres-10 from bosh release

aramprice:
Items of note
- worker_ctl.erb had a reference to `postgres-10/bin` which was changed to `postgres-13/bin`; we suspect this may not be needed in $PATH
- a warning about upgrading `postgres-13` was commented out in the prestart script since we currently still support upgrades from v13


jpalermo:
Default [was changed](https://github.com/cloudfoundry/bosh-deployment/commit/a6b28a0ce0b4ccd3a918aee10f33c0d10ea1c1f9) from 10 only 2 years ago, so this will prevent upgrades from any envs at that point.

aramprice:
Removing this package will save ~10-15 minutes of compile time.

xtreme-nitin-ravindran (approved):

jpalermo (commented):
Could you change the version needed to upgrade. Looks like we shipped another one.

ystros:
Updated the version in error messages for the last Postgres-10-compatible Director.

PR #2585, opened 2024-11-21
Nats syc uses net http

aramprice:
Integration one-off build
- ✅ https://bosh.ci.cloudfoundry.org/builds/174950558

### DIFF
--- a/src/bosh-nats-sync/lib/nats_sync/users_sync.rb
+++ b/src/bosh-nats-sync/lib/nats_sync/users_sync.rb
@@ -1,10 +1,11 @@
-require 'rest-client'
 require 'base64'
 require 'nats_sync/nats_auth_config'
 require 'open3'
 
 module NATSSync
   class UsersSync
+    HTTP_SUCCESS = "200"
+
     def initialize(nats_config_file_path, bosh_config, nats_server_executable, nats_server_pid_file)
       @nats_config_file_path = nats_config_file_path
       @bosh_config = bosh_config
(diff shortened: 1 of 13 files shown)


src/bosh-nats-sync/lib/nats_sync/users_sync.rb
```diff
@@ -42,15 +43,13 @@
     end
 
     def self.reload_nats_server_config(nats_server_executable, nats_server_pid_file)
-      output, status = Open3.capture2e("#{nats_server_executable} --signal reload=#{nats_server_pid_file}")
+      nats_command = "#{nats_server_executable} --signal reload=#{nats_server_pid_file}"
```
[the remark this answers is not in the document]
aramprice: This code is relocated, not added. Probably something to address in the future.

src/bosh-nats-sync/lib/nats_sync/users_sync.rb
```diff
@@ -103,27 +107,21 @@ def query_all_running_vms
...
-      @director_info = JSON.parse(body)
+      @director_info = JSON.parse(bosh_api_response_body('/info', auth: false))
     end
 
-    def create_authentication_header
-      NATSSync::AuthProvider.new(info, @bosh_config).auth_header
+    def build_headers(auth: true)
+      if auth
+        auth_header = "#{NATSSync::AuthProvider.new(info, @bosh_config).auth_header}"
+        NATSSync.logger.debug 'auth_header is empty, next REST call could fail' if auth_header.empty?
```
ystros, line 119: This used to have a `auth_header.nil?` check as well, which may be necessary if we do not explicitly include something like ActiveSupport to define `#empty?` on `NilClass`.

aramprice: Using `"#{...auth_header}"` should produce `""` in the case `.auth_header` returns `nil`
> `"#{nil}".empty? => true`

ystros (approved):

PR #2595, opened 2024-12-31
Fixes a bug where a VM ends up in a state where it does not have a "rendered_templates_archive" in the spec json

jpalermo:
Steps to reproduce:
- Deploy a VM
- Redeploy the VM without making any changes (this updates the spec json in the instances table to no longer have the rendered_templates_archive data since no templates were rendered)
- bosh restart vm/0 --no-converge

The restart action won't cause templates to be re-rendered, but will cause an update of the VM. Since rendered_templates_archive is no longer part of the spec json, the agent thinks it should not be starting any jobs and "monit summary" will be empty.

This fix makes two changes.
- When sending rendered templates over the NATS bus to VMs, rather than through the blobstore, it will persist the entry into the "rendered_templates_archive" table. This seems a bit weird since the templates aren't actually in the blobstore, but this odd pattern is already in use. The spec json on the VM always has a "blobstore_id" even though the templates never went through the blobstore. So it's a bit odd, but it is consistent.
- When constructing the instances in the deployment plan from existing instance models, it will check if the database has the existing rendered templates archive and will merge that into the instance plan model. During a normal deploy, this will simply get replaced after the templates are rendered as part of the deployment process. However, this makes the template archive details available for deploys that do not end up rendering templates.

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/instance.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/instance.rb
@@ -148,6 +148,13 @@ module Bosh::Director
         @model = existing_instance_model
         @desired_variable_set = @deployment_model.last_successful_variable_set || @deployment_model.current_variable_set
         @previous_variable_set = existing_instance_model.variable_set
+        templates_archive = existing_instance_model.latest_rendered_templates_archive
+        if templates_archive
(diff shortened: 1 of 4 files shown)


lnguyen (approved):

beyhan:
Good catch! IIRC I've been investigating this but never got until the root cause of it.

PR #2599, opened 2025-02-03
set retry interval to avoid bosh workers sleeping forever in case of error

fmoehler:

We noticed that if bosh workers fail to startup for some reason, they go to sleep forever. This is due to the @retry_interval not being set [here](https://github.com/cloudfoundry/bosh/blob/d1148167076763449bd1cd616e70200657b87353/src/bosh-director/lib/bosh/director/worker.rb#L51).
In our case the workers were not able to reach the database for some time, but instead of retrying they just went to sleep forever. Due to that the worker processes were still reported as running when doing `monit summary `. However, tasks where not picked up anymore, resulting in a long task queue. We could temporarily fix it by stopping and starting the director process:

```

monit stop director
monit start director
```

unit tests

Fix bug leading to forever sleeping workers

no

@DennisAhaus

### DIFF
--- a/src/bosh-director/lib/bosh/director/worker.rb
+++ b/src/bosh-director/lib/bosh/director/worker.rb
@@ -7,6 +7,7 @@ module Bosh
       def initialize(config, index = 0)
         @config = config
         @index = index
+        @retry_interval = 5
       end
 
       def prep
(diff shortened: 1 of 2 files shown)


a-hassanin (approved):
Great finding ... Thanks!

DennisAhaus (approved):

benjaminguttmann-avtq (approved):

ramonskie:
looks like a sane thing to do.
but i am wondering about the underlying issues. the database.

fmoehler:
This was an external hyperscaler database on gcp probably doing some maintenance.

lnguyen (approved):

PR #2603, opened 2025-02-10
Extend bosh tagging for cpi

DennisAhausSAP:

This change provides the default runtime tags to the upload stemcell cpi call. With this a stemcell can be tagged when it is uploaded.

We wanted to have a possibility to tag all relevant and "bound" resources. With tags clean ups of cloud resources is much easier because it is clear which resources belong together and have certain tags how to handle them. In this PR we want to provide the tags for the cpi.

Unit tests

Provides default runtime tags to the cpi when a stemmcell is created.

No

@anshrupani, @fmoehler

### DIFF
--- a/src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
@@ -109,7 +109,16 @@ module Bosh::Director
             upload_suffix = ' (already exists, skipped)' unless needs_upload
             track_and_log("Uploading stemcell #{@name}/#{@version} to the cloud#{cpi_suffix}#{upload_suffix}") do
               if needs_upload
-                stemcell.cid = cloud.create_stemcell(@stemcell_image, @cloud_properties, {})
+                tags = {}
+                runtime_configs = Bosh::Director::Api::RuntimeConfigManager.new.list(1, "default")
+                if !runtime_configs.nil? && !runtime_configs[0].nil?
+                  raw = runtime_configs[0]
+                  if !raw.nil? 
+                    hash = raw.to_hash
+                    tags.merge!(hash["tags"])
+                  end
+                end
+                stemcell.cid = cloud.create_stemcell(@stemcell_image, @cloud_properties, {"tags" => tags})
                 logger.info("Cloud created stemcell#{cpi_suffix}: #{stemcell.cid}")
               else
                 logger.info("Skipping stemcell upload, already exists#{cpi_suffix}")
--- a/src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
+++ b/src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
@@ -109,7 +109,7 @@ module Bosh::Director
             upload_suffix = ' (already exists, skipped)' unless needs_upload
             track_and_log("Uploading stemcell #{@name}/#{@version} to the cloud#{cpi_suffix}#{upload_suffix}") do
               if needs_upload
-                stemcell.cid = cloud.create_stemcell(@stemcell_image, @cloud_properties)
+                stemcell.cid = cloud.create_stemcell(@stemcell_image, @cloud_properties, {})
                 logger.info("Cloud created stemcell#{cpi_suffix}: #{stemcell.cid}")
               else
                 logger.info("Skipping stemcell upload, already exists#{cpi_suffix}")
--- a/src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
+++ b/src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
@@ -52,7 +52,22 @@ describe Bosh::Director::Jobs::UpdateStemcell do
       end
     end
 
+
+      let(:runtime_config_manager) { instance_double(Bosh::Director::Api::RuntimeConfigManager) }
+      let(:config) { instance_double(Bosh::Director::Models::Config) }
+      let(:runtime_config_list) do
+        [config]
+
+    end
+
     before do
+
+      allow(Bosh::Director::Api::RuntimeConfigManager).to receive(:new).and_return(runtime_config_manager)
+      allow(Bosh::Director::Models::Config).to receive(:new).and_return(config)
+      allow(config).to receive(:to_hash).and_return({"tags"=> {"any"=> "value"}})
+      allow(runtime_config_manager).to receive(:list).with(1, 'default').and_return(runtime_config_list)
+
+
       allow(Bosh::Director::Config).to receive(:event_log).and_return(event_log)
       allow(Bosh::Director::Config).to receive(:uuid).and_return('meow-uuid')
       allow(Bosh::Director::Config).to receive(:cloud_options).and_return({'provider' => {'path' => '/path/to/default/cpi'}})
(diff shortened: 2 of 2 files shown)


DennisAhausSAP:
The PR for the cpi changes will follow.

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -109,7 +109,16 @@ def perform
             upload_suffix = ' (already exists, skipped)' unless needs_upload
             track_and_log("Uploading stemcell #{@name}/#{@version} to the cloud#{cpi_suffix}#{upload_suffix}") do
               if needs_upload
-                stemcell.cid = cloud.create_stemcell(@stemcell_image, @cloud_properties)
+                tags = {}
```
fmoehler, line 112: Can we put this part in a separate method?

DennisAhausSAP: You mean like `get_tags()` returning the tags hash?

fmoehler: yeah something like that would be nice I think, what do you think?

DennisAhausSAP: @fmoehler I added a new method. Please check.

a-hassanin (approved):
LGTM

beyhan:
@DennisAhausSAP this is API change for the create_stemcell function. Have you checked on the CPI behaviour? Do we need to implement this for CPIs or they are resilient against this change. The CPI APIs are documented in https://bosh.io/docs/cpi-api-v2-method/create-stemcell/

DennisAhausSAP:
Thanks @beyhan. I tested on another scenario and found an issue. I am going to fix it. Regards Dennis

a-hassanin (changes requested):
It seems like supporting AMI tagging on all CPIs would be a lot of work across 10 repositories. If we are going anyhow with a new CPI version for AWS that seems to be required for IPV6 dual stack support, see here https://github.com/cloudfoundry/community/pull/1077#discussion_r1974821522 I would suggest is that we include this feature with a CPI version 3. What do you think?

Alternatively, of course we could provide an empty hash in all CPIs for the create_stemcell method which does nothing and people can provide this feature slowly at other CPIs as well. If you think that is fine, we could go ahead with this approach 

@beyhan @jpalermo

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
```diff
@@ -146,6 +147,21 @@ def perform
...
+        if !runtime_configs.nil?
+          raw = runtime_configs[0]
+          if !raw.nil? 
+            hash = raw.to_hash
+            if !hash.nil? && !hash["tags"].nil?
+              tags.merge!(hash["tags"])
+            end
+          end
+        end
+        return tags
```
fmoehler, line 162: Return not needed in ruby

src/bosh-director/spec/unit/cloud/external_cpi_response_wrapper_spec.rb
```diff
@@ -137,6 +137,8 @@ def self.it_calls_cpi_method(method, *arguments)
           }
         elsif method == :create_disk
           expected_arguments[1] = { 'type' => '<redacted>' }
+        elsif method == :create_stemcell && cpi_api_version >= 3
+          # expected_arguments[2] = { "tags" => {"any":"value"} }
```
fmoehler, line 141: Should this be commented?

src/bosh-director/lib/bosh/director/jobs/update_stemcell.rb
DennisAhausSAP: You are right. But this makes the code cleaner and easier to read.

src/bosh-director/spec/unit/cloud/external_cpi_response_wrapper_spec.rb
DennisAhausSAP: Done

a-hassanin (approved):
LGTM

src/bosh-director/spec/unit/cloud/external_cpi_response_wrapper_spec.rb
```diff
@@ -137,6 +137,7 @@ def self.it_calls_cpi_method(method, *arguments)
           }
         elsif method == :create_disk
           expected_arguments[1] = { 'type' => '<redacted>' }
+        elsif method == :create_stemcell && cpi_api_version >= 3
```
fmoehler, line 140: Is this needed or a leftover?

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
```diff
@@ -52,7 +52,22 @@
...
+      let(:runtime_config_list) do
+        [config]
+
+    end
+
     before do
+
+      allow(Bosh::Director::Api::RuntimeConfigManager).to receive(:new).and_return(runtime_config_manager)
+      allow(Bosh::Director::Models::Config).to receive(:new).and_return(config)
+      allow(config).to receive(:to_hash).and_return({"tags"=> {"any"=> "value"}})
```
fmoehler, line 67: Minor point you can also dismiss: Maybe instead of having "any" we could write "key"?

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
```diff
@@ -78,7 +93,7 @@
       context 'uploading a local stemcell' do
         it 'should upload a local stemcell' do
           expect(cloud).to receive(:info).and_return('stemcell_formats' => ['dummy'])
-          expect(cloud).to receive(:create_stemcell).with(anything, { 'ram' => '2gb' }) do |image, _|
+          expect(cloud).to receive(:create_stemcell).with(anything, { 'ram' => '2gb' }, {"tags"=>{"any"=>"value"}}) do |image, _|
```
fmoehler, line 96: I think here are some spaces missing

a-hassanin (changes requested):
Please fix failing tests.

src/bosh-director/spec/unit/cloud/external_cpi_response_wrapper_spec.rb
DennisAhausSAP: Done!

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
DennisAhausSAP: I like any:value more :-).

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
```diff
@@ -319,9 +333,9 @@
           expect(cloud3).to receive(:info).and_return('stemcell_formats' => ['dummy'])
 
           expect(cloud_factory).to receive(:all_names).exactly(3).times.and_return(%w[cloud1 cloud2 cloud3])
-          expect(cloud_factory).to receive(:get).with('cloud1').and_return(cloud1)
-          expect(cloud_factory).to receive(:get).with('cloud2').and_return(cloud2)
-          expect(cloud_factory).to receive(:get).with('cloud3').and_return(cloud3)
+          expect(cloud_factory).to receive(:get).with('cloud1').and_return(Bosh::Clouds::ExternalCpiResponseWrapper.new(cloud1, 1))
```
a-hassanin, line 336: From what I see that cloud1, cloud2, cloud3 are actually instances of ExternalCpi
https://github.com/cloudfoundry/bosh/blob/1f587eb1e2294d023b8448f0f7bd4ee05a3d1d71/src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb#L73

a-hassanin (changes requested):

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
a-hassanin: Ah sorry, my mistake. Looks good!

a-hassanin (approved):
LGTM

src/bosh-director/spec/unit/jobs/update_stemcell_spec.rb
a-hassanin: But generally speaking instance_double is better than actually initializing an instance, will do that and merge the PR right

a-hassanin:
Ready to merge

PR #2607, opened 2025-03-12
Inline some of bosh_common into `bosh-director`

aramprice:


jpalermo (approved):

src/bosh-director/lib/bosh/core_ext/kernel.rb
```diff
@@ -0,0 +1,9 @@
+module Kernel
+  def with_thread_name(name)
+    old_name = Thread.current[:name]
+    Thread.current[:name] = name
+    yield
+  ensure
+    Thread.current[:name] = old_name
+  end
```
ystros, line 8: This is used within the CPIs, so we can't explicitly move it to the `bosh-director` gem. E.g.:
https://github.com/cloudfoundry/bosh-aws-cpi-release/blob/b2d5ace97886ca8533d7f93a323d863056712bc2/src/bosh_aws_cpi/lib/cloud/aws/cloud_v1.rb#L108

ystros (changes requested):
Looks good overall, though it seems like the CPIs are using the `with_thread_name` method / including the thread formatter file. Those would need to be updated before moving these.

src/bosh-director/lib/bosh/core_ext/kernel.rb
aramprice: This will live on in a future extracted `bosh-common` gem, likely as (something like):

```ruby
Bosh::Common::Thread.with_name(name) do
...
end
```

... however we don't currently have any evidence that any CPI code actually accesses `Thread.current[:name]`.

jpalermo (approved):

fmoehler (approved):

PR #2611, opened 2025-04-15
Introduce prefix allocation

fmoehler:


peanball:
Hi @aramprice, I just watched the recording of the foundational infra meeting from April 17th. 

To clarify a bit:
Diego is currently using Silk as overlay network that is independent from the IP address(es) of the Diego VM itself.

The idea for IPv6 is to use as much native network routing as possible, i.e. not use an overlay network for IPv6. This means that the IP addresses that would be assigned to the containers will be delegated from an IPv6 prefix that is assigned to the Diego VM.

Your question about whether those IP addresses would move in case of an evacuation: no. These IP addresses are not "sticky" to the app or app instance and would not move.

The goal is to make the networking setup for Diego simpler by using "native" IPv6 addressing. Traffic aimed at a particular container will reach the Diego VM via its CIDR range, and then the Diego VM's kernel can forward the traffic to the container's virtual NIC.

Please also note that the networks are supposed to be dual stack, not pure IPv6. So you would want to assign multiple (at least one IPv4 and one IPv6) networks to the same VM, as @fmoehler mentioned in the call already.

The "prefix" parameter is also the size of the prefix to delegate to _each_ VM from a larger range. Have a look at the [discussion while creating the RFC](https://github.com/cloudfoundry/community/pull/1077#discussion_r1991435740) for a more extensive example.

aramprice:
Hi @peanball,

Thanks for that additional context - the info and the RFC were helpful.

In thinking about the overall changes to Bosh a few goals or principles (maybe too strong a word) came to mind:
- Make conditional behavior / IP-version specific code and settings as limited as possible
  - I really like that `prefix:` is used and not `ipv6_prefix:`
  - Should an IPv4 network allow/require a `prefix:` (one that selects a single IP) for consistency?
- Reduce the back and forth conversion / formatting where possible (`format_ip()`, `ip.to_i` and similar)
  - Ideally a single `IpAddrOrCidr` (I apologize for the name) or `IpAddress` representation would be used, and conversions would be handled only where need _by the class itself_, ex: `#to_db_value` - for persistence, or `#to_s` - for logs
- Avoid assumption in the code that bosh will always be dual-stack
  - I realize the system will start out dual-stack but this may not be forever

This isn't to imply disagreement about the current state of things, only to captures my thoughts at the moment.

fmoehler:
@aramprice thanks for your feedback!

I just want to elaborate a little on our current idea, but of course this is open for ideas. Regarding your points:

1. We will introduce a "prefix" property in the subnet section of the network (Maybe we can also put it to the network section itself, so that it applies automatically to all subnets?). This property will be considered for ipv4 and ipv6 networks. However if the property is not defined (e.g. for existing networks), the director will consider the "prefix" as /32 (ipv4) or /128 (ipv6) and maintain this accordingly in the database.
2. Yes we will try to get rid of as many conversion as possible. Sometimes it can be handy to have the ips in integer representation, but most of the time we will pass them as IpAddrOrCidr Objects
3. Agree with that. I did not test this (yet), but it might not even be specific for a dual stack setup. Probably it would also need changes in the bosh agent.

peanball:
@fmoehler I have a general comment on the term `prefix` in the config. It's a bit terse and does not convey the meaning very clearly.

In the end this is the prefix (size) delegated to each instance attached to this subnet, right? Can we somehow express this better? My suggestion would be 
* `delegate_prefix`
* `delegate_prefix_size`
* `delegate_cidr_size`

alternatively `s/delegate/assign/g`.

DennisAhausSAP:
One thing regarding the display of the ip addresses. The bosh-cli is showing the ips as they come from the directors API. As @fmoehler mentioned the prefixes are stored in the db with their prefixes. This also means that the bosh-cli will show the ip addresses with their prefix. Because all the docs are showing examples with ipv4 only this would also mean that everybody has to be prepared that with this change the cli could possibly show the ips (i.e. with command bosh ... vms) in format "ip<v4\v6>/prefix".

fmoehler:
> One thing regarding the display of the ip addresses. The bosh-cli is showing the ips as they come from the directors API. As @fmoehler mentioned the prefixes are stored in the db with their prefixes. This also means that the bosh-cli will show the ip addresses with their prefix. Because all the docs are showing examples with ipv4 only this would also mean that everybody has to be prepared that with this change the cli could possibly show the ips (i.e. with command bosh ... vms) in format "ip<v4\v6>/prefix".

So far we ensured not to change the API and only return the ip (without the prefix), please refer: https://github.com/cloudfoundry/bosh/pull/2611/files#diff-c9cf4222f4ba33b6b3f35d7f32bfbb2c03a91470e18672bd3dda5468c38a0808R22 so it should not change the way ip addresses are displayed.

src/bosh-director/lib/bosh/director/models/vm.rb
```diff
@@ -19,6 +19,16 @@ def network_spec=(spec)
     end
 
     def ips
+      ips_cidr.map do | cidr_ip |
+       if ( cidr_ip.include?(':') && cidr_ip.include?('/128') ) || ( cidr_ip.include?('.')  && cidr_ip.include?('/32') )
+        cidr_ip.split('/')[0]
+       else
+        cidr_ip
+       end
+      end
```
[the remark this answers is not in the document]
fmoehler: This is supposed to ensure backwards compatibility in the cli and other consumers of the api, which would break if suddenly "1.1.1.1/32" is returned and not "1.1.1.1". Furthermore for single ip addresses the prefix information is not important, but for prefix ip addresses it actually is important.

fmoehler:
@beyhan yes this is ready for review. I have checked the failing tests, but they are not failing locally for me. I will check those again, but maybe also someone has a hint  on why they are going wrong.

beyhan:
@fmoehler try to run the unit tests with the same version of Ruby `3.3.8` which is used by the check. If I do that I see that the test fails locally.

src/bosh-director/spec/unit/bosh/director/core/templates/source_erb_spec.rb
```diff
@@ -36,7 +36,7 @@ module Bosh::Director::Core::Templates
           allow(logger).to receive(:debug)
         end
 
-        let(:original_error) { "undefined method `no_method' for nil" }
+        let(:original_error) { "undefined method `no_method' for nil:NilClass" }
```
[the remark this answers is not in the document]
fmoehler: Thanks @beyhan , you are right.  I was using ruby 3.2.2, for which this change was needed. Running it with 3.3.8 fails. I reverted this change.

fmoehler:
BTW. During our tests we found a bug that is related to the vast number of ip addresses in the ipv6 address space. The issue originates from the "reserved" section in the cloud config. As the code in some places iterates over all the reserved ip addresses it can cause the director to not become unresponsive if the ranges becoming to big (in our example we reserved a /80 range which are 281,474,976,710,656 ip addresses). However, this issue is not related to this change but has been present before. We will try to tackle this as well in a separate commit in this PR, but this can still already be reviewed in the current state.

src/bosh-director/lib/bosh/director/deployment_plan/dynamic_network.rb
```diff
@@ -21,16 +22,24 @@ def self.parse(network_spec, availability_zones, logger)
           subnets = network_spec['subnets'].map do |subnet_properties|
             name_servers = name_server_parser.parse(subnet_properties['name'], subnet_properties)
             cloud_properties = safe_property(subnet_properties, 'cloud_properties', class: Hash, default: {})
+            prefix = safe_property(subnet_properties, 'prefix', class: Integer, default: nil)
+            if prefix.nil?
+              prefix = 32 # we need to set the ipv4 default value (dynamic networks only support ipv4)
```
aramprice, line 27: It seems odd to be setting `prefix = 32` here `if prefix.nil?` when we are defaulting it to `nil` above on line 25...

Should we be setting `default: 32` and then using the following guard:

```ruby
raise NetworkInvalidProperty, "Prefix property is not supported for dynamic networks." unless prefix == 32
```

src/bosh-director/lib/bosh/director/deployment_plan/dynamic_network.rb
```diff
@@ -21,16 +22,24 @@ def self.parse(network_spec, availability_zones, logger)
...
+            else
+              raise NetworkInvalidProperty, "Prefix property is not supported for dynamic networks."
+            end
             subnet_availability_zones = parse_availability_zones(subnet_properties, availability_zones, name)
-            DynamicNetworkSubnet.new(name_servers, cloud_properties, subnet_availability_zones)
+            DynamicNetworkSubnet.new(name_servers, cloud_properties, subnet_availability_zones, prefix)
           end
         else
           cloud_properties = safe_property(network_spec, 'cloud_properties', class: Hash, default: {})
+          prefix = 32 # we need to set the ipv4 default value (dynamic networks only support ipv4)
```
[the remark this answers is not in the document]
aramprice: Could we use a constant like: `<SomeRubyClass::>IPV4_PREFIX_SIZE = 32` here

src/bosh-director/lib/bosh/director/deployment_plan/dynamic_network.rb
aramprice: Could we use a constant like: `<SomeRubyClass::>IPV4_PREFIX_SIZE = 32` here

src/bosh-director/lib/bosh/director/deployment_plan/manual_network_subnet.rb
```diff
@@ -75,6 +83,28 @@ def self.parse(network_name, subnet_spec, availability_zones, managed = false)
 
             static_ips.add(ip)
           end
+
+          if prefix.nil?
+            if range.ipv6?
+              prefix = "128"
+            else
+              prefix = "32"
```
aramprice, line 91: Constants would be better here

src/bosh-director/lib/bosh/director/ip_addr_or_cidr.rb
```diff
@@ -3,26 +3,68 @@
...
+        end
+      end
+
       def count
         (@ipaddr.to_range.last.to_i - @ipaddr.to_range.first.to_i) + 1
       end
 
       def to_cidr_s
-        "#{@ipaddr}/#{@ipaddr.prefix}"
+        "#{@ipaddr.to_string}/#{@ipaddr.prefix}"
```
aramprice, line 46: `#to_string` isn't needed in an interpolation context -> `"...#{}..."`

src/bosh-director/spec/unit/bosh/director/api/controllers/deployments_controller_spec.rb
```diff
@@ -1453,7 +1453,7 @@ def manifest_with_errand(deployment_name='errand')
                   ip_addresses_params = {
                     'instance_id' => instance.id,
                     'task_id' => i.to_s,
-                    'address_str' => ip_to_i("1.2.#{i}.#{j}").to_s,
+                    'address_str' => ("1.2.#{i}.#{j}/32").to_s,
```
aramprice, line 1456: I don't think `#to_s` is needed any longer here.

src/bosh-director/lib/bosh/director/ip_addr_or_cidr.rb
```diff
@@ -3,26 +3,68 @@
...
+          yield base_address_int
+          base_address_int += step_size
+        end
+      end
+
       def count
         (@ipaddr.to_range.last.to_i - @ipaddr.to_range.first.to_i) + 1
       end
 
       def to_cidr_s
```
aramprice, line 45: We seem to have replaced `#to_s` with `#to_cidr_s` in most places... should we make `#to_cidr_s` the implementation of `#to_s` and add a method to this class like [`Bosh::Director::Models::IpAddress#base_address`](https://github.com/cloudfoundry/bosh/pull/2611/files#diff-e81c65a9da1d988d8b3ffcbec2b04842d6aecdb60d46a05de0e8397cf4f56a47R34-R36) below?

aramprice (changes requested):

src/bosh-director/lib/bosh/director/ip_addr_or_cidr.rb
fmoehler: I wanted to get the ip address in canonical form. This is what the .to_string method does. In contrast the .to_s method would return the whole ip address
https://ruby-doc.org/stdlib-2.5.1/libdoc/ipaddr/rdoc/IPAddr.html#method-i-to_string

src/bosh-director/lib/bosh/director/ip_addr_or_cidr.rb
fmoehler: You mean we overwrite the [ipaddr to_s method](https://ruby-doc.org/stdlib-2.5.1/libdoc/ipaddr/rdoc/IPAddr.html#method-i-to_s)? We can do that. The "to_cidr_s" does something a little different. Not sure if maybe in the future wanted to use the default "to_s" method. But I am certainly open to this refactoring. Please just confirm that I understood it correctly :D

aramprice: I wasn't suggesting rewriting `IPAddr#to_s`... my thought was for `IpAddrOrCidr#to_cidr_s` to be renamed to `IpAddrOrCidr#to_s` (or `#to_string`) since for the class `IpAddrOrCidr` we seem to almost always call `#to_cidr_s` in "string-ish" contexts, and only rarely call it's `#to_s` method.

Based on usage, it make me thing that the "string-ish" version of `IpAddrOrCidr` should actually be what's currently returned by `#to_cider_s`. My reference to `Bosh::Director::Models::IpAddress#base_address` was to note that this class has an existing name for a "string-ish representation without address-prefix" method.

---

Separately there is (in my mind at least) a question about whether `IpAddrOrCidr` should be combined with `Bosh::Director::Models::IpAddress`, rather than having two separate objects.

fmoehler: Thanks for the clarification @aramprice . I checked your proposal again and also started to implement the changes. However I came across the following issue:

In some cases we need to get just the ip address without the prefix. E.g. [in the network_settings of the manual network](https://github.com/cloudfoundry/bosh/blob/b3fab8ae93f0522dc072103c725db694c6a21ab9/src/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb#L57), but in other places it is important to get the ip and the prefix, e.g. [when we save the ip address to the database](https://github.com/cloudfoundry/bosh/pull/2611/files#diff-c920100cbc21d0c9dcba1cd3a2707ac7789867ae9570983471ecdbb96b9f150aR108) to not lose the prefix information. What we could think about refactoring the to_cidr_s method to be something like this:

```
def to_cidr_s --> rename to to_s
  if ipaddr.prefix == single_ip_prefix 
   ip_addr.to_s (will get rid of the prefix, by calling the to_s method of ipaddr)
  else
    "#{@ipaddr}/#{@ipaddr.prefix}"
  end
end
```
However the to_cidr_s method was not implemented by this PR, so I would have to check if this might destroy existing behavior. On a quick look it would not.
What do you think about this approach?

Regarding the `Bosh::Director::Models::IpAddress#base_address` method in IPAddress, I think I can get rid of it entirely as it is not needed anymore. 

And regarding your last point to merge IpAddress and IpAddrOrCidr: The IpAddress class is the database model, which basically means (a least that is how I understand it) every instance is an entry inside the database. However often in the coding it is necessary to just have an ipaddress without accessing the database. E.g. when calculating the next available ip etc. So I think it makes sense to keep these separated.

aramprice: > And regarding your last point to merge IpAddress and IpAddrOrCidr: The IpAddress class is the database model, which basically means (a least that is how I understand it) every instance is an entry inside the database. However often in the coding it is necessary to just have an ipaddress without accessing the database. E.g. when calculating the next available ip etc. So I think it makes sense to keep these separated.

^ yeah, I realized this was probably the case after I posted. I agree that they should stay separate.

> Regarding the Bosh::Director::Models::IpAddress#base_address method in IPAddress, I think I can get rid of it entirely as it is not needed anymore.

^ 👍🏻

With respect to changing `#to_cidr_s` -> `#to_s` (or `to_string`), my thought was that  nother method would be added to `IpAddrOrCidr` which provides access to the bare IP, something like the following:

```ruby
module Bosh
  module Director
    class IpAddrOrCidr
      include Comparable

      # remove :to_string from "delegate ... to: :@ipaddr"
      delegate :==, :include?, :ipv4?, :ipv6?, :netmask, :mask, :to_i, :to_range :prefix, :succ, :<=>, to: :@ipaddr
      alias :to_s :to_string

# ... snip

      # was to_cider_s
      # => making this the default "string representation" means that all of the interpolated
      #       context like "#{ip_or_cidr.to_cidr_s}" can become "#{ip_or_cidr}" because
      #       `#to_s` will be called automatically, and the "address/prefix" notation _is_
      #       the most complete string representation of an `IpAddrOrCidr` class
      def to_string 
        "#{@ipaddr.to_string}/#{@ipaddr.prefix}"
      end

      def base_addr # new method to use in the few places that called IpAddrOrCidr#to_s
        @ipaddr.to_string
      end

# ... snip

    end
  end
end
```

src/bosh-director/lib/bosh/director/deployment_plan/vip_network.rb
```diff
@@ -15,8 +15,14 @@ def self.parse(network_spec, availability_zones, logger)
           DeploymentPlan::VipNetworkSubnet.parse(subnet_spec, name, availability_zones)
         end
 
+        unless subnets.empty?
+          prefix = subnets.first.prefix
+        else
+          prefix = "32"
```
aramprice, line 21: Can if `Network::IPV4_DEFAULT_PREFIX_SIZE` be used here?

src/bosh-director/lib/bosh/director/models/vm.rb
```diff
@@ -19,6 +19,16 @@ def network_spec=(spec)
     end
 
     def ips
+      ips_cidr.map do | cidr_ip |
+       if ( cidr_ip.include?(':') && cidr_ip.include?('/128') ) || ( cidr_ip.include?('.')  && cidr_ip.include?('/32') )
```
aramprice, line 23: Can `Network::IPV6_DEFAULT_PREFIX_SIZE` and `Network::IPV4_DEFAULT_PREFIX_SIZE` be used here to avoid magic numbers?

src/bosh-director/lib/bosh/director/ip_addr_or_cidr.rb
fmoehler: I addresses this in: [5e73f23](https://github.com/cloudfoundry/bosh/pull/2611/commits/5e73f233b941fb39595b1906b0544e9c7fc9d1ea)

src/bosh-director/lib/bosh/director/deployment_plan/vip_network_subnet.rb
```diff
@@ -4,28 +4,34 @@ class VipNetworkSubnet < Subnet
...
 
-      attr_reader :static_ips, :availability_zone_names
+      attr_reader :static_ips, :availability_zone_names, :prefix
 
       def self.parse(subnet_spec, network_name, azs)
         static_ips = Set.new
 
+        prefix = safe_property(subnet_spec, 'prefix', optional: true)
+        if prefix.nil?
+          prefix = "32"
```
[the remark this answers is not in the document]
fmoehler: Fixed with https://github.com/cloudfoundry/bosh/pull/2611/commits/5d75ab50ee1ebe60c743e9e3804baeeedcc18e69

src/bosh-director/lib/bosh/director/ip_util.rb
```diff
@@ -32,30 +34,83 @@ def ip_address?(ip)
       return false
     end
 
+    def ip_in_array?(ip_to_check, ip_objects_array)
+      ip_to_check = Bosh::Director::IpAddrOrCidr.new(ip_to_check)
```
aramprice, line 38: Should this also use `to_ipaddr(ip)`?

src/bosh-director/lib/bosh/director/ip_util.rb
```diff
@@ -18,10 +18,12 @@ def to_ipaddr(ip)
...
-    # @param [Integer] ip Integer IP representation
-    # @return [String] Human-readable IP representation
-    def format_ip(ip)
-      to_ipaddr(ip).to_s
+    def base_addr(ip)
+      Bosh::Director::IpAddrOrCidr.new(ip).base_addr
+    end
+
+    def format_cidr_ip(ip)
+      ip.to_s
```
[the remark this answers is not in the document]
fmoehler: Got rid of this entirely. But reused to_ipaddr(ip) in many other places

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
```diff
@@ -69,64 +71,96 @@ def allocate_vip_ip(reservation, subnet)
...
-      unless subnet.range == ip_address || subnet.range.include?(ip_address)
-        raise NoMoreIPsAvailableAndStopRetrying
+      if subnet.range.ipv6?
+        addresses_we_cant_allocate.delete_if { |ipaddr| ipaddr.ipv4? }
+      else
+        addresses_we_cant_allocate.delete_if { |ipaddr| ipaddr.ipv6? }
       end
 
-      save_ip(ip_address, reservation, false)
+      prefix = subnet.prefix
```
aramprice, line 99: nit: maybe inline this?

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
```diff
@@ -22,7 +23,7 @@ def delete(ip)
     end
 
     def add(reservation)
-      ip_or_cidr = Bosh::Director::IpAddrOrCidr.new(reservation.ip)
+      ip_or_cidr = reservation.ip
```
aramprice, line 26: nit: maybe inline?

src/bosh-director/lib/bosh/director/models/ip_address.rb
```diff
@@ -1,55 +1,60 @@
...
+      end
 
-    def address
-      unless address_str.match?(/\A\d+\z/)
-        info_display = ''
-        begin
-          info_display = info
-        rescue StandardError
-          info_display = 'missing_info'
+      def formatted_ip
```
aramprice, line 33: Is `IpAddress#formatted_ip` still used?

src/bosh-director/lib/bosh/director/models/ip_address.rb
```diff
@@ -1,55 +1,60 @@
...
-      Bosh::Director::IpAddrOrCidr.new(address).to_s
-    end
+      def before_create
+        self.created_at ||= Time.now
+      end
 
-    def type
-      static ? 'static' : 'dynamic'
-    end
+      def info
```
aramprice, line 25: This method appears unused.. if so maybe inline into `#to_s`?

aramprice (dismissed):
Looks good. I had a few comments about small things which could be addressed now, or later (or never).

Thanks!

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
Ivaylogi98: refactored in [8e88f6e6183b4cb44730771c208d265e7faa8ab6](https://github.com/cloudfoundry/bosh/pull/2611/commits/8e88f6e6183b4cb44730771c208d265e7faa8ab6)

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
Ivaylogi98: decided to leave as is to have it more descriptive

fmoehler:
Rebased on 282.0.7

fmoehler:
Rebased to 282.0.8

src/bosh-director/lib/bosh/director/models/vm.rb
```diff
@@ -19,6 +19,16 @@ def network_spec=(spec)
...
 
     def ips
+      ips_cidr.map do | cidr_ip |
+       if ( cidr_ip.include?(':') && cidr_ip.include?("/#{Bosh::Director::DeploymentPlan::Network::IPV6_DEFAULT_PREFIX_SIZE}") ) || ( cidr_ip.include?('.')  && cidr_ip.include?("/#{Bosh::Director::DeploymentPlan::Network::IPV4_DEFAULT_PREFIX_SIZE}") )
+        cidr_ip.split('/')[0]
+       else
+        cidr_ip
+       end
+      end
+    end
```
neddp, line 29:
```suggestion
      ips_cidr.map do |cidr_ip|
        if ( cidr_ip.include?(':') && cidr_ip.include?("/#{Bosh::Director::DeploymentPlan::Network::IPV6_DEFAULT_PREFIX_SIZE}") ) || ( cidr_ip.include?('.')  && cidr_ip.include?("/#{Bosh::Director::DeploymentPlan::Network::IPV4_DEFAULT_PREFIX_SIZE}") )
          cidr_ip.split('/')[0]
        else
          cidr_ip
        end
      end
    end
```
Indentation.

src/bosh-director/lib/bosh/director/models/vm.rb
```diff
@@ -29,7 +39,15 @@ def manual_or_vip_ips
...
-      network_spec.map { |_, network| network['ip'] }
+      network_spec.map do |_, network|
+        prefix = network['prefix'].to_s
+        if network['ip'].include?(':') && prefix.empty?
+          prefix = Bosh::Director::DeploymentPlan::Network::IPV6_DEFAULT_PREFIX_SIZE
+        elsif network['ip'].include?('.') && prefix.empty?
+          prefix = Bosh::Director::DeploymentPlan::Network::IPV4_DEFAULT_PREFIX_SIZE
+        end
+        "#{network['ip']}/#{prefix}"
+      end
```
neddp, line 50:
```suggestion
      network_spec.map do |_, network|
        ip = network['ip']
        prefix = network['prefix'].to_s
        
        if prefix.empty?
          prefix = ip.include?(':') ? 
            Bosh::Director::DeploymentPlan::Network::IPV6_DEFAULT_PREFIX_SIZE :
            Bosh::Director::DeploymentPlan::Network::IPV4_DEFAULT_PREFIX_SIZE
        end
        
        "#{ip}/#{prefix}"
      end
```
Readability.

src/bosh-director/lib/bosh/director/network_reservation.rb
```diff
@@ -1,11 +1,12 @@
...
   class NetworkReservation
-    attr_reader :ip, :instance_model, :network, :type
+    attr_reader :ip, :instance_model, :network, :type, :nic_group
 
-    def initialize(instance_model, network)
+    def initialize(instance_model, network, nic_group)
       @instance_model = instance_model
       @network = network
       @ip = nil
+      @nic_group = nic_group.to_i if nic_group
```
neddp, line 9:
```suggestion
      @nic_group = nic_group&.to_i
```
More idiomatic.

src/bosh-director/spec/unit/bosh/director/api/controllers/deployments_controller_spec.rb
```diff
@@ -1485,6 +1485,69 @@ def manifest_with_errand(deployment_name='errand')
...
+                  vm = Models::Vm.create(vm_params)
+
+                  if j == 0
+                    instance.active_vm = vm
+                  end
+
+                  ip_addresses_params = {
+                    'instance_id' => instance.id,
+                    'task_id' => i.to_s,
+                    'address_str' => ("1.2.#{i}.#{j*16}/28").to_s,
```
neddp, line 1519:
```suggestion
                    'address_str' => "1.2.#{i}.#{j*16}/28",
```
Lint/RedundantTypeConversion

src/bosh-director/lib/bosh/director/addon/addon.rb
```diff
@@ -4,6 +4,7 @@ module Addon
     RUNTIME_LEVEL = :runtime
 
     class Addon
+      include IpUtil
```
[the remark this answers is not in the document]
fmoehler: I have to check this. I remember I had to include it for the unit tests to work

src/bosh-director/lib/bosh/director/cloudcheck_helper.rb
```diff
@@ -1,5 +1,6 @@
 module Bosh::Director
   module CloudcheckHelper
+    include IpUtil
```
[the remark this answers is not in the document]
fmoehler: removed

src/bosh-director/lib/bosh/director/addon/addon.rb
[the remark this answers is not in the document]
fmoehler: removed

rkoster:
For people who are wondering about the `nic_group`. This is what the robots have found so far:

     ## Flow Path:

     1. Configuration Parse: nic_group is parsed from the deployment manifest in InstanceGroupNetworksParser.parse() at
     src/bosh-director/lib/bosh/director/deployment_plan/instance_group_networks_parser.rb:36
     2. JobNetwork Creation: The parsed value is stored in a JobNetwork object at src/bosh-director/lib/bosh/director/deployment_plan/job_network.rb:11
     3. Network Reservation: When creating network reservations, the nic_group flows from JobNetwork to DesiredNetworkReservation via the network planner at
     src/bosh-director/lib/bosh/director/deployment_plan/network_planner/planner.rb:9,15
     4. Network Settings Generation: In NetworkSettings.to_hash() at src/bosh-director/lib/bosh/director/deployment_plan/network_settings.rb:36, it calls reservation.network.
     network_settings(reservation, ...)
     5. Manual Network Processing: The key step is in ManualNetwork.network_settings() at src/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb:80-83 where:
     nic_group = reservation.nic_group
     if nic_group
       config["nic_group"] = nic_group.to_s
     end

     6. Instance Plan: The network settings hash is generated via InstancePlan.network_settings_hash() at src/bosh-director/lib/bosh/director/deployment_plan/instance_plan.rb:289-290
     7. CPI Call: Finally, in CreateVmStep.perform() at src/bosh-director/lib/bosh/director/deployment_plan/steps/create_vm_step.rb:31, the network settings hash is passed to create()
     method, which then calls cloud.create_vm() at line 150 with the network_settings parameter containing the nic_group field.

     The nic_group field flows from the deployment manifest through the network planning system, gets added to the network configuration hash in the ManualNetwork.network_settings
     method, and is ultimately passed to the CPI's create_vm method as part of the network_settings parameter.
     
     Based on the analysis of https://github.com/cloudfoundry/bosh-aws-cpi-release/pull/181, the nic_group field is used to group multiple networks together onto the same network interface in AWS EC2 instances. Here's how it works:

     Purpose:

     • Groups multiple BOSH networks to share a single AWS network interface
     • Allows multiple IP addresses (IPv4/IPv6) to be assigned to one network interface
     • Enables network configuration where multiple logical networks use the same physical interface

     Key behaviors:

     1. Network grouping: Networks with the same nic_group value are combined into a single NicGroup object (src/bosh_aws_cpi/lib/cloud/aws/network_interface_manager.rb:18-20)
     2. Default behavior: If no nic_group is specified, it defaults to the network name (src/bosh_aws_cpi/lib/cloud/aws/cloud_props.rb:306)
     3. Subnet validation: All networks in the same nic_group must use the same subnet ID (src/bosh_aws_cpi/lib/cloud/aws/nic_group.rb:71-74)
     4. IP assignment: The first IPv4 and IPv6 addresses from networks in the group are assigned to the interface (src/bosh_aws_cpi/lib/cloud/aws/nic_group.rb:82-100)
     5. Interface creation: Each unique nic_group results in one AWS network interface being created (src/bosh_aws_cpi/lib/cloud/aws/network_interface_manager.rb:42-44)

     This allows BOSH deployments to have multiple logical networks (e.g., for different services or VLANs) while using fewer AWS network interfaces, which can be useful for instance
     types with limited network interface capacity.
 
Is grouping of logical networks onto a single network interface a feature that is supported by all cloud providers? Or is it something that is CPI specific? Because if it is cloud specific it should be propagated through cloud_properties.

neddp (approved):

fmoehler:
Rebased to 282.0.9

a-hassanin (approved):
All review comments has been addressed, and it has been running stable for the last 4 weeks on SAP landscapes, with that it has my approval.

PR #2621, opened 2025-08-27
Add 'file' package to Dockerfile dependencies for intergration

ramonskie:
file package is missing since latest update from ubuntu docker image. 
this is needed for some ci tasks within the bosh ecosystem


ramonskie:
https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-apt-resources
it is fixed with https://github.com/cloudfoundry/bosh-apt-resources/blob/f62d5457410ce67f90d445d9ca1bbabd64d1005f/ci/tasks/create-debian-pkg-from-binary.sh#L50

rkoster (approved):

PR #2629, opened 2025-10-02
dont stringify prefix for vip networks

fmoehler:


src/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
```diff
@@ -38,7 +38,7 @@ def managed?
       def initialize(name, subnets, prefix, logger, managed = false)
         super(name, TaggedLogger.new(logger, 'network-configuration'))
         @subnets = subnets
-        @prefix = prefix
+        @prefix = prefix.to_s
```
a-hassanin, line 41: let us remove from the manual_network_subnet constructor the default value nil to be consistent:

https://github.com/cloudfoundry/bosh/blob/7a5b2075726b5741a0e002a6c464edd7c63e7eaa/src/bosh-director/lib/bosh/director/deployment_plan/manual_network_subnet.rb#L132

src/bosh-director/lib/bosh/director/deployment_plan/vip_network.rb
```diff
@@ -35,7 +35,7 @@ def initialize(name, cloud_properties, subnets, prefix, logger)
         super(name, logger)
         @cloud_properties = cloud_properties
         @subnets = subnets
-        @prefix = prefix
+        @prefix = prefix.to_s
```
a-hassanin, line 38: we need a nil check on line 19 before we pass it to the constructor

src/bosh-director/spec/unit/bosh/director/deployment_plan/network_planner/reservation_reconciler_spec.rb
```diff
@@ -210,7 +210,7 @@ module Bosh::Director::DeploymentPlan
         end
 
         context 'when the existing ip is part of a vip static network but the desired network is a global vip network' do
-          let(:network) { VipNetwork.new('static-vip-network', nil, [], '32', per_spec_logger) }
+          let(:network) { VipNetwork.new('static-vip-network', nil, [], 32, per_spec_logger) }
```
a-hassanin, line 213: this change should not be necessary, we always pass a string, also for the other networks.

a-hassanin (changes requested):
Please check

src/bosh-director/lib/bosh/director/deployment_plan/manual_network.rb
a-hassanin: Is the constructor used some where else ? I see here the value is always filled:

https://github.com/cloudfoundry/bosh/blob/7a5b2075726b5741a0e002a6c464edd7c63e7eaa/src/bosh-director/lib/bosh/director/deployment_plan/manual_network_subnet.rb#L88

aramprice (approved):

PR #2633, opened 2025-10-20
adapt create_vm and attach_disk call for new cpi version 3

fmoehler:

Bosh director should handle all CPI API Versions correctly.

This will solve the failing BATS for the aws cpi: https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-aws-cpi/jobs/bats

The AWS CPI API Version was recently [bumped to version 3](https://github.com/cloudfoundry/bosh-aws-cpi-release/commit/1d03941083988d1cb61c22419f3ffb9c817047dc). This leads to issues when the director tries to create a vm because the response is cleared unexpectedly as previously there is only handling for version 1 and 2. Back then the update to support version 3 was only done for the create_stemcell method. I changed this to be compatible with all versions higher than 2. To me this change appears pretty straightforward, however I could see that this coding was like this already in the past, but was changed to the case statement. This was done [7 years ago](https://github.com/cloudfoundry/bosh/commit/eaeae57f4508812c35ba5bf66f12bd1fd9f8c9a0) so probably the reason why this was changed is not valid anymore? Otherwise we could also change it back to a case statement instead just including version 3.

Unit Tests

Introduce support for CPI API Versions higher than 2 for create_vm and attach_disk.

No

@Ivaylogi98

### DIFF
--- a/src/bosh-director/lib/clouds/external_cpi_response_wrapper.rb
+++ b/src/bosh-director/lib/clouds/external_cpi_response_wrapper.rb
@@ -39,10 +39,9 @@ module Bosh::Clouds
       cpi_response = @cpi.create_vm(*args)
 
       response = []
-      case @cpi_api_version
-      when 2
+      if @cpi_api_version >= 2
         response = cpi_response
-      when 1
+      else
         response << cpi_response
       end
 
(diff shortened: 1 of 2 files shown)


a-hassanin:
I see that the change done [7 years ago](https://github.com/cloudfoundry/bosh/commit/eaeae57f4508812c35ba5bf66f12bd1fd9f8c9a0) was to overcome the error "undefined method cid' for nil:NilClass" when the director does not support the CPI version, but they already have included unit tests for that, and are taking the minimum version supported_cpi_version = [cpi_api_version, MAX_SUPPORTED_CPI_VERSION].min into account. So I think it should be OK to get rid of the case statement.

PR #2635, opened 2025-10-22
Remove unnecessary -z flag from tar extraction commands

mdzhigarov:
## Summary

This PR removes the unnecessary `-z` flag from tar extraction commands throughout the BOSH codebase. Modern tar implementations support automatic compression detection based on file signatures and fallback to filename suffixes.

## Changes Made

- **BOSH Director Components**: Updated release and stemcell extraction in `update_release.rb`, `update_stemcell.rb`, `release_job.rb`, and `job_template_loader.rb`
- **Integration Tests**: Updated nginx and NATS service utilities
- **CI Scripts**: Updated export-release script

## Technical Details

The `-z` flag is only required when reading from pipes or tape drives without random access. For regular file operations, modern tar implementations can automatically detect compression formats using:

1. **File signatures** (magic bytes at the beginning of files)
2. **Filename suffixes** as fallback (e.g., `.tgz`, `.tar.gz`)

## Compatibility

- ✅ Maintains compatibility with modern tar implementations (GNU tar, BSD tar)
- ✅ Preserves `-z` flag where still needed (pipes, test utilities)
- ✅ All existing unit tests pass (they mock the execution layer)
- ✅ No functional changes to error handling or cleanup logic

## Preparation for Future Enhancement

This change is done in preparation for introducing a `--no-compression` flag in the `bosh create-release` command, which will allow creating uncompressed release tarballs that can be processed with the simplified tar commands.

## Testing

- All unit tests continue to pass
- Tests mock the `Exec.sh` calls, so they verify error handling behavior rather than specific command syntax
- No integration test changes required as the functionality remains identical

## References

- [GNU tar documentation on automatic compression detection](https://www.gnu.org/software/tar/manual/html_node/gzip.html)
- Modern tar implementations recognize compression formats automatically when reading archives

### DIFF
--- a/ci/tasks/export-release.sh
+++ b/ci/tasks/export-release.sh
@@ -5,8 +5,8 @@ bosh_repo_dir="$(realpath "$(cd "$(dirname "${BASH_SOURCE[0]}")/../../" && pwd)"
 main() {
   set -eu
 
-  tar -xzf release/*.tgz "$( tar -tzf release/*.tgz | grep 'release.MF' )"
-  tar -xzf stemcell/*.tgz "$( tar -tzf stemcell/*.tgz | grep 'stemcell.MF' )"
+  tar -xf release/*.tgz "$( tar -tzf release/*.tgz | grep 'release.MF' )"
+  tar -xf stemcell/*.tgz "$( tar -tzf stemcell/*.tgz | grep 'stemcell.MF' )"
 
   export STEMCELL_OS=$( grep -E '^operating_system: ' stemcell.MF | awk '{print $2}' | tr -d "\"'" )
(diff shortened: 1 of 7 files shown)


KauzClay (commented):
Just trying to clarify my understanding.

This is related to the `--no-compression` flag change because the `-z` flag on non-compressed files will cause tar to fail?

Any by removing the `z`, since tar can auto-detect, it can safely expand things that were tar-ed without compression?

mdzhigarov:
> Just trying to clarify my understanding.
> 
> This is related to the `--no-compression` flag change because the `-z` flag on non-compressed files will cause tar to fail?
> 
> Any by removing the `z`, since tar can auto-detect, it can safely expand things that were tar-ed without compression?

Yep, exactly

rkoster (approved):

PR #2636, opened 2025-10-22
remove duplicate ip addresses with smaller prefix

fmoehler:

We have to clear duplicate ip address with a smaller prefix in the addresses we cant allocate during dynamic ip allocation.

In the context of introducing prefix delegation for the bosh director we refactored the way restricted_ips are handled in the code. Instead of having single ips as an array we switched to having the biggest cidr ranges possible. This would solve a bug where the bosh director would become unresponsive (out of memory) if the number of ip addresses exceeded a certain limit, especially relevant for ipv6 use cases where ranges get very big very easily.

E.g. a /80 network contains 281,474,976,710,656 (approximately 281.5 trillion) addresses. Previously this would result in an array with the same amount of elements. With the change it could be as low as only 1.

However this leads to issues when existing addresses randomly match one of the restricted ips cidr range, because in the code we delete all ip addresses below the current ip we are trying to allocate not paying attention to the cidr range. 

This can result in the error:

 Failed to reserve IP '10.0.72.2/32' for network 'diego-cells': IP belongs to reserved range

Unit tests

Fix for  "Failed to reserve IP 'xxx' for network 'xxx': IP belongs to reserved range"

no
@dudejas

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
@@ -98,7 +98,7 @@ module Bosh::Director::DeploymentPlan
 
       addresses_we_cant_allocate.reject! do |ip|
         addresses_we_cant_allocate.any? do |other_ip|
-          includes = other_ip.include?(ip) rescue false
+          includes = other_ip.include?(ip)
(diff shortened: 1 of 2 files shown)


a-hassanin (approved):
LGTM

aramprice (dismissed):

PR #2641, opened 2025-11-18
add prefix to network settings for dynamic and vip network

fmoehler:

This PR will add the prefix to the network settings of dynamic and vip networks which was previously missed. Although these network types do not support prefix allocation (yet) they should have this property as well to make all network types (manual, dynamic, vip) have the same config.

This issue was revealed after https://github.com/cloudfoundry/bosh/commit/9ec3b74683a2098f05a3a4fd7d55c76772eb283b#r170570246 was merged. 

Reran previously failing integration tests + unit tests.
 
Set prefix in network_settings of dynamic and vip networkds

No

@neddp , @Ivaylogi98 , @mariash

### DIFF
--- a/src/bosh-director/lib/bosh/director/models/vm.rb
+++ b/src/bosh-director/lib/bosh/director/models/vm.rb
@@ -39,7 +39,9 @@ module Bosh::Director::Models
     end
 
     def dynamic_ips
-      network_spec.map { |_, network| "#{network['ip']}/#{network['prefix']}" }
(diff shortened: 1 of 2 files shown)


a-hassanin:
multiple similar errors "Invalid IP or CIDR format: 166.179.121.68/" 
@fmoehler didn't we implement a default to a single address ?

aramprice:
@fmoehler - thanks this looks good. I think go ahead and merge so we can get the pipeline back in working order.

PR #2649, opened 2025-12-04
Add metrics for VM states

yuriadam-sap:

This adds new bosh metrics regarding the VM states. Initially the idea was proposed in this [PR](https://github.com/cloudfoundry/bosh/pull/2597) that which was reverted [here](https://github.com/cloudfoundry/bosh/pull/2600) due to breaking the system design. 

Initial PR which was reverted defined the state "unhealthy" if the VM state is not "running". In this PR, each state has its own metric, and there is additional metric "unhealthy" when the VM state is running but doesn't have any processes. This requires modification in the agent heartbeat to include the number of processes. The corresponding PR for the agent heartbeat is [here](https://github.com/cloudfoundry/bosh-agent/pull/390).

The new metrics are:

- bosh_unhealthy_agents
- bosh_total_available_agents
- bosh_failing_instances
- bosh_stopped_instances
- bosh_unknown_instances

Ran the unit tests and also tested a bosh dev release on a development environment with the modified bosh-agent.

Bosh now emits different states of VM as metrics, which can be used to monitor the state of the instances per deployment over time

No

@DennisAhausSAP

### DIFF
--- a/src/bosh-director/lib/bosh/director/metrics_collector.rb
+++ b/src/bosh-director/lib/bosh/director/metrics_collector.rb
@@ -154,7 +154,12 @@ module Bosh
         response = Net::HTTP.get_response('127.0.0.1', endpoint, @config.health_monitor_port)
         return unless response.is_a?(Net::HTTPSuccess)
 
-        deployment_counts = JSON.parse(response.body) rescue nil
+        begin
+          deployment_counts = JSON.parse(response.body)
+        rescue JSON::ParserError => e
+          @logger.warn("Failed to parse JSON response from #{endpoint}: #{e.message}")
+          return
+        end
         return unless deployment_counts.is_a?(Hash)
 
         existing_deployment_names = gauge.values.map do |key, _|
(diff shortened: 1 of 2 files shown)


aramprice:
I think these changes may be breaking CI.

The first failure is in https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-director/jobs/integration-mysql/builds/889

```
rspec ./spec/integration/health_monitor/hm_stateless_spec.rb:72 # health_monitor: 1 resurrects vms that were down before resurrector started
rspec ./spec/integration/health_monitor/hm_stateless_spec.rb:96 # health_monitor: 1 when the vm is launching and has incomplete heartbeats only outputs complete heartbeats
```

Since then CI has been failing consistently on `spec/integration/health_monitor/hm_stateless_spec.rb:96`
=> Builds 889 through latest (892 at time of writing)

https://bosh.ci.cloudfoundry.org/teams/main/pipelines/bosh-director/jobs/integration-mysql/builds/892

PR #2652, opened 2025-12-15
Add dynamic disks support

mariash:

This PR adds support for dynamically managed disks. 

The following API endpoints are added for dynamic disk management:

* `POST /dynamic_disks/provide`
* `POST /dynamic_disks/:disk_name/detach`
* `DELETE /dynamic_disks/:disk_name`

The following corresponding jobs were added:

* `ProvideDynamicDisk` - creates disk in the cloud if it doesn't exist, saves record in dynamic_disks database table and attaches it to VM, calls `add_dynamic_disk` agent method
* `DetachDynamicDisk` - detaches disk from VM, calls `remove_dynamic_disk` agent method
* `DeleteDynamicDisk` - deletes disk in the cloud and from the database.

These operations are idempotent.

The implementation also introduces:

* New database migration adds dynamic_disks table

* VM lock for VM cloud operations - attach disk, detach disk, reboot VM and delete VM to make sure only 1 operation is executed per VM.

* New UAA scopes for dynamic disk operations introduced:
  * `bosh.dynamic_disks.update (provide/detach)`
  * `bosh.dynamic_disks.delete (delete)`
  * included in Director admin scopes
* Dynamic disks are deleted when deployment is deleted
* Dedicated dynamic disk workers and queue (configurable via a new `dynamic_disks_workers` Director property; when > 0, a separate worker process is started and drain behavior is updated accordingly).

* Corresponding bosh agent PR - https://github.com/cloudfoundry/bosh-agent/pull/393

* `bundle exec rake spec`
* manual deploy of updated bosh release, updating CF and using dynamic disks feature.

Add support for dynamic disks

No, this PR only adds new functionality.

@Alphasite @mkocher @cppforlife

### DIFF
--- a/src/bosh-director/lib/bosh/director/api/controllers/dynamic_disks_controller.rb
+++ b/src/bosh-director/lib/bosh/director/api/controllers/dynamic_disks_controller.rb
@@ -5,7 +5,7 @@ module Bosh::Director
     class DynamicDisksController < BaseController
       include ValidationHelper
 
-      post '/provide', scope: :update_dynamic_disks, consumes: :json do
+      post '/provide', scope: :provide_dynamic_disks, consumes: :json do
         request_hash = JSON.parse(request.body.read)
 
         instance_id = safe_property(request_hash, 'instance_id', class: String, min_length: 1)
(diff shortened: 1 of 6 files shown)


mariash:
Marking as "Ready for review" since RFC is in FCP. @Alphasite updated this PR to use scopes defined in the RFC.

PR #2657, opened 2026-01-29
Fix IP Allocation Bug: Reserved Range Not Detected

neddp:

This PR potentially fixes a bug where BOSH allocates IPs from reserved CIDR ranges, causing CPI failures with "Address is in subnet's reserved address range" (AWS) errors and similar for other providers.

### The Bug

We were unable to reproduce the issue. We can only speculate what could be causing it.

### The Fix

**1. Add `overlaps?` method to `IpAddrOrCidr`** — uses integer range comparison, bypassing `IPAddr#coerce_other` entirely:

```ruby
def overlaps?(other)
  my_first = @ipaddr.to_range.first.to_i
  my_last = @ipaddr.to_range.last.to_i
  other_first = other.to_range.first.to_i
  other_last = other.to_range.last.to_i
  my_first <= other_last && other_first <= my_last
end
```

**2. Prune by last address** — only discard a CIDR when its entire range is behind the current position:

```ruby
filtered_ips.reject! { |ip| ip.to_range.last.to_i < current_prefix.to_i }
```

**3. Capture `blocking_ip` before mutation** — stable reference for the advance calculation:

```ruby
blocking_ip = filtered_ips.first
if blocking_ip&.overlaps?(current_prefix)
  if blocking_ip.count > current_prefix.count
    current_ip = to_ipaddr(blocking_ip.to_i + blocking_ip.count)
  else
    current_ip = to_ipaddr(current_ip.to_i + current_prefix.count)
  end
```

### What tests have you run?

- All existing unit tests pass
- Added edge case tests for overlapping CIDR blocks with different prefix sizes

### Release notes

**Fixed:** IP allocation now correctly detects reserved CIDR ranges. Previously, when allocating single IPs (`/32`) from a subnet with reserved ranges specified as larger blocks (e.g., `/30`), the algorithm failed to detect the overlap and attempted to allocate reserved IPs, causing CPI errors.

### Breaking change?

No. This is a bug fix.

@dudejas
@Ivaylogi98

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
@@ -106,7 +106,7 @@ module Bosh::Director::DeploymentPlan
       ip_address_cidr = find_next_available_ip(addresses_we_cant_allocate, first_range_address, subnet.prefix)
 
       if !(subnet.range == ip_address_cidr || subnet.range.include?(ip_address_cidr))
-       raise NoMoreIPsAvailableAndStopRetrying
+        raise NoMoreIPsAvailableAndStopRetrying
       end
 
       save_ip(ip_address_cidr, reservation, false)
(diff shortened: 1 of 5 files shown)


src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
```diff
@@ -96,49 +96,57 @@ def try_to_allocate_dynamic_ip(reservation, subnet)
         addresses_we_cant_allocate.delete_if { |ipaddr| ipaddr.ipv6? }
       end
 
-      addresses_we_cant_allocate.reject! do |ip|
-        addresses_we_cant_allocate.any? do |other_ip|
-          includes = other_ip.include?(ip)
-          includes && other_ip.prefix < ip.prefix
+      # Sort by address first, then by prefix (smaller prefix = larger block = earlier)
+      sorted_ips = addresses_we_cant_allocate.sort_by { |ip| [ip.to_i, ip.prefix] }
```
Ivaylogi98:
```suggestion
      uniq_ips = addresses_we_cant_allocate.uniq
      sorted_ips = uniq_ips.sort_by { |ip| [ip.to_i, ip.prefix] }
```

`uniq` would remove all duplicate items in the array, which I'm not sure is possible, to have duplicate items, but would be a small optimisation and make it safer

src/bosh-director/lib/bosh/director/deployment_plan/ip_provider/ip_repo.rb
fmoehler: Please see comment https://github.com/cloudfoundry/bosh/pull/2657#issuecomment-3883966099

neddp:
Moving to draft for the moment. Needs further attention.

PR #2661, opened 2026-02-03
(feat) - add storage-cli to blob.yml

gowrisankar22:

This change is to bring new strorage-cli https://github.com/cloudfoundry/storage-cli/tree/main/s3 later to migrate from bosh-s3cli. 

https://github.com/cloudfoundry/bosh/compare/main...gowrisankar22:bosh:support-storage-cli?expand=1#diff-4773dca3c56343b4eb62c73e472115949bc2b7d384e82b6b9ebf00efe0451675L263

NO


gowrisankar22:
will be aligned

PR #2662, opened 2026-02-04
Fix undefined method `detached?` for String

neddp:
Following on PR #2659 which broke the integration tests for BOSH:
`instance.state.detached?` is called, but `instance.state` returns a String (not a StringInquirer), so it doesn't have the `.detached?` method.

### Changes
- **Add state constants** to `DeploymentPlan::Instance`:
  - `STATE_STARTED`, `STATE_STOPPED`, `STATE_DETACHED`
- **Add predicate methods** to `DeploymentPlan::Instance`:
  - `started?`, `stopped?`, `detached?`
- **Refactor all state checks** to use predicate methods across:
- Fix lint warning
- Remove duplicate `attr_accessor` declarations in `instance_group.rb`

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/instance_group.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/instance_group.rb
@@ -14,7 +14,13 @@ module Bosh::Director
       # recreate and restart are two virtual states
       # (both set  target instance state to "started" and set
       # appropriate instance spec modifiers)
-      VALID_STATES = %w[started stopped detached recreate restart].freeze
+      VALID_STATES = [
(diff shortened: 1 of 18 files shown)


aramprice (dismissed):

src/bosh-director/lib/bosh/director/deployment_plan/instance_group.rb
```diff
@@ -56,13 +56,6 @@ class InstanceGroup
       # @return [DeploymentPlan::UpdateConfig] Instance group update settings
       attr_reader :update
 
-      # @return [Array<DeploymentPlan::Instance>] All instances
```
aramprice: Unsure about this deletion

aramprice (approved):

PR #2664, opened 2026-02-04
Add syslog gem for Ruby 3.4+ compatibility

neddp:
Takes care of the following error and any potential issues when we move to Ruby 3.4+:
```
warning: syslog was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
```

### DIFF
--- a/src/Gemfile
+++ b/src/Gemfile
@@ -7,6 +7,7 @@ gem 'bosh-common', path: 'bosh-common'
 
 gem 'mysql2'
 gem 'pg'
+gem 'syslog'
 
 group :development, :test do
   gem 'bundle-audit'


aramprice (changes requested):
This change looks fine though I'd rather _not_ add a transitive dependency to our Gemfile if we can help it.

This appears to be coming from the `logging` gem which has a guard around the loading of `syslog`[1] so I think we are ok w/o this change.

[1] https://github.com/TwP/logging/blob/master/lib/logging.rb#L9-L14

If this is annoying enough that we do want to make this addition to the Gemfile, let's move it next to where we specify 

```ruby
spec.add_dependency 'logging'
```

in the gemspec files with a comment noting that the dependency is there because of `logging`.

PR #2716, opened 2026-04-20
Fix static IP subnet mismatch when adding network to nic_group

vlast3k:
## Summary

- When adding a new network to existing instances in the same `nic_group`, the placement planner picked the first unclaimed static IP in the AZ without checking which cloud subnet it belonged to. This caused CPI errors (`Networks in nic_group have different subnet_ids`) when an AZ has multiple cloud subnets with interleaved static IP ranges.
- Store `cloud_properties` from the containing subnet on each `StaticIpToAzs` entry. When selecting an IP for a network that shares a `nic_group` with an already-assigned sibling, constrain the lookup to IPs on a subnet with matching `cloud_properties`.
- Fail with a clear error when no matching IP is available, instead of silently assigning a mismatched one.

## What changed

**`networks_to_static_ips.rb`**
- `StaticIpToAzs` struct now includes `cloud_properties`
- `create()` records the subnet's `cloud_properties` for each IP
- Added `find_by_network_az_and_cloud_properties` and `next_ip_for_network_with_cloud_properties`

**`static_ips_availability_zone_picker.rb`**
- `create_network_plan_with_az` looks up the cloud_properties of an already-assigned sibling in the same nic_group, then constrains IP selection to matching subnets
- Added `nic_group_sibling_cloud_properties` helper

**Tests**
- `networks_to_static_ips_spec.rb`: cloud_properties storage and filtered lookup
- `static_ips_availability_zone_picker_spec.rb`: multi-subnet nic_group pairing, mismatch error, no-nic_group unchanged behavior

## Context

On AWS, all networks in a `nic_group` must resolve to the same ENI subnet. Landscapes with multiple `/24` subnets per AZ and interleaved static IP ranges hit this when enabling IPv6 (adding a second network to the nic_group). BOSH assigned IPv6 IPs on a different subnet than the existing IPv4 IPs because it only filtered by AZ, not by cloud subnet.

Verified against a production cloud config: paired IPv4/IPv6 subnets sharing a `nic_group` have identical `cloud_properties` (same AWS subnet ID, same security groups), because AWS subnets are dual-stack and a `nic_group` maps to a single ENI. Full `cloud_properties` hash equality is the correct comparison — narrowing to just the `subnet` key would be fragile and CPI-specific.

## Test plan

- [ ] Existing placement planner specs pass (no behavioral change when nic_group is not used)
- [ ] New spec: two networks in nic_group, two subnets per AZ with different cloud_properties — IPs are paired on matching subnets
- [ ] New spec: mismatched static IPs (no IP on sibling subnet) raises clear error
- [ ] New spec: without nic_group, behavior unchanged
- [ ] Integration test on a landscape with multi-subnet AZ + dual-stack nic_group

### DIFF
--- a/src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
@@ -123,9 +123,13 @@ module Bosh
                                    @networks_to_static_ips.next_ip_for_network(network)
                                  end
               if static_ip_to_azs.nil?
+                if sibling_cloud_props
+                  raise Bosh::Director::NetworkReservationError,
+                        "Failed to find a static IP for network '#{network.name}' on a matching subnet " \
+                        "in nic_group '#{network.nic_group}'"
+                end
                 raise Bosh::Director::NetworkReservationError,
-                      "Failed to find a static IP for network '#{network.name}' on a matching subnet " \
-                      "in nic_group '#{network.nic_group}'"
+                      'Failed to distribute static IPs to satisfy existing instance reservations'
               end
               if static_ip_to_azs.az_names.size == 1
                 az_name = static_ip_to_azs.az_names.first
@@ -290,6 +294,7 @@ module Bosh
 
             instance_plan.network_plans.each do |plan|
               sibling_network = @job_networks.find { |jn| jn.deployment_network == plan.reservation.network }
+              next if sibling_network == network
               next unless sibling_network&.nic_group == network.nic_group
               next unless plan.reservation.ip
 
--- a/src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
+++ b/src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
@@ -115,8 +115,18 @@ module Bosh
           def create_network_plan_with_az(instance_plan, network, instance_plans)
             desired_instance = instance_plan.desired_instance
             instance = instance_plan.instance
+            sibling_cloud_props = nic_group_sibling_cloud_properties(instance_plan, network)
             if desired_instance.az.nil?
-              static_ip_to_azs = @networks_to_static_ips.next_ip_for_network(network)
+              static_ip_to_azs = if sibling_cloud_props
+                                   @networks_to_static_ips.next_ip_for_network_with_cloud_properties(network, sibling_cloud_props)
+                                 else
+                                   @networks_to_static_ips.next_ip_for_network(network)
+                                 end
+              if static_ip_to_azs.nil?
+                raise Bosh::Director::NetworkReservationError,
+                      "Failed to find a static IP for network '#{network.name}' on a matching subnet " \
+                      "in nic_group '#{network.nic_group}'"
+              end
               if static_ip_to_azs.az_names.size == 1
                 az_name = static_ip_to_azs.az_names.first
                 @logger.debug("Assigning az '#{az_name}' to instance '#{instance}'")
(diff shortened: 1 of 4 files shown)


vlast3k:
## Validation: `cloud_properties` hash equality is safe

A potential concern was raised during review: is full `cloud_properties == cloud_properties` hash equality too strict? Could paired IPv4/IPv6 subnets in the same `nic_group` have slightly different `cloud_properties`?

**Verified against a production AWS cloud config.** When two BOSH networks share a `nic_group` (e.g., IPv4-ext and IPv6-single-ext), their subnets in each AZ reference the **same AWS subnet** — because AWS subnets are dual-stack by default and a `nic_group` maps to a single ENI. The `cloud_properties` hashes are identical:

```yaml
# IPv4 network, z1, subnet 1
cloud_properties:
  security_groups: [sg-aaa, sg-bbb, sg-ccc, sg-ddd]
  subnet: subnet-12345

# IPv6 network, z1, subnet 1
cloud_properties:
  security_groups: [sg-aaa, sg-bbb, sg-ccc, sg-ddd]
  subnet: subnet-12345
```

All 8 subnets across both networks were 1:1 matched — same `subnet` ID, same `security_groups` list. The only differences between the two BOSH subnets are in non-cloud_properties fields (`range`, `gateway`, `dns`, `static`).

**Conclusion:** Full hash equality is correct. Narrowing the comparison to just the `subnet` key would be fragile — it assumes the cloud_properties schema, which varies by CPI (AWS vs GCP vs Azure vs vSphere).

rkoster (approved):

plamen-bardarov (approved):

vlast3k:
The PR was successfully validated in a test landscape. The challenge to reproduce it was the fact that if the IPs are properly distributed across the VMs, and subnets it does not trigger. So they had to be intentionally mixed, what actually happens in real life, when VMs are added and removed and assigned VMs from subnets (we faced it in 7+ old CF Deployment)
@neddp

fmoehler (approved):

src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
```diff
@@ -115,8 +115,22 @@ def place_new_instance_plans(desired_instances, instance_plans)
...
             desired_instance = instance_plan.desired_instance
             instance = instance_plan.instance
+            sibling_cloud_props = nic_group_sibling_cloud_properties(instance_plan, network)
             if desired_instance.az.nil?
-              static_ip_to_azs = @networks_to_static_ips.next_ip_for_network(network)
+              static_ip_to_azs = if sibling_cloud_props
+                                   @networks_to_static_ips.next_ip_for_network_with_cloud_properties(network, sibling_cloud_props)
+                                 else
+                                   @networks_to_static_ips.next_ip_for_network(network)
+                                 end
```
aramprice, line 124: This same conditional appears on lines [143-147](https://github.com/cloudfoundry/bosh/pull/2716/changes#diff-ecf18ba14f5a2a51e6b44bc8704f191acfd33ab78dd46f16234b6c3d778f2f7eR143-R147) below, should it move into some method? Possibly on `@networks_to_static_ips`?

aramprice (approved):
Looks good, thanks for tracking this down. I suggested an optional cleanup.

neddp (approved):
LGTM

src/bosh-director/lib/bosh/director/deployment_plan/placement_planner/static_ips_availability_zone_picker.rb
vlast3k: @aramprice the code is actually a bit different, and if we extract them to a dedicated method, it would need to take a handful of params. IMO  if this becomes a bit more complex it would make sense to refactor it. I would leave it like this, but if you prefer to encapsulate it within networks_to_static_ips i could do it as well

Ivaylogi98 (approved):

aramprice (approved):

PR #2720, opened 2026-04-28
Rename ca property

aramprice:
Improves naming of the CA Certificate files passed to bosh components.


ragaskar (approved):
@aramprice walked me through this code and it lgtm -- there is a corresponding change in the deployment to pass the uaa cert if uaa is being used (although we expect to this cert to effectively be the same as the director cert for most deployments), but if this field isn't passed we'll fall back to passing the director cert.

PR #2739, opened 2026-06-02
Fix Sequel postgres SQL threading race condition

colins:
Sequel's Dataset#select_sql caches and returns the same mutable String object to all concurrent callers. Under high concurrency, log_connection_yield prepends "(conn: NNNNN)" in-place to that shared buffer, which is then sent verbatim to Postgres as SQL:
  PG::SyntaxError: ERROR: syntax error at or near "conn"

Add apply_postgres_thread_safety_patch to Config#configure_db, called after Sequel.connect when the postgres adapter is loaded. The patch prepends a module onto Sequel::Postgres::Adapter that wraps execute_query to pass String.new(sql).freeze to super, giving each call its own frozen independent C-level string buffer so concurrent PQexec calls cannot interfere.

An idempotency guard prevents double-application if configure_db is called multiple times during startup.

Specs pin the upstream Sequel caching behaviour so any gem upgrade that fixes the issue upstream will alert us to remove the patch. Unit tests cover: String.new.freeze isolation, args passthrough, idempotency, and no-op for non-postgres adapters.

We have a threading test that exposes the issue. The corruption occurs in the form of SQL the query string being mutated before being sent to Postgres. Examples of the mutated thread are this string being sent to Postgres
(conn: 152) SELECT * FROM "bug_templates" INNER JOIN "join_r...

The string corruption occurs in all of the tested 3.3 ruby release; 3.3.9, 3.3.11, 3.4.9

The string corruption does not occur in ruby 4.0.5

This patch will be short lived until the bump to ruby 4.0

[TNZ-104200, TNZ-103317]
ai-assisted=yes

A customer reported Postgres failures caused by a syntax error in the received SQL string.  This seems to have
been caused by a concurrency problem that was mutating a string. This patches the symptom of the failure.
The long term fix is to bumnp ruby to 4.0+

We created an issue with the sequel gem but it appears that ruby is the real cuplrit.
https://github.com/jeremyevans/sequel/issues/2368

cd src/spec
bundle exec rake spec:unit:parallel

A temporary fix to mitigate a string corruption race condition that only affects the ruby 3.3 releases

No

@aramprice

### DIFF
--- a/src/bosh-director/lib/bosh/director/config.rb
+++ b/src/bosh-director/lib/bosh/director/config.rb
@@ -350,6 +350,8 @@ module Bosh::Director
 
         db = wait_for_db_connection(connection_config, connection_wait_timeout)
 
+        apply_postgres_thread_safety_patch(connection_config['adapter'])
+
         Bosh::Common.retryable(sleep: DEFAULT_DB_CONNECTION_RETRY_INTERVAL, tries: 20, on: [Exception]) do
           db.extension :connection_validator
           true
@@ -363,6 +365,32 @@ module Bosh::Director
         end
 
         db
+      end
+
+      # Sequel's Dataset#select_sql caches and returns the same mutable
+      # String object to all concurrent callers. Under high concurrency, log_connection_yield
+      # prepends a "(conn: NNNNN)" debug prefix in-place to this shared string, which is then
+      # sent verbatim to Postgres as SQL, causing:
+      #   PG::SyntaxError: ERROR: syntax error at or near "conn"
+      #
+      # Fix: String.new(sql) forces a memcpy, giving each execute_query call its own independent
+      # C-level string buffer so concurrent PQexec calls cannot interfere.
+      #
+      # Must run after Sequel.connect (when the postgres adapter module is first loaded).
+      # Guarded against double-application in case configure_db is called more than once.
+      def apply_postgres_thread_safety_patch(adapter_name)
+        return unless %w[postgres postgresql].include?(adapter_name)
+        return unless defined?(Sequel::Postgres::Adapter)
+        return if @postgres_thread_safety_patched
+
+        @postgres_thread_safety_patched = true
+        Sequel::Postgres::Adapter.prepend(Module.new do
+          private
+
+          def execute_query(sql, args)
+            super(String.new(sql).freeze, args)
+          end
+        end)
       end
 
       def wait_for_db_connection(connection_config, timeout)
--- a/src/bosh-director/spec/unit/bosh/director/postgres_thread_safety_patch_spec.rb
+++ b/src/bosh-director/spec/unit/bosh/director/postgres_thread_safety_patch_spec.rb
@@ -1,0 +1,95 @@
+require 'spec_helper'
+
+RSpec.describe Bosh::Director::Config, '#apply_postgres_thread_safety_patch' do
+  # Reset idempotency guard so tests are independent
+  before { described_class.instance_variable_set(:@postgres_thread_safety_patched, nil) }
+  after  { described_class.instance_variable_set(:@postgres_thread_safety_patched, nil) }
+
+  let(:fake_adapter) do
+    Class.new do
+      def execute_query(sql, args)
+        [sql, args]
+      end
+      private :execute_query
+    end
+  end
+
+  context 'when adapter is postgres and Sequel::Postgres::Adapter is defined' do
+    before { stub_const('Sequel::Postgres::Adapter', fake_adapter) }
+
+    it 'prepends a module that passes String.new(sql).freeze to prevent cross-thread SQL corruption' do
+      described_class.send(:apply_postgres_thread_safety_patch, 'postgres')
+
+      original_sql = 'SELECT * FROM templates INNER JOIN release_versions_templates ON id = 1'
+      result = fake_adapter.new.send(:execute_query, original_sql, nil)
+
+      expect(result[0]).to eq(original_sql), 'SQL content must be preserved'
+      expect(result[0].object_id).not_to eq(original_sql.object_id),
+        'Expected a fresh String.new copy to isolate thread buffers and prevent ' \
+        '(conn: NNNNN) prefix injection into the actual SQL sent to Postgres'
+      expect(result[0]).to be_frozen,
+        'Expected the SQL copy to be frozen so in-place C-level mutations raise FrozenError ' \
+        'rather than silently corrupting concurrent threads'
+    end
+
+    it 'passes args through to the underlying execute_query unmodified' do
+      described_class.send(:apply_postgres_thread_safety_patch, 'postgres')
+
+      query_args = [42, 'bar']
+      result = fake_adapter.new.send(:execute_query, 'SELECT 1', query_args)
+
+      expect(result[1]).to be(query_args), 'args must be passed through without modification'
+    end
+
+    it 'is idempotent — applies the patch only once even when called multiple times' do
+      2.times { described_class.send(:apply_postgres_thread_safety_patch, 'postgres') }
+
+      patched_count = fake_adapter.ancestors.take_while { |m| m != fake_adapter }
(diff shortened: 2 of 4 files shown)


src/bosh-director/spec/unit/bosh/director/sequel_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    end
+
+    it 'concurrent callers receive the same object_id, demonstrating the shared-buffer race' do
+      dataset = db[:templates]
+      ids = Array.new(4) { dataset.select_sql.object_id }
+
+      expect(ids.uniq.length).to eq(1),
+        'All concurrent callers are expected to get the SAME String object (same object_id). ' \
+        'If this fails the upstream caching behaviour has changed.'
+    end
```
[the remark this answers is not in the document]
aramprice: Probably not significant as we expect this patch not to be needed for long.

src/bosh-director/spec/unit/bosh/director/sequel_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+    # to all concurrent callers. Sequel's log_connection_yield prepends "(conn: NNNNN)"
+    # in-place to that shared string, which gets injected into the SQL sent to Postgres:
+    #   PG::SyntaxError: ERROR: syntax error at or near "conn"
+    #
+    # Fix: String.new(sql) performs memcpy, giving each execute_query call its own
+    # independent C-level buffer — concurrent PQexec calls cannot interfere.
+    #
+    # This patch is applied in Config#apply_postgres_thread_safety_patch after
+    # Sequel.connect (when Sequel::Postgres::Adapter is first loaded).
+    # See config_spec.rb for integration-level coverage.
```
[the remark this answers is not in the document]
aramprice: Probably not significant as we expect this patch not to be needed for long.

src/bosh-director/lib/bosh/director/config.rb
```diff
@@ -365,6 +367,31 @@ def configure_db(db_config)
...
+        return if @postgres_thread_safety_patched
+
+        @postgres_thread_safety_patched = true
+        Sequel::Postgres::Adapter.prepend(Module.new do
+          private
+
+          def execute_query(sql, args)
+            super(String.new(sql).freeze, args)
+          end
+        end)
```
beyhan: Two minor comments: 
* Should we add a # TODO: remove after Ruby 4.0 upgrade comment to the method so it doesn't outlive its purpose?
* Frozen string caveat: if any code after the execute_query call in the Postgres adapter mutates sql (e.g., for logging), it will now raise FrozenError. This is likely intentional (surfacing bugs) but worth confirming nothing downstream mutates the sql argument after this call site.

beyhan (commented):
It looks good to me. I just have the two minor points I commented.

src/bosh-director/lib/bosh/director/config.rb
```diff
@@ -365,6 +367,31 @@ def configure_db(db_config)
         db
       end
 
+      # TNZ-104200 / TNZ-103317: Sequel's Dataset#select_sql caches and returns the same mutable
```
rkoster: What are these TNZ-104200 / TNZ-103317 references? Could not find anything public when searching for these.

src/bosh-director/spec/unit/bosh/director/postgres_thread_safety_patch_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'spec_helper'
+
+RSpec.describe Bosh::Director::Config, '#apply_postgres_thread_safety_patch (TNZ-104200)' do
```
rkoster: again. TNZ-104200?

rkoster (dismissed):
Tests could be better, but probably not worth it, since we intent to remove this code once we bump to ruby 4.x.

src/bosh-director/spec/unit/bosh/director/sequel_spec.rb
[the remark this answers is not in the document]
colins: The test is confirming that all of the object_ids are identical for each caller. Not that there is a concurrency problem.

src/bosh-director/lib/bosh/director/config.rb
colins: I added a test that will fail when ruby gets bumped to 4.0+
  it 'is obsolete on Ruby 4.0+ and must be removed'
  
  the .freeze will cause a failure in tests should anything change in the sequel gem.

src/bosh-director/lib/bosh/director/config.rb
colins: removed

src/bosh-director/spec/unit/bosh/director/postgres_thread_safety_patch_spec.rb
colins: Removed

src/bosh-director/spec/unit/bosh/director/sequel_spec.rb
```diff
@@ -0,0 +1,113 @@
+require 'spec_helper'
+
+RSpec.describe 'Bosh::Director sequel extensions' do
+  # -------------------------------------------------------------------------
+  # Upstream Sequel gem behaviour pinning (TNZ-104200 / TNZ-103317)
```
aramprice: probably still want to yank these

src/bosh-director/spec/unit/bosh/director/sequel_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+      dataset = db[:templates]
+      ids = Array.new(4) { dataset.select_sql.object_id }
+
+      expect(ids.uniq.length).to eq(1),
+        'All concurrent callers are expected to get the SAME String object (same object_id). ' \
+        'If this fails the upstream caching behaviour has changed.'
+    end
+  end
+
+  describe 'String.new(sql) thread-safety pattern (TNZ-104200 / TNZ-103317)' do
```
aramprice: ibid

src/bosh-director/spec/unit/bosh/director/postgres_thread_safety_patch_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    it 'passes args through to the underlying execute_query unmodified' do
+      described_class.send(:apply_postgres_thread_safety_patch, 'postgres')
+
+      query_args = [42, 'bar']
+      result = fake_adapter.new.send(:execute_query, 'SELECT 1', query_args)
+
+      expect(result[1]).to be(query_args), 'args must be passed through without modification'
+    end
+
```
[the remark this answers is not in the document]
aramprice: Unless this codebase allows both variants we probably don't care.

src/bosh-director/lib/bosh/director/config.rb
```diff
@@ -365,6 +367,32 @@ def configure_db(db_config)
...
+      #
+      # Fix: String.new(sql) forces a memcpy, giving each execute_query call its own independent
+      # C-level string buffer so concurrent PQexec calls cannot interfere.
+      #
+      # Must run after Sequel.connect (when the postgres adapter module is first loaded).
+      # Guarded against double-application in case configure_db is called more than once.
+      def apply_postgres_thread_safety_patch(adapter_name)
+        return unless adapter_name == 'postgres'
+        return unless defined?(Sequel::Postgres::Adapter)
+        return if @postgres_thread_safety_patched
```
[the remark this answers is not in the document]
aramprice: Unless this codebase allows both variants we probably don't care.

aramprice (approved):
EasyCLA appears confused by the git-bot co-signing on the latest commit.

We can override, or you can re-push the commit without any git-bot attribution, and that should fix it.

rkoster (approved):

PR #2745, opened 2026-06-19
Fail BATs teardown task if cleanup fails

beyhan:
We observe many orphaned disks in our GCP account. We shuld cleanup properly to avoid unnecessary costs.


beyhan:
@aramprice, @rkoster this is the result of our FI WG discussion yesterday. We don't need to set `max_orphaned_age_in_days` because the BATs are setting up a director and then it gets teared down. The teardown takes care to cleanup all resource but failures in cleanup were ignored. That is why, I'm proposing this change.

rkoster (approved):

PR #2746, opened 2026-06-19
Replace Ruby bosh-nats-sync with Golang implementation

aramprice:
## Summary

- Replaces the Ruby `bosh-nats-sync` gem with a Go binary (`src/bosh-nats-sync/`) that mirrors all Ruby behavior including TLS verification preferences (`uaa_ca_cert` > `director_ca_cert`), HTTP peer verification, and startup reliability
- Removes all Ruby runtime dependencies from the `nats` BOSH job and package (`director-ruby-3.3`, gem bundling, `BUNDLE_GEMFILE`/`GEM_HOME` env vars)
- Adds GitHub Actions CI coverage for the new Go code (lint + test jobs in `go.yml`); removes the defunct `nats_sync:parallel` Ruby matrix entry from `ruby.yml`

## Commits

1. **Golang bosh-nats-sync** — full Go implementation with parity to the Ruby version:
   - Correct TLS CA selection (`uaa_ca_cert` preferred over `director_ca_cert`)
   - Proper Director API TLS peer verification (no `InsecureSkipVerify`)
   - `isConnectionError` retry logic covering `deadline exceeded` / `eof` (matching Ruby's `Net::OpenTimeout`, `Net::ReadTimeout`, `Errno::ECONNRESET`)
   - Fatal startup behavior on failed NATS reload (matching Ruby)
   - Full Ginkgo/Gomega test suite

2. **Remove ruby-isms now that nats is golang** — strips Ruby from `jobs/nats/`, `packages/nats/`, `src/Gemfile`, and CI:
   - `packages/nats/packaging` now builds the Go binary via `go build`
   - `jobs/nats/` wrapper script calls Go binary directly (no Ruby runtime sourcing)
   - CI `go.yml`: adds `lint (bosh-nats-sync)` and `test (bosh-nats-sync)` jobs
   - CI `ruby.yml`: removes `nats_sync:parallel` matrix entry
   - Adds `src/bosh-nats-sync/.golangci.yml`

## Test plan

- [ ] `go.yml` / lint (bosh-nats-sync) passes
- [ ] `go.yml` / test (bosh-nats-sync) passes
- [ ] `ruby.yml` / unit_specs passes for all remaining sub-projects (`common:parallel`, `monitor:parallel`, `release`)
- [ ] Release ERB template specs (`spec/nats_templates_spec.rb`) pass


src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,344 @@
...
+	names := make([]string, len(deployments))
+	for i, d := range deployments {
+		names[i] = d.Name
+	}
+	return names, nil
+}
+
+func (u *UsersSync) getVMsByDeployment(deploymentName string) ([]natsauthconfig.VM, error) {
+	body, err := u.boshAPIResponseBody(fmt.Sprintf("/deployments/%s/vms", deploymentName), true)
+	if err != nil {
```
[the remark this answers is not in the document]
aramprice: prefer filepath.Join()

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,351 @@
...
+	var vms []natsauthconfig.VM
+	overwriteableConfigFile := true
+
+	err := u.withDirectorConnection(func() error {
+		var queryErr error
+		vms, queryErr = u.queryAllRunningVMs()
+		return queryErr
+	})
+
+	if err != nil {
```
colins, line 97: The ruby code used to do this:
```ruby
  it 'raises an error after exhausting retries' do
    expect { subject.execute_users_sync }.to raise_error(Errno::ECONNREFUSED)
  end
```
This is now logging an error and possibly updating the auth file. I assume this is to address the "startup reliability" 

Could this be called out somewhere?

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+
+	names := make([]string, len(deployments))
+	for i, d := range deployments {
+		names[i] = d.Name
+	}
+	return names, nil
+}
+
+func (u *UsersSync) getVMsByDeployment(ctx context.Context, provider *authprovider.AuthProvider, deploymentName string) ([]natsauthconfig.VM, error) {
+	body, err := u.boshAPIResponseBody(ctx, path.Join("/deployments", url.PathEscape(deploymentName), "vms"), provider)
```
aramprice: **[Must Fix] `path.Join` is a filesystem-path function, not a URL-path function.**

```go
body, err := u.boshAPIResponseBody(ctx, path.Join("/deployments", url.PathEscape(deploymentName), "vms"), provider)
```

`path.Join` calls `path.Clean` under the hood, which **drops empty path segments**. If `url.PathEscape(deploymentName)` ever produces an empty string (e.g., blank `agent_id` slips through), `path.Join("/deployments", "", "vms")` silently produces `"/deployments/vms"`, querying the wrong endpoint without error.

Use explicit string construction instead:
```go
vmPath := "/deployments/" + url.PathEscape(deploymentName) + "/vms"
body, err := u.boshAPIResponseBody(ctx, vmPath, provider)
```

Or use Go 1.19+ `url.JoinPath` which correctly handles URL semantics and lets you remove the `"path"` import entirely.

aramprice (commented):
## Code Review: Go rewrite of bosh-nats-sync

Reviewed with the combined perspective of Mat Ryer, Peter Bourgon, and Dave Cheney. Tests are green, `go vet` is clean, and the build is clean. Overall the rewrite is structurally sound — the type-based `isConnectionError`, the CA cert fallback chain (uaa_ca_cert → director_ca_cert → system trust), and the context propagation are all done well. The `Bootstrap()` feature is a genuine improvement over the Ruby version.

**Two must-fix items** (correctness bugs): `path.Join` for URL construction, and a temp-directory leak in the tests. The rest are quality/consistency issues that should still be addressed before merge.

src/bosh-nats-sync/pkg/userssync/users_sync_test.go
```diff
@@ -0,0 +1,1197 @@
...
+		var err error
+		natsConfigFile, err = os.CreateTemp("", "nats_config_*.json")
+		Expect(err).NotTo(HaveOccurred())
+		natsConfigFilePath = natsConfigFile.Name()
+		natsConfigFile.Close()
+
+		natsExecutable = "/var/vcap/packages/nats/bin/nats-server"
+		natsServerPIDFile = "/var/vcap/sys/run/bpm/nats/nats.pid"
+
+		tmpDir, err := os.MkdirTemp("", "nats_sync_test_*")
```
aramprice: **[Must Fix] Temp directory created in `BeforeEach` is never cleaned up.**

```go
tmpDir, err := os.MkdirTemp("", "nats_sync_test_*")
Expect(err).NotTo(HaveOccurred())
// ... directorSubjectFile and hmSubjectFile placed inside tmpDir ...
```

The `AfterEach` removes the individual files (`os.Remove(directorSubjectFile)`, `os.Remove(hmSubjectFile)`) but never removes `tmpDir` itself. Every test run leaks a directory under the OS temp path.

Fix by replacing the three individual `os.Remove` calls for the subject files with:
```go
os.RemoveAll(tmpDir)
```
(The `natsConfigFilePath` is a separate temp file and still needs its own `os.Remove`.)

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+		return true
+	}
+	var parsed map[string]interface{}
+	if err := json.Unmarshal(data, &parsed); err != nil {
+		return true
+	}
+	return len(parsed) == 0
+}
+
+func readSubjectFile(filePath string) *string {
```
aramprice, line 320: **[Should Fix] `readSubjectFile` has a TOCTOU race and a redundant Stat call.**

```go
func readSubjectFile(filePath string) *string {
    info, err := os.Stat(filePath)
    if err != nil || info.Size() == 0 {
        return nil
    }
    data, err := os.ReadFile(filePath)
    ...
}
```

Two issues:

1. **TOCTOU**: `os.Stat` and `os.ReadFile` are two separate syscalls. The file can change between them (e.g., written by the BOSH director). This is a classic time-of-check/time-of-use race.

2. **`info.Size() == 0` is misleading**: It returns `nil` for a zero-byte file, but a file containing only whitespace (size > 0) is also semantically empty. The `TrimSpace` check below correctly handles this — the `Stat` check does not.

Simplify to eliminate both issues:
```go
func readSubjectFile(filePath string) *string {
    data, err := os.ReadFile(filePath)
    if err != nil {
        return nil
    }
    s := strings.TrimSpace(string(data))
    if s == "" {
        return nil
    }
    return &s
}
```

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+	var netErr net.Error
+	if errors.As(err, &netErr) && netErr.Timeout() {
+		return true
+	}
+	// Any other operation against a network connection (dial/read/write).
+	var opErr *net.OpError
+	return errors.As(err, &opErr)
+}
+
+func (u *UsersSync) userFileOverwritable() bool {
```
aramprice, line 308: **[Should Fix] `userFileOverwritable` and `hasExistingUsers` use inconsistent semantics for "is the file empty?".**

`userFileOverwritable` (used by `Execute`) unmarshals into `map[string]interface{}` and checks `len(parsed) == 0`. A file with content `{"authorization":{"users":[]}}` has a non-empty top-level map, so `userFileOverwritable` returns **false** — Execute will NOT overwrite.

`hasExistingUsers` (used by `Bootstrap`) unmarshals into `AuthorizationConfig` and checks `len(cfg.Authorization.Users) > 0`. The same file has an empty users slice, so `hasExistingUsers` returns **false** — Bootstrap **will** overwrite.

This means Bootstrap and Execute disagree about whether `{"authorization":{"users":[]}}` is an overwriteable state. While this exact file content is unlikely in production, the divergence is a latent bug and makes the codebase harder to reason about.

**Suggestion**: Unify behind a single helper, e.g.:
```go
// hasRegisteredUsers reports whether auth.json already contains at least one user entry.
func (u *UsersSync) hasRegisteredUsers() bool {
    data, err := os.ReadFile(u.natsConfigFilePath)
    if err != nil { return false }
    var cfg AuthorizationConfig  // or map[string]interface{} — pick one and use it everywhere
    if err := json.Unmarshal(data, &cfg); err != nil { return false }
    return len(cfg.Authorization.Users) > 0
}
```
Then use `!u.hasRegisteredUsers()` in both `Execute` and `Bootstrap`.

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+	// Wait for the director to be reachable and, on success, build a single
+	// AuthProvider from the /info response. The provider is reused for every
+	// authenticated request in this pass, so /info is fetched once and (for a
+	// UAA director) a token is minted once per sync rather than per request.
+	provider, err := u.withDirectorConnection(ctx)
+	if err == nil {
+		vms, err = u.queryAllRunningVMs(ctx, provider)
+	}
+
+	if err != nil {
```
aramprice, line 97: **[Should Fix] Misleading log message when context is cancelled mid-director-probe.**

When `r.Stop()` is called while `Execute` is waiting for the director, `withDirectorConnection` returns `context.Canceled`. The error falls into the `if err != nil` branch and logs:

```
"Could not query all running vms" error="context canceled"
```

This is misleading — no VM query was ever attempted, and the cancel is not an error from the operators perspective; it is expected shutdown.

Add an early return for context errors before the error-handling block:
```go
if err != nil {
    if ctx.Err() != nil {
        // Context cancelled (graceful shutdown); skip this sync pass cleanly.
        return nil
    }
    u.logger.Error("Could not query all running vms", "error", err)
    ...
}
```

The runner already checks `ctx.Err() != nil` after `Execute` returns, so this aligns with that pattern.

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+		return nil, fmt.Errorf("cannot access: %s, Status Code: %d, %s", apiPath, resp.StatusCode, string(body))
+	}
+
+	return body, nil
+}
+
+// httpClient lazily builds and caches the director HTTP client so its transport
+// (connection pool + CA trust store) is reused for the lifetime of the process.
+func (u *UsersSync) httpClient() (*http.Client, error) {
+	u.clientOnce.Do(func() {
```
aramprice, line 391: **[Should Document] `sync.Once` makes a failed `buildHTTPClient` permanently fatal — document this intention.**

```go
func (u *UsersSync) httpClient() (*http.Client, error) {
    u.clientOnce.Do(func() {
        u.client, u.clientErr = u.buildHTTPClient()
    })
    return u.client, u.clientErr
}
```

If `buildHTTPClient` fails (e.g., the CA cert file is corrupt), every subsequent call to `httpClient()` returns the same error for the lifetime of the process. This is the correct behaviour — a misconfigured CA cert cannot be fixed at runtime without a restart — but it is non-obvious.

Add a comment so the next reader does not reach for a retry loop:
```go
// httpClient lazily builds and caches the director HTTP client. If construction
// fails (e.g., an invalid CA cert), the error is permanent for this process:
// the underlying sync.Once guarantees Do is not retried, which is intentional
// since CA-cert misconfiguration requires operator intervention and a restart.
func (u *UsersSync) httpClient() (*http.Client, error) {
```

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+		return nil
+	}
+	s := strings.TrimSpace(string(data))
+	if s == "" {
+		return nil
+	}
+	return &s
+}
+
+func (u *UsersSync) natsFileHash() string {
```
aramprice, line 332: **[Nice to Have] `natsFileHash` reads the file from disk after `writeNATSConfigFile` already has the bytes.**

```go
currentFileHash := u.natsFileHash()
// ... write config ...
newFileHash := u.natsFileHash()
if currentFileHash != newFileHash { ... }
```

`writeNATSConfigFile` computes the serialized bytes into a `bytes.Buffer` and then writes them. `natsFileHash` then reads those same bytes back from disk to compute the hash. The second read can be avoided by having `writeNATSConfigFile` return the hash of what it wrote:

```go
func (u *UsersSync) writeNATSConfigFile(vms []VM, directorSubject, hmSubject *string) (string, error) {
    ...
    data := bytes.TrimRight(buf.Bytes(), "\n")
    if err := os.WriteFile(..., data, 0644); err != nil {
        return "", err
    }
    return fmt.Sprintf("%x", md5.Sum(data)), nil
}
```

This removes a disk read per sync cycle and makes the "did the file change?" check exact (no race between write and re-read).

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
+package userssync
+
+import (
+	"bytes"
```
aramprice, line 4: **[Nice to Have] `crypto/md5` will be flagged by `gosec` (G501).**

MD5 is used for change-detection (not cryptographic hashing), which matches the Ruby implementation and is functionally correct. However, `gosec` unconditionally flags any import of `crypto/md5` with rule G501 ("Blocklisted import crypto/md5").

Two options:
1. Switch to `crypto/sha256` — identical API, no functional change, no linter noise.
2. If MD5 parity with Ruby is intentional, add `//nolint:gosec` on the `md5.Sum` call with a comment explaining why (`"non-cryptographic change-detection hash, mirrors Ruby Digest::MD5"`).

Option 1 is preferable — there is no reason to keep MD5 here.

src/bosh-nats-sync/pkg/authprovider/auth_provider.go
```diff
@@ -0,0 +1,145 @@
...
+	"bosh-nats-sync/pkg/config"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
+)
+
+const ExpirationDeadline = 60 * time.Second
+
+// httpClientTimeout bounds a single UAA token request.
+const httpClientTimeout = 30 * time.Second
```
aramprice, line 30: **[Nice to Have] `httpClientTimeout` is duplicated across `authprovider` and `userssync` packages.**

`authprovider.httpClientTimeout` and `userssync.httpClientTimeout` are both `30 * time.Second` with similar meaning (single outbound HTTP request). If the timeout ever changes — or if they should have different values — a maintainer has to find both.

Either:
- Extract to a shared `internal/defaults` package with a single named constant, or
- Add a comment at each site cross-referencing the other: `// 30s — intentionally matches userssync.httpClientTimeout`

This prevents them from silently diverging in future.

src/bosh-nats-sync/pkg/runner/runner.go
```diff
@@ -0,0 +1,105 @@
...
+			cancel()
+		case <-ctx.Done():
+		}
+	}()
+
+	// Bootstrap: write the initial NATS config from on-disk subject files
+	// immediately, before the director is queried.  This replaces the
+	// placeholder token written by pre-start so that health_monitor and the
+	// director can authenticate against NATS during director startup.
+	if err := r.sync.Bootstrap(); err != nil {
```
aramprice, line 70: **[Should Verify] Bootstrap skips the NATS reload when no subject files exist — Ruby always reloads.**

The Ruby implementation calls `UsersSync.reload_nats_server_config` unconditionally on startup. Go's `Bootstrap()` is a no-op when neither subject file exists.

This means: if bosh-nats-sync starts when no subject files have been written yet, NATS does NOT receive a reload signal. NATS continues running with whatever config it had before (typically the `fooBar` placeholder from pre-start). The director and health-monitor will not be able to authenticate until the first periodic `Execute` succeeds.

**Question for reviewer/author**: Is there a pre-start guarantee that subject files always exist before bosh-nats-sync starts? If yes, the skip is safe and should be documented. If not, Bootstrap should fall back to at least sending the reload signal so NATS is not left with a stale placeholder indefinitely.

src/bosh-nats-sync/pkg/natsauthconfig/nats_auth_config_test.go
```diff
@@ -0,0 +1,159 @@
...
+		directorSubject *string
+		hmSubject       *string
+	)
+
+	BeforeEach(func() {
+		vms = []natsauthconfig.VM{
+			{PermanentNATSCredentials: false, AgentID: "fef068d8-bbdd-46ff-b4a5-bf0838f918d9"},
+			{PermanentNATSCredentials: false, AgentID: "c5e7c705-459e-41c0-b640-db32d8dc6e71"},
+		}
+		directorSubject = strPtr("subject=C=USA, O=Cloud Foundry, CN=default.hm.bosh-internal")
```
aramprice: **[Nit] Test fixture data is confusing — directorSubject has wrong CN and a spurious prefix.**

```go
directorSubject = strPtr("subject=C=USA, O=Cloud Foundry, CN=default.hm.bosh-internal")
hmSubject       = strPtr("C=USA, O=Cloud Foundry, CN=default.hm.bosh-internal")
```

Two issues:
1. `directorSubject` has a spurious `"subject="` prefix that does not appear in real BOSH subject files.
2. Both subjects refer to `hm.bosh-internal` — the director subject should say `director.bosh-internal`.

The test happens to pass because `CreateConfig` treats the subject value as an opaque string, but a reviewer reading this fixture will be confused about what format the real values have. Use realistic values:

```go
directorSubject = strPtr("C=USA, O=Cloud Foundry, CN=default.director.bosh-internal")
hmSubject       = strPtr("C=USA, O=Cloud Foundry, CN=default.hm.bosh-internal")
```

src/bosh-nats-sync/pkg/authprovider/auth_provider_test.go
```diff
@@ -0,0 +1,450 @@
...
+
+				header, err := provider.AuthHeader(context.Background())
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("failed to obtain token from UAA"))
+				Expect(header).To(BeEmpty())
+			})
+		})
+
+		Context("with director_ca_cert file", func() {
+			It("uses the director_ca_cert when the file exists and is non-empty", func() {
```
aramprice: **[Nit] The "director_ca_cert" test does not exercise TLS — it connects to a plain-HTTP test server.**

The test verifies that building the HTTP client with a valid CA cert does not break ordinary HTTP requests, which is useful. However since `uaaServer` is an `httptest.NewServer` (not `httptest.NewTLSServer`), the configured `RootCAs` pool is never exercised in a TLS handshake.

To fully test that `director_ca_cert` is actually trusted during TLS connections, the test should use `httptest.NewTLSServer` and pass the server's certificate as the CA cert — mirroring the pattern already used in the `users_sync_test.go` TLS verification suite. The existing "does not skip TLS verification" test in the same file is the correct model.

src/bosh-nats-sync/pkg/userssync/users_sync.go
```diff
@@ -0,0 +1,482 @@
...
+	}
+}
+
+// isConnectionError reports whether err is a transient connection-class failure
+// worth retrying, mirroring the Ruby DIRECTOR_CONNECTION_ERRORS list
+// (ECONNREFUSED, ECONNRESET, ETIMEDOUT, EHOSTUNREACH, ENETUNREACH,
+// Net::OpenTimeout, Net::ReadTimeout, SocketError). It classifies by error type
+// via errors.Is/errors.As rather than matching message text, so HTTP status
+// errors (which carry no network error type) are never misclassified.
+func isConnectionError(err error) bool {
```
aramprice, line 259: **[Observation] The final catch-all `errors.As(err, &opErr)` in `isConnectionError` may be broader than intended.**

```go
var opErr *net.OpError
return errors.As(err, &opErr)
```

Any error that wraps `*net.OpError` is classified as a connection error worth retrying. In practice this is fine because:
- TLS handshake failures (`x509.UnknownAuthorityError`) are NOT wrapped in `*net.OpError` by Go's HTTP client, so they correctly fall through as non-retryable (verified by the test suite).
- The `withDirectorConnection` deadline bounds the total retry window.

However, the catch-all is non-obvious and could become a correctness problem if Go's HTTP stack changes how it wraps certain errors in a future release. Consider adding a comment:

```go
// Catch-all for low-level network operation errors (dial, read, write).
// Note: TLS certificate verification errors are NOT *net.OpError; they fall
// through this function and are not retried.
var opErr *net.OpError
return errors.As(err, &opErr)
```

src/bosh-nats-sync/pkg/userssync/users_sync.go
aramprice: This is now documented in the code. The intentional difference from Ruby is: Go logs the error and falls back to writing a director/HM-only config (when auth.json is empty), rather than crashing and relying on BPM to restart. This improves startup reliability by giving health_monitor a chance to authenticate against NATS while the director is still coming up. A comment was added to Execute calling this out explicitly: https://github.com/cloudfoundry/bosh/pull/2746/commits/82bfb98ea9

aramprice:
## Behavioral parity audit: Ruby vs Go

A complete pass comparing the Ruby source (`src/bosh-nats-sync/lib/nats_sync/`) against this Go implementation. All deviations are noted; deviations marked **improvement** are intentional and correct.

### Startup sequence

| # | Aspect | Ruby | Go |
|---|--------|------|----|
| 1 | Signal handling | `Signal.trap('INT')` only | SIGINT **and SIGTERM** ✅ improvement (BPM sends SIGTERM) |
| 2 | Initial NATS reload | Always sends reload; picks up whatever config pre-start wrote | Bootstrap(): reads subject files, writes explicit director/HM config, **then** sends reload ✅ improvement |
| 3 | First sync timing | `Rufus::Scheduler.interval` fires after the interval | `time.Ticker` fires after the interval — identical timing |

### execute_users_sync / Execute

| # | Aspect | Ruby | Go |
|---|--------|------|----|
| 4 | AuthProvider lifecycle | `NATSSync::AuthProvider.new(info, …)` **created on every authenticated request** → new UAAToken → new UAA grant per request | AuthProvider created **once per Execute pass**; token cached across all /deployments + /deployments/*/vms calls ✅ improvement (reduces UAA load) |
| 5 | UAA token failure | Logs error, sets `@uaa_token = nil`, returns `nil` auth header → director returns 401 → RuntimeError caught → fallback | Returns error to caller → fallback. Same outcome, clearer error chain ✅ |
| 6 | Connection error retry | `Bosh::Common.retryable(tries: N, on: DIRECTOR_CONNECTION_ERRORS)` — number-of-attempts based | Deadline-based with 1 s sleep between retries — equivalent timeout |
| 7 | Deployment URL in path | `/deployments/#{deployment}/vms` — **no URL encoding** | `/deployments/` + `url.PathEscape(name)` + `/vms` — URL-encodes name ✅ (BOSH names are `[a-z0-9-]` so no practical difference) |
| 8 | Empty agent_id VMs | No check — writes `CN=.agent.bosh-internal` (invalid entry) | Skips VMs with empty `agent_id` ✅ bug fixed |
| 9 | JSON serialization | `JSON.unparse` — no HTML-escaping (preserves `>` literally) | `json.NewEncoder` with `SetEscapeHTML(false)` — identical output ✅ |
| 10 | Change-detection hash | `Digest::MD5.file(path).hexdigest` | `sha256.Sum256` — same semantics (non-cryptographic change detection) ✅ |
| 11 | read_subject_file (whitespace-only) | `File.empty?` is false for whitespace-only; `strip` returns `""` → returns `""` (not nil!) | TrimSpace → returns nil | Minor difference: Ruby would add a director user with `""` subject; Go correctly skips it. ✅ |

### auth_header / AuthProvider

| # | Aspect | Ruby | Go |
|---|--------|------|----|
| 12 | Basic auth encoding | `Base64.encode64(…).strip` — strip removes leading/trailing whitespace but NOT mid-string newlines Ruby inserts every 60 chars. For long credentials, produces an invalid RFC 2617 header | `base64.StdEncoding.EncodeToString` — no newlines, RFC-compliant ✅ |
| 13 | CA cert fallback | `uaa_ca_cert` → `director_ca_cert` → system trust | `CAFilePath()`: identical fallback chain ✅ |
| 14 | TLS verification | `verify_mode = VERIFY_PEER` always | `tls.Config{MinVersion: TLS12}` — always verifies ✅ |

### Error / shutdown handling

| # | Aspect | Ruby | Go |
|---|--------|------|----|
| 15 | Sync error | Scheduler `on_error` logs fatal + `shutdown` | Runner returns error → main.go `os.Exit(1)` — equivalent |
| 16 | Stop | `@scheduler.shutdown` | Close `stopCh` + wait for Run() — equivalent graceful shutdown |

### NatsAuthConfig

| # | Aspect | Ruby | Go |
|---|--------|------|----|
| 17 | Permissions structure | Identical `publish`/`subscribe` arrays per user type | Identical ✅ |
| 18 | CN format | `"C=USA, O=Cloud Foundry, CN=#{cn}.agent.bosh-internal"` | Identical ✅ |
| 19 | bootstrap user | Included when `!vm['permanent_nats_credentials']` | Identical ✅ |

**Verdict**: The Go implementation is a faithful and improved replacement for the Ruby version. Every behavioral difference is an improvement (SIGTERM support, per-Execute auth caching, empty-agent-id guard, RFC-compliant Base64, clearer UAA error propagation).

aramprice:
## Idiomatic Go audit

A second pass reviewing the implementation for Go clarity and convention (not Ruby-translation patterns).

### Strengths ✅

- `errors.Is`/`errors.As` for connection-error classification — correct and idiomatic vs. string-matching
- `context.Context` propagated through every I/O call — correct
- `sync.Once` for lazy HTTP client — correct (documented as permanently-fatal, which is the right semantic)
- Functional options pattern (`WithCommandRunner`) — idiomatic for testability
- `slog.Logger` — modern (Go 1.21+)
- `export_test.go` for exposing unexported symbols in external test packages — correct Go idiom
- Error wrapping with `%w` throughout — correct
- `defer func() { _ = resp.Body.Close() }()` — correct ignore-close-error pattern
- Table-driven tests via `DescribeTable`/`Entry` — well-suited to the `isConnectionError` matrix

### Issues found and fixed in this review pass

1. **`getCommandRunner()` was dead code** — `NewUsersSync` now normalises a nil cmdRunner in the constructor, and all call-sites use `u.commandRunner` directly. Removed.

2. **Runner test temp file leak** — `dirSubjectFile` and `hmSubjectFile` were never cleaned up in `AfterEach`. Fixed (analogous to the `users_sync_test.go` fix).

### Design observations (no code change required, but worth noting)

3. **`uaaTokenHeader` holds the mutex for the entire HTTP call** — `a.mu.Lock()` is held while `a.cfg.Token(tokenCtx)` makes an outbound HTTP request (up to 30 s). Since there is only one sync goroutine at a time, this is safe and correct. A future refactor could use double-checked locking or `sync.Once` per token, but the current design is fine.

4. **`readSubjectFile` returns `*string`** — this is idiomatic in Go when nil is semantically meaningful. An alternative `(string, bool)` return would also be idiomatic; either is fine.

5. **Module path `bosh-nats-sync` without a domain prefix** — acceptable for a standalone binary; would need a qualified path (e.g., `github.com/cloudfoundry/bosh/src/bosh-nats-sync`) if external importability were ever needed.

6. **`Bootstrap()` always sends a NATS reload even if the config hasn't changed** — unlike `Execute` which skips reload when the hash matches, Bootstrap always reloads after writing. This is correct for startup (we want NATS to definitely apply the new config), and the extra SIGHUP is harmless.

rkoster:
# Code Review

**PR #2746**: Replace Ruby bosh-nats-sync with Golang implementation

## Summary

Replaces the Ruby `bosh-nats-sync` gem with a Go binary that mirrors the Ruby behavior (TLS CA selection preferring uaa_ca_cert over director_ca_cert, peer verification always on, connection-error retry classification) and removes all Ruby runtime dependencies from the `nats` job/package. This is a high-quality, well-structured port: code is idiomatic, extensively documented with rationale comments, and well-tested (84-100% statement coverage; verified locally to pass `go vet`, `golangci-lint` with 0 issues, and `go test -race`). The author (aramprice) has already self-reviewed thoroughly and addressed nearly all earlier human + bot feedback across 15 commits (md5->sha256, path.Join->explicit URL build, TOCTOU in readSubjectFile, sync.Once doc, timeout cross-reference, temp-dir cleanup, intentional Ruby-vs-Go behavior documented). The PR is still a Draft. The one real merge-blocker is the new golang-1.26-linux package this PR introduces: the Go port needs a Go toolchain to compile the nats package, but main has no golang package, and the PR's current packaging stub does not actually install a Go toolchain. The remaining points are a documentation nit and PR hygiene (BATs tooling duplicated from PR #2753).

## Critical Issues - Must Fix

1. **The new golang-1.26-linux package does not install a Go toolchain, so a clean nats-package compile fails; it needs to be a proper vendored package**
   - File: `packages/golang-1.26-linux/packaging`
   - main has no golang package at all - this PR introduces packages/golang-1.26-linux because the Go port of bosh-nats-sync now needs a Go toolchain to build the nats package. The author first added it as a vendored-blob (commit e724cc2a added only spec.lock with fingerprint 8c9cb9ef...), then in commit 6044f412 ('Fix bosh create-release to work for integration tests') removed that spec.lock and replaced it with a hand-written spec + packaging stub. That stub only renders compile.env/runtime.env and never installs a Go distribution into BOSH_INSTALL_TARGET. Since packages/nats/packaging sources that compile.env (GOROOT=/var/vcap/packages/golang-1.26-linux, PATH=$GOROOT/bin:...) and then runs `go build`, a clean `bosh create-release` + compile fails with 'go: command not found'. The fly-bats pipeline papers over this with an inline /usr/local/bin/go shim (commit 024fa9cd), which confirms the package itself does not provide Go.
   - Fix: Make golang-1.26-linux a real vendored package: `bosh vendor-package golang-1.26-linux <path-to-checkout-of-cloudfoundry/bosh-package-golang-release>` (that release builds Go from source / ships the toolchain). This regenerates packages/golang-1.26-linux/spec.lock and uploads the toolchain blob to the release blobstore, so the nats package compiles from a clean checkout with no external Go shim. Then remove the stub spec/packaging and src/golang-1.26-linux/*.env.unix, and drop the /usr/local/bin/go workaround from ci/fly-bats.yml. Verify `bosh create-release` + compilation end-to-end. The author's own commit sequence (spec.lock first, stub second 'to work for integration tests') shows the vendored-blob approach was the intent and the stub is a temporary stand-in until the blob is published.

## Positive Notes

- Excellent behavioral-parity documentation: comments explicitly call out where and why the Go implementation intentionally differs from Ruby (startup-reliability fallback in Execute, empty agent_id skip in CreateConfig, sync.Once making CA-cert errors fatal).
- Strong test suite: 84-100% statement coverage per package; verified locally to pass go vet, golangci-lint (0 issues), and go test -race cleanly. TLS verification, per-error-type retry classification, and the no-CA system-trust path are all covered.
- Good security posture: TLS peer verification is always on (no InsecureSkipVerify), MinVersion TLS 1.2, a configured-but-unparseable CA cert is a hard error rather than a silent fallback, and CGO_ENABLED=0 static build.
- The HTML-escape fix (SetEscapeHTML(false)) for '>' in NATS subjects and the null agent_id skip are real bug fixes over the Ruby behavior, each backed by a clear comment and tests.
- Thorough author self-review: nearly all earlier reviewer and bot findings (path.Join, readSubjectFile TOCTOU, md5->sha256, temp-dir leak, test fixture realism, timeout duplication) were already addressed in follow-up commits.
- vendor/ is committed (807 files) and packaging uses -mod=vendor, so the Go build is offline/reproducible - the one piece still missing is the Go toolchain package itself.

## Recommendation

Request changes - critical issues need to be addressed before merging.

rkoster (changes requested):
See the detailed review comment above. The one merge-blocker is the new golang-1.26-linux package needing to be a proper vendored package (bosh vendor-package from cloudfoundry/bosh-package-golang-release) so the nats package compiles from a clean checkout. Otherwise this is a strong, well-tested Go port.

rkoster:
## Design question: reuse existing BOSH/CF Go libraries instead of hand-rolling?

Non-blocking, but worth deciding explicitly and recording in the PR. This package re-implements three things that BOSH/CF already ship as Go libraries.

### 1. TLS config -> `code.cloudfoundry.org/tlsconfig` (clearest win)

`authprovider/auth_provider.go` and `userssync/users_sync.go` both build their TLS client by hand:

```go
tlsCfg := &tls.Config{MinVersion: tls.VersionTLS12}
// ... manual x509.NewCertPool() + AppendCertsFromPEM() + tlsCfg.RootCAs = pool
```

[`tlsconfig`](https://github.com/cloudfoundry/tlsconfig) is the CF-standard helper for exactly this and maps directly:

```go
cfg, err := tlsconfig.Build(
    tlsconfig.WithExternalServiceDefaults(),
).Client(
    tlsconfig.WithAuthorityFromFile(caPath), // only when a CA is configured
)
```

`WithExternalServiceDefaults()` additionally pins `MaxVersion` TLS 1.3 and a curated Mozilla-Intermediate cipher-suite list, and `WithAuthorityFromFile` replaces the manual cert-pool loading. So the current hand-rolled config is actually *less* hardened than the org-standard one (no cipher-suite pinning, no `MaxVersion`).

Unlike the bosh-cli reuse below, footprint is not a real objection here: `tlsconfig` is single-purpose with a tiny dependency set (the file-CA path is essentially stdlib). The one behavior to preserve is "fall back to the system trust store when no CA is configured" - apply `WithAuthority*` only when a pool is present.

This one looks like a near-unconditional improvement.

### 2. Director API -> `github.com/cloudfoundry/bosh-cli/v7/director`

[`director.Director`](https://github.com/cloudfoundry/bosh-cli/blob/main/director/interfaces.go) exposes `Info()`, `Deployments()`, and `Deployment.VMInfos()` - exactly the three calls `userssync` hand-rolls (`/info` + auth-type detection, `/deployments`, `/deployments/{name}/vms`). `director.FactoryConfig` already encodes the `uaa_ca_cert`-over-`director_ca_cert` precedence and the TLS/retry behavior you re-derived from the Ruby.

### 3. UAA -> `github.com/cloudfoundry/bosh-cli/v7/uaa`

[`uaa.ClientTokenSession.TokenFunc()`](https://github.com/cloudfoundry/bosh-cli/blob/main/uaa/client_token_session.go) is the same memoized client-credentials-grant + `"Type Value"` header pattern as `authprovider.uaaTokenHeader`/`formatBearer`.

Reusing (2)/(3) would remove roughly 250 lines of bespoke HTTP/JSON/UAA code and keep CA/TLS/retry semantics in lock-step with the rest of BOSH. The real cost is dependency weight: `bosh-cli/v7` pulls `bosh-utils` + reporters + a large transitive tree (vs. the current lean `x/oauth2`/`x/net`/`yaml`/ginkgo set), the binary is intentionally static (`CGO_ENABLED=0`), and `director.Factory.New` wants `TaskReporter`/`FileReporter` args that are irrelevant here (no-op reporters). So (2)/(3) are a genuine reuse-vs-footprint trade-off, not a clear win.

### Ask

Were these (especially `tlsconfig`) considered and rejected, or are they candidate follow-ups? Either answer is fine - just worth a note in the PR so the next person does not wonder why TLS, the director API, and UAA were reimplemented by hand.

aramprice:
> See the detailed review comment above. The one merge-blocker is the new golang-1.26-linux package needing to be a proper vendored package (bosh vendor-package from cloudfoundry/bosh-package-golang-release) so the nats package compiles from a clean checkout. Otherwise this is a strong, well-tested Go port.

The golang package situation will need some manual intervention, also true for https://github.com/cloudfoundry/bosh/pull/2747, since using `bosh vendor-package` to add golang requires access to the `private.yml` for uploading things to the blobstore. I think the right move will be to manually add the vendored golang package to `main`, and then rebase these two PR's and remove the hacked in golang package.

aramprice:
> ### Ask
> Were these (especially `tlsconfig`) considered and rejected, or are they candidate follow-ups? Either answer is fine - just worth a note in the PR so the next person does not wonder why TLS, the director API, and UAA were reimplemented by hand.

The use of `code.cloudfoundry.org/tlsconfig` makes sense, adding that. 

In principle I agree with reusing the bosh-cli code as well, though I am not excited about creating additional consumers given how much hand-rolled non-`stdlib` code exists in the bosh-cli /bosh-utils codebases. I'd rather consume these at a later date once things like [FileSystem](https://github.com/cloudfoundry/bosh-utils/blob/master/system/file_system_interface.go#L19) have been replaced with go's stdlib implementations.

PR #2757, opened 2026-06-27
Remove CPI v1 Support

aramprice:
CPI v1 has been deprecated for a long time, we should stop testing it.

=> https://bosh.ci.cloudfoundry.org/builds/365976656 ✅


src/bosh-director/lib/clouds/external_cpi_response_wrapper.rb
```diff
@@ -20,7 +20,6 @@ def set_disk_metadata(*arguments); invoke_cpi_method(__method__.to_s, *arguments
     def create_disk(*arguments); invoke_cpi_method(__method__.to_s, *arguments); end
     def has_disk(*arguments); invoke_cpi_method(__method__.to_s, *arguments); end
     def delete_disk(*arguments); invoke_cpi_method(__method__.to_s, *arguments); end
-    def attach_disk(*arguments); invoke_cpi_method(__method__.to_s, *arguments); end
```
aramprice, line 23: Why is the `attach_disk()` method removed?

aramprice: Delegator was removed

src/tasks/spec.rake
```diff
@@ -85,13 +85,13 @@ namespace :spec do
...
     desc 'Run all unit tests in parallel'
-    multitask parallel: %w[spec:unit:release:parallel] + component_dir_names.map{|d| "spec:unit:#{component_symbol(d)}:parallel" } do
+    multitask parallel: %w[spec:unit:release:parallel spec:unit:integration_support:parallel] + component_dir_names.map{|d| "spec:unit:#{component_symbol(d)}:parallel" } do
       trap('INT') { exit }
     end
   end
 
   desc 'Run all unit tests'
-  task unit: %w[spec:unit:release] + component_dir_names.map{|d| "spec:unit:#{component_symbol(d)}" }
+  task unit: %w[spec:unit:release spec:unit:integration_support] + component_dir_names.map{|d| "spec:unit:#{component_symbol(d)}" }
```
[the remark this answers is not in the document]
aramprice: This is not correct, without explicit inclusion of specs under `spec/integration_support/spec/` they are not executed.

src/bosh-director/lib/clouds/external_cpi_response_wrapper.rb
aramprice: The generic one-liner delegator for `attach_disk` was removed from the list because the class already has an explicit `attach_disk` implementation at lines 46-50. Having both definitions caused a `Lint/DuplicateMethods` RuboCop warning.

The explicit implementation is what we need — it raises `Bosh::Clouds::AttachDiskResponseError` when the CPI returns nil or an empty disk hint. The generic delegator would have silently overridden this validation logic (Ruby uses the last definition of a method), so removing it ensures the validation is always active.

aramprice:
@coderabbitai review

aramprice:
@coderabbitai review

src/bosh-director/lib/bosh/director/cloud_factory.rb
```diff
@@ -58,9 +58,9 @@ def get(cpi_name, stemcell_api_version = nil)
 
       begin
         info_response = cloud.info || {}
-        cpi_api_version = info_response.fetch('api_version', 1)
-      rescue
-        cpi_api_version = 1
+        cpi_api_version = info_response.fetch('api_version')
+      rescue StandardError
+        raise Bosh::Clouds::NotSupported, 'CPI must report api_version via info'
       end
```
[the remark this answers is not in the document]
aramprice: Fixed — narrowed the rescue to specifically catch `Bosh::Clouds::NotImplemented` (for CPIs that don't implement `info`) and use a `fetch` block for the missing `api_version` key, so real transport or execution errors propagate naturally.

src/spec/integration_support/dummy_cpi.rb
```diff
@@ -0,0 +1,936 @@
...
+
+    def read_all
+      @logger.debug("Reading all inputs: #{File.read(ordered_file_path)}")
+      result = []
+      File.read(ordered_file_path).split("\n").each do |request|
+        data = JSON.parse(request)
+        result << CpiInvocation.new(data['method_name'], data['inputs'], data['context'])
+      end
+      result
+    end
```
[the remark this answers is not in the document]
aramprice: Fixed — `read_all` now returns an empty array early if the file doesn't exist, and reads the file only once into `content` before logging and parsing.

rkoster (approved):

neddp (approved):

src/bosh-director/lib/bosh/director/cloud_factory.rb
```diff
@@ -57,10 +57,12 @@ def get(cpi_name, stemcell_api_version = nil)
...
-        info_response = cloud.info || {}
-        cpi_api_version = info_response.fetch('api_version', 1)
-      rescue
-        cpi_api_version = 1
+        info_response = cloud.info
+      rescue Bosh::Clouds::NotImplemented
+        raise Bosh::Clouds::NotSupported, 'CPI must report api_version via info'
+      end
+      cpi_api_version = (info_response || {}).fetch('api_version') do
+        raise Bosh::Clouds::NotSupported, 'CPI must report api_version via info'
```
ystros, line 65: 🎉 🔥

ystros (approved):

PR #2766, opened 2026-07-10
fix: semver resource uses git

mkocher:

This makes the semver resource use git instead of a gcs bucket that runs the risk of being garbage collected.

 #2765


ystros (approved):

aramprice (approved):

aramprice:
It seems like the pipeline may have already been re-configured, going to merge this.

PR #2781, opened 2026-07-15
packaging: copy `.bundle/confg` into package

aramprice:
This file is created by the `bosh_bundle_local` function used to install gems during packaging, but because the function is not executed while in `$BOSH_INSTALL_TARGET` the `.bundle/` directory is not present in the package.

### DIFF
--- a/packages/director/packaging
+++ b/packages/director/packaging
@@ -36,5 +36,4 @@ bundle config set build.pg \
 bosh_bundle_local
 bosh_generate_runtime_env
 
-cp Gemfile ${BOSH_INSTALL_TARGET}
-cp Gemfile.lock ${BOSH_INSTALL_TARGET}
+cp -vr Gemfile Gemfile.lock .bundle "${BOSH_INSTALL_TARGET}"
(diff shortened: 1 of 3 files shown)


packages/director/packaging
```diff
@@ -36,5 +36,4 @@ bundle config set build.pg \
 bosh_bundle_local
 bosh_generate_runtime_env
 
-cp Gemfile ${BOSH_INSTALL_TARGET}
-cp Gemfile.lock ${BOSH_INSTALL_TARGET}
+cp -vr Gemfile Gemfile.lock .bundle "${BOSH_INSTALL_TARGET}"
```
ystros, line 39: Do we truly want verbose mode on these copies?

ystros (approved):

packages/director/packaging
aramprice: ¯\_(ツ)_/¯ - seems relatively harmless since it's only 3 files, but I'm not sure where this output would ever be visible

aramprice: Changed this in https://github.com/cloudfoundry/bosh/pull/2782

aramprice:
Closing in favor of https://github.com/cloudfoundry/bosh/pull/2782

PR #2785, opened 2026-07-16
Switch config-server resource to config-server-release repo

aramprice:
The cloudfoundry/config-server repo has been archived; its contents were merged into cloudfoundry/config-server-release. Update the pipeline resource to use HTTPS against the new repo (removing the SSH deploy key), target the main branch, and update SOURCE_DIR in config_server_service.rb to point to src/config-server within the release repo where the Go source now lives.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -1879,9 +1879,8 @@ resources:
   - name: config-server
     type: git
     source:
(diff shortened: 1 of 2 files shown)


aramprice:
EasyCLA seems to be having issues. At some point it can be re-run using:

```
/easycla
```

beyhan (approved):

PR #2804, opened 2026-08-14
brats: fix upgrade from postgres 13 test

mkocher:

Postgres 10 does not compile on Resolute. This test was pulling in a version of postgres 13 that itself pulled in Postgres 10 for upgrade purposes. Switching to a version where postgres 13 does not include postgres 10.

I've manually validated the steps this test goes through using the manifests from the repo.

No


aramprice:
This upcoming test addition might be worth cribbing from
=> https://github.com/cloudfoundry/bosh/pull/2779

PR #2808, opened 2026-08-17
bdrats: run on the docker CPI, and actually apply local-stemcell.yml

mkocher:

 Resolute doesn't work on the warden CPI yet, and in general we want to standardize on the docker CPI and deprecate the warden CPI. So moving this build to use Docker.

Also fixes a silent bug: local-stemcell.yml was written but never passed with -o (omitted in 2e006f1b81), so the director ran on warden/cpi.yml's pinned stemcell instead of the tarball the job fetched.

Candidate release and stemcell are applied last so nothing can replace them, plus two assertions to keep this class of failure loud: the tarball's operating_system must match STEMCELL_OS, and the rendered director manifest's release and stemcell URLs must both be file://.

I've run a one off concourse task validating the tests pass when run under Docker.

No

### DIFF
--- a/ci/tasks/test-bdrats.sh
+++ b/ci/tasks/test-bdrats.sh
@@ -16,10 +16,19 @@ chmod +x "${BBR_BINARY_PATH}"
 export OVERRIDDEN_BOSH_DEPLOYMENT="${src_dir}/bosh-deployment"
 if [[ -e ${OVERRIDDEN_BOSH_DEPLOYMENT}/bosh.yml ]];then
   export BOSH_DEPLOYMENT_PATH=${OVERRIDDEN_BOSH_DEPLOYMENT}
+else
+  export BOSH_DEPLOYMENT_PATH=/usr/local/bosh-deployment
 fi
 
 STEMCELL_PATH="${PWD}/stemcell/$(basename stemcell/*.tgz)"
 STEMCELL_SHA1=$(sha1sum "${STEMCELL_PATH}" | awk '{print $1}')
+
+STEMCELL_OS="$(tar -Oxzf "${STEMCELL_PATH}" stemcell.MF \
+  | awk '/^operating_system:/ {gsub(/"/, "", $2); print $2}')"
+if [[ -z "${STEMCELL_OS}" ]]; then
+  echo "no operating_system in $(basename "${STEMCELL_PATH}")'s stemcell.MF" >&2
+  exit 1
+fi
 
 cat > "${BOSH_DEPLOYMENT_PATH}/local-stemcell.yml" <<'OPSEOF'
 - name: stemcell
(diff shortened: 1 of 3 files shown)


ci/tasks/test-bdrats.sh
```diff
@@ -30,18 +39,35 @@ cat > "${BOSH_DEPLOYMENT_PATH}/local-stemcell.yml" <<'OPSEOF'
...
-source start-bosh \
-  -o bbr.yml \
-  -o local-bosh-release-tarball.yml \
-  -o hm/disable.yml \
+# The candidate release and the stemcell under test come last, so no earlier
+# ops file can replace the artifacts being tested.
+#
+# Sourced rather than executed so the docker service can be stopped on exit;
+# see ci/dockerfiles/docker-cpi/README.md.
+source "${script_dir}/../dockerfiles/docker-cpi/start-bosh.sh" \
```
aramprice: I think that `start-bosh` is paved down on the docker-cpi image so there's a small risk in using `"${script_dir}/../dockerfiles/docker-cpi/start-bosh.sh"` (<- this get's put into `$PATH` on the image as `start-bosh`) since there might be drift between the CPI image, and the contents of `bosh-ci/tasks/dockerfiles/...`

aramprice: ... if this is intentional, carry on!

aramprice: Long term I'm not sure where the best "home" for this script is, though I would strongly prefer if we could execute it instead of `source`'ing it[1].

[1] I think the origin of `source start-bosh` is that it set a bunch of ENV vars... I think this may not be un-necessary now that many/all of the necessary vars are present under `/tmp/local-bosh/director/` and the like.

mkocher: I updated it to use the one in the container image, I think it should be the source of truth.

Sourcing it seems odd to me, the readme says it's necessary to avoid hanging the worker. No idea if that's still true or not, but seems outside the scope of these changes.

ci/tasks/test-bdrats.sh
```diff
@@ -30,18 +39,43 @@ cat > "${BOSH_DEPLOYMENT_PATH}/local-stemcell.yml" <<'OPSEOF'
...
+# check that the items being tested are the local copies
+for path in /releases/name=bosh/url /resource_pools/name=vms/stemcell/url; do
+  url="$(bosh int /tmp/local-bosh/director/bosh-director.yml --path "${path}")"
+  if [[ "${url}" != file://* ]]; then
+    echo "expected ${path} to be a local file, got '${url}'" >&2
+    exit 1
+  fi
+done
 
 BOSH_SSH_KEY="$(bosh int /tmp/local-bosh/director/creds.yml --path /jumpbox_ssh/private_key --json | jq .Blocks[0])"
```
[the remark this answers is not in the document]
mkocher: this is a CI container where it's running as root. file permissions would not provide any security.

ci/tasks/test-bdrats.sh
```diff
@@ -30,18 +39,43 @@ cat > "${BOSH_DEPLOYMENT_PATH}/local-stemcell.yml" <<'OPSEOF'
...
+source /tmp/local-bosh/director/bosh-env
+
+# check that the items being tested are the local copies
+for path in /releases/name=bosh/url /resource_pools/name=vms/stemcell/url; do
+  url="$(bosh int /tmp/local-bosh/director/bosh-director.yml --path "${path}")"
+  if [[ "${url}" != file://* ]]; then
+    echo "expected ${path} to be a local file, got '${url}'" >&2
+    exit 1
+  fi
+done
```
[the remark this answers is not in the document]
mkocher: this is just a sanity check, and it's good enough to prevent misordering the ops files

aramprice (approved):

PR #2810, opened 2026-08-18
Fix PostgresVersionHelper#local_version on macOS

aramprice:
Due to changes in Homebrew's postgress version string the existing implimentation has begun returning `(Homebrew)` instead of a major-version number. This commit updates the helper to fix this.

Verified that it continues to work on Linux (docker: ubuntu-noble, and bosh-integration-psotgres image).

Created in service of locally testing: https://github.com/cloudfoundry/bosh/pull/2811

### DIFF
--- a/src/spec/integration_support/postgres_version_helper.rb
+++ b/src/spec/integration_support/postgres_version_helper.rb
@@ -17,7 +17,7 @@ module IntegrationSupport
       end
 
       def local_version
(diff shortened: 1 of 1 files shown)


mkocher (approved):

Alphasite (approved):
This likely works but seems a little fragile. I assume it works on other OSes?

PR #2820, opened 2026-09-14
Distribute dynamic IPs across subnets via dynamic_subnet_strategy

### LINKED ISSUE #2622: Feature request: Even distribution of IP addresses, when multiple subnets per AZ are available
**Is your feature request related to a problem? Please describe.**

See background in "additional context" below

When a BOSH network contains multiple subnets for the same AZ, it will create VMs distributed into the specified AZs with even distribution. It will however take addresses from the first subnet, until that subnet is exhausted / filled. Only then will it move on to the next subnet, etc.

It is understood that this is an additional feature and should not change default behavior.

**Describe the solution you'd like**
It would be nice if BOSH could evenly distribute VMs into all available subnets evenly, as it does for AZs.

**Describe alternatives you've considered**

We are currently assigning static IP addresses manually from known pools, based on the subnets provided by the BOSH cloud config, and assign them manually.

**Additional context**

When deploying CF using BOSH on hyperscalers, very large CF foundations require NAT gateways in order to manage the apps' egress requests.

NAT gateways are finite resources and can be exhausted with connections. Using further NAT gateways requires separate subnets, at least on AWS and some other hyperscalers. In order to divide egress traffic onto these multiple NAT gateways, subnets are created, each of which is tied to a NAT gateway.

This is also why the even distribution of VMs to subnets becomes important, as it in turn distributes the egress from those VMs to the different NAT gateways.

plamen-bardarov:

Implements the even-distribution requested in #2622: when an AZ has more than one subnet on a manual network, spread auto-allocated (dynamic) IPs across those subnets instead of filling the first before using the next.

Adds `director.dynamic_subnet_strategy` (default `first_fit`):

- `first_fit` (default): unchanged legacy behavior.
- `least_loaded`: place each new VM in the least-loaded subnet of its AZ, with nic_group follow-the-leader co-location so all networks sharing an ENI land on the same subnet.

Implemented as a strategy under `DeploymentPlan::SubnetDistribution`; `IpProvider` selects the strategy and delegates subnet ordering. `least_loaded` caches per-subnet counts per deploy (seeded lazily, kept current on allocate/release) so balancing stays O(1) per reservation.

Closes #2622. No related PRs or discussions.

- Unit: `config_spec`, `subnet_distribution` + `first_fit` + `least_loaded` specs, the full `ip_provider` spec directory.
- Release ERB template rendering: `spec/director.yml.erb_spec.rb`.
- `rubocop` on all changed files.
- Full unit suite and integration run via CI on this PR.

The director can now distribute auto-allocated (dynamic) IPs evenly across multiple subnets within an AZ via the new opt-in `director.dynamic_subnet_strategy` property (`least_loaded`). The default (`first_fit`) preserves existing behavior.

No. The default `first_fit` is byte-for-byte the previous behavior; `least_loaded` is opt-in.

-

### AI Review Feedback

All CodeRabbit comments addressed: single-candidate co-location bypass, static-sibling co-location, static-release count guard, and the least_loaded performance (per-deploy count caching); docstring coverage and this description updated.


plamen-bardarov:
Closing to recreate the PR — the review bot run got stuck on this one and can't be retriggered. Reopening as a fresh PR with the same branch, title, and description.

